PR #4, opened 2025-03-20
Port `.github` templates and checks to arrow-rs-object-store crate

alamb:
- Part of https://github.com/apache/arrow-rs-object-store/issues/2

This PR ports relevant files from https://github.com/apache/arrow-rs/tree/main/.github/ to this repo

Issue #2, opened 2025-03-20
[EPIC] Port object_store content from arrow-rs repository


alamb:
- Follow on to https://github.com/apache/arrow-rs/issues/6183

This epic tracks the items needed to move the object_store code from https://github.com/apache/arrow-rs to this repo
- [x] Bring the contents of `object_store` to this repo, preserving history: https://github.com/apache/arrow-rs-object-store/pull/3
- [x] Get CI running in this repo: https://github.com/apache/arrow-rs-object-store/pull/4
- [x] https://github.com/apache/arrow-rs-object-store/pull/6
- [x] Remove old object_store code from arrow-rs: https://github.com/apache/arrow-rs/pull/7316
- [x] Update links in documentation, leave back links in arrow-rs: https://github.com/apache/arrow-rs/pull/7317
- [x] Move all tickets from arrow-rs to this repo

alamb:
I will begin moving the code now

assignUser:
@alamb do you want to move only open issues or closed issues as well? Caveat, if a label on an issue doesn't exists in the target repo it is dropped. I have done this for go and java with the `gh` cli

alamb:
> [@alamb](https://github.com/alamb) do you want to move only open issues or closed issues as well? Caveat, if a label on an issue doesn't exists in the target repo it is dropped. I have done this for go and java with the `gh` cli

Moving closed issues would be amazing. 

@assignUser  do you have the scripts you used (or can you point me at them) for how you migrated the issues using `gh`?

assignUser:
```bash
gh issue list -R apache/arrow-rs --label "object-store" -L 1000 --state "all" --json 'number' -q '.[].number' |  xargs -I% gh issue transfer % https://github.com/apache/arrow-rs-object-store
```
This will transfer all 248 issues with the "object-store" label to this repo. Any references to these issues in PRs etc. will break afaik. As we have those in commit messages in the monorepo we only transfered open issues but from what I can see this is not the case for you.

alamb:
Thanks @assignUser 

I ran the following commands:

```
gh issue list -R apache/arrow-rs --label "object-store" -L 1000 --state "all" --json 'number' -q '.[].number' > object_store_prs.txt
```

And spot checked a few random numbers in to make sure they were actually issues with object store (which they were)
- https://github.com/apache/arrow-rs/issues/5329
- https://github.com/apache/arrow-rs/issues/2096

Which look good

[...]

PR #6, opened 2025-03-20
Fix CI jobs

alamb:

- Part of https://github.com/apache/arrow-rs-object-store/issues/2
- Follow on to https://github.com/apache/arrow-rs-object-store/pull/4

# Rationale for this change

I needed to have the workflows run before I could debug them. 

This PR fixes them and gets a green CI  


.github/workflows/dev_pr.yml
```diff
@@ -1,51 +0,0 @@
-# Licensed to the Apache Software Foundation (ASF) under one
```
alamb, line 1: The labeler job isn't going to be useful if it just puts the object_store label on every PR so I removed it

PR #257, opened 2025-03-20
Add content length to PUT GCP multipart complete

jkosh44:

None

Closes #.

# Rationale for this change
 
Fixes unreported error

This commit fixes the GCP mulipart complete implementation by adding the Content-Length header to the PUT XML requests that happens when the completed parts are empty. GCP is strict about setting the Content-Length header on requests with empty bodies, so previously this would result in a 411 error.

Yes, empty multipart uploads don't cause 411 in GCP.

### DIFF
--- a/src/gcp/client.rs
+++ b/src/gcp/client.rs
@@ -517,6 +517,7 @@ impl GoogleCloudStorageClient {
             // GCS doesn't allow empty multipart uploads
             let result = self
                 .request(Method::PUT, path)
+                .header(&CONTENT_LENGTH, "0")
                 .idempotent(true)
                 .do_put()
                 .await?;


alamb (approved):
Thank you @jkosh44

jkosh44:
@alamb I wasn't able to come up with a good unit test for this.l

alamb:
Thanks @jkosh44  -- I am going to wait to merge this PR for a day or so to let some of the other maintainers double check I got this new repo set up correctly, but I think this PR is ready to merge from my perspective

src/gcp/client.rs
```diff
@@ -517,6 +517,7 @@ impl GoogleCloudStorageClient {
             // GCS doesn't allow empty multipart uploads
             let result = self
                 .request(Method::PUT, path)
+                .header(&CONTENT_LENGTH, "0")
                 .idempotent(true)
                 .do_put()
                 .await?;
```
tustvold, line 523:
```suggestion
            let result = self.put(path, PutPayload::new(), Default::default()).await?;
```

This will help to avoid any potential divergence in future, and also perhaps be a bit more clear what is going on?

jkosh44: I've updated to do this, but I also had to update the implementation of `self.put` to set `Content-Length` for empty payloads.

src/gcp/client.rs
```diff
@@ -397,8 +397,13 @@ impl GoogleCloudStorageClient {
...
 
-        let builder = self
-            .request(Method::PUT, path)
+        let builder = self.request(Method::PUT, path);
+        let builder = if payload.content_length() == 0 {
+            builder.header(&CONTENT_LENGTH, "0")
+        } else {
+            builder
+        };
+        let builder = builder
```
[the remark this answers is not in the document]
jkosh44-datadog: Ah got it, I just reverted this change.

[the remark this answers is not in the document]
jkosh44: Ah got it, I just reverted this change.

tustvold (approved):

src/gcp/client.rs
```diff
@@ -516,9 +516,7 @@ impl GoogleCloudStorageClient {
         if completed_parts.is_empty() {
             // GCS doesn't allow empty multipart uploads
             let result = self
-                .request(Method::PUT, path)
-                .idempotent(true)
-                .do_put()
+                .put(path, PutPayload::new(), Default::default())
                 .await?;
             self.multipart_cleanup(path, multipart_id).await?;
```
jkosh44, line 521: @tustvold I was also trying to think about how to make this more obvious. Perhaps the following structure is more clear? The comment clarifies that we're falling back to a non-multipart upload. Also moving the PUT after the cleanup makes it clear that the cleanup is not affecting the PUT. What do you think?

```rust
            // GCS doesn't allow empty multipart uploads, so fallback to regular upload.
            self.multipart_cleanup(path, multipart_id).await?;
            let result = self
                .put(path, PutPayload::new(), Default::default())
                .await?;
```

tustvold: I don't feel strongly either way

jkosh44: Ok, I made the change.

alamb:
Thank you @jkosh44  -- this is the first PR in the new repo to merge with feature change. So exciting!

jkosh44:
Thank you for the reviews @alamb and @tustvold!

andrew-pienso:
Thank you all!

andrew-pienso:
> > Thank you all!
> 
> @andrew-pienso if you still have a repro of #278, it would definitely be helpful to double check that this does in fact resolve your issue.

When I filed this bug, this [example here would repro](https://github.com/delta-io/delta-rs/issues/2394#issuecomment-2132886429). I played around with this for a good half hour and am admittedly not able to repro this anymore in my environment.

PR #303, opened 2025-03-21
Bump `rand` to 0.9

mbrobbel:

- Closes #5.

# Rationale for this change
 
https://rust-random.github.io/book/update-0.9.html

The required changes to migrate from `0.8` to `0.9`.

`rand` dependency bumped to `0.9`.

### DIFF
--- a/src/client/backoff.rs
+++ b/src/client/backoff.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use rand::{rng, prelude::*};
+use rand::{prelude::*, rng};
 use std::time::Duration;
 
 /// Exponential backoff with decorrelated jitter algorithm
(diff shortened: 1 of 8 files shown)


src/client/backoff.rs
```diff
@@ -78,7 +78,7 @@ impl Backoff {
 
     /// Creates a new `Backoff` with the optional `rng`
     ///
-    /// Used [`rand::thread_rng()`] if no rng provided
+    /// Used [`rand::rng()`] if no rng provided
```
alamb, line 81: I think sadly this change means it is a breaking API change, right?

alamb (commented):
Thank you @mbrobbel  🙏

alamb (approved):
Thanks @mbrobbel

PR #305, opened 2025-03-24
Update dependabot config after repo move

mbrobbel:

None.

# Rationale for this change
 
The dependabot config can be updated.

Removed `/object_store` path config.

No.


tustvold (commented):
Should probably update the labels

PR #316, opened 2025-04-01
feat: make some helpers/utils public

crepererum:
\-

# Rationale for this change
Makes it easier to implement out-of-tree / 3rd-party stores w/o replicating the code.

- public `GetRange` helpers
- public `GetOptions::check_precondition`

More public APIs.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -208,7 +208,8 @@ pub enum GetRange {
 }
 
 #[derive(Debug, thiserror::Error)]
-pub(crate) enum InvalidGetRange {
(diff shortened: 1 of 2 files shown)


src/util.rs
```diff
@@ -208,7 +208,8 @@ pub enum GetRange {
 }
 
 #[derive(Debug, thiserror::Error)]
-pub(crate) enum InvalidGetRange {
+#[expect(missing_copy_implementations)] // allow easier extension
```
tustvold, line 211: Might also want non_exhaustive

tustvold (approved):
Just a minor nit

Issue #302, opened 2025-03-21
Enable anonymous access to GCS buckets


paraseba:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently we are unable to use object_store without authorization on GCP. Skipping authorization is required for interacting with public buckets that allow anonymous read access.

**Describe the solution you'd like**
Introduce a mechanism similar to Azure and AWS: `skip_signature`

PR #322, opened 2025-04-04
Enable anonymous access to GCS buckets

kylebarron:

Closes https://github.com/apache/arrow-rs-object-store/issues/302.

# Rationale for this change

- Add `skip_signature` config parameter
- Add `with_skip_signature` to `GoogleCloudStorageBuilder`
- Add `CredentialExt` in `src/gcp/credential.rs`, similar to what already exists for AWS in `src/aws/credential.rs`. This allows for easily calling `with_bearer_auth` in the request builder chain without having to do an `if/else` to handle `credentials` being `None` there.

**Questions**

- How should I test this? I looked for other existing tests for `skip_signature` and it looks like the one existing one for AWS is ignored: https://github.com/apache/arrow-rs-object-store/blob/e975177814778b711e11b048bb5c703d2b893427/src/aws/mod.rs#L690-L714

I was able to successfully test this in the obstore Python bindings (https://github.com/developmentseed/obstore/pull/404) on a GCP open data bucket:

```py
store = GCSStore("gcp-public-data-arco-era5", skip_signature=True)
```

```py
next(store.list())
# [{'path': 'ar/1959-2022-1h-240x121_equiangular_with_poles_conservative.zarr/.zattrs',
#   'last_modified': datetime.datetime(2023, 9, 14, 7, 6, 30, 187000, tzinfo=datetime.timezone.utc),
#   'size': 2,
#   'e_tag': '"99914b932bd37a50b983c5e7c90ae93b"',
#   'version': None},
#  {'path': 'ar/1959-2022-1h-240x121_equiangular_with_poles_conservative.zarr/.zgroup',
#   'last_modified': datetime.datetime(2023, 9, 14, 7, 6, 28, 987000, tzinfo=datetime.timezone.utc),
#   'size': 24,
#   'e_tag': '"e20297935e73dd0154104d4ea53040ab"',
#   'version': None},
# ...

orjson.loads(
    memoryview(
        store.get(
            "ar/1959-2022-1h-240x121_equiangular_with_poles_conservative.zarr/.zmetadata",
        ).bytes()
    ),
)

# {'metadata': {'.zattrs': {},
#   '.zgroup': {'zarr_format': 2},
#   '10m_u_component_of_wind/.zarray': {'chunks': [8, 240, 121],
#    'compressor': {'blocksize': 0,
#     'clevel': 5,
#     'cname': 'lz4',
#     'id': 'blosc',
#     'shuffle': 1},
#    'dtype': '<f4',
#    'fill_value': 'NaN',
#    'filters': None,
#    'order': 'C',
#    'shape': [552264, 240, 121],
#    'zarr_format': 2},
#   '10m_u_component_of_wind/.zattrs': {'_ARRAY_DIMENSIONS': ['time',
#     'longitude',
#     'latitude'],
#    'long_name': '10 metre U wind component',
#    'short_name': 'u10',
#    'units': 'm s**-1'},
# ...
```

(cc @paraseba @tomnicholas @maxrjones)

Adds `skip_signature`.

### DIFF
--- a/src/gcp/client.rs
+++ b/src/gcp/client.rs
@@ -632,13 +632,8 @@ impl GetClient for GoogleCloudStorageClient {
             request = request.query(&[("generation", version)]);
         }
 
-        request = request.with_bearer_auth(credential.as_deref());
-
-        // if !credential.bearer.is_empty() {
-        //     request = request.bearer_auth(&credential.bearer);
-        // }
-
         let response = request
+            .with_bearer_auth(credential.as_deref())
             .with_get_options(options)
             .send_retry(&self.config.retry_config)
             .await
--- a/src/gcp/client.rs
+++ b/src/gcp/client.rs
@@ -25,6 +25,7 @@ use crate::client::s3::{
     ListResponse,
 };
 use crate::client::{GetOptionsExt, HttpClient, HttpError, HttpResponse};
+use crate::gcp::credential::CredentialExt;
 use crate::gcp::{GcpCredential, GcpCredentialProvider, GcpSigningCredentialProvider, STORE};
 use crate::multipart::PartId;
 use crate::path::{Path, DELIMITER};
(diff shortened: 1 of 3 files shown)


src/gcp/client.rs
```diff
@@ -640,9 +632,11 @@ impl GetClient for GoogleCloudStorageClient {
...
         }
 
-        if !credential.bearer.is_empty() {
-            request = request.bearer_auth(&credential.bearer);
-        }
+        request = request.with_bearer_auth(credential.as_deref());
+
+        // if !credential.bearer.is_empty() {
+        //     request = request.bearer_auth(&credential.bearer);
+        // }
```
kylebarron, line 639: Reading this code, it looks like previously if `credential.bearer` was `""`, it would also skip signing the request, but only for `GET` requests. Should we maintain that functionality? I.e. should we change the diff below to

```rs
impl CredentialExt for HttpRequestBuilder {
    fn with_bearer_auth(self, credential: Option<&GcpCredential>) -> Self {
        match credential {
            Some(credential) => {
                if credential.bearer.is_empty() {
                    return self;
                }

                self.bearer_auth(&credential.bearer)
            }
            None => self,
        }
    }
}
```

kylebarron: [`640f7f2` (#322)](https://github.com/apache/arrow-rs-object-store/pull/322/commits/640f7f21e82ce4cbc9402c0f9e7aa3884fff3dc5) implements this change, so we also now skip signing when bearer is `""`, to ensure this isn't a breaking change.

kylebarron:
It looks like there are extremely few public datasets on GCS (https://cloud.google.com/storage/docs/public-datasets#available_public_datasets_on) ([most](https://cloud.google.com/datasets?hl=en) are stored in BigQuery), so this might be considered kinda niche, but especially the [ERA5 dataset](https://console.cloud.google.com/marketplace/product/bigquery-public-data/arco-era5) is useful for us to be able to access.

paraseba (approved):
This looks great @kylebarron !

tustvold (approved):
Looks good to me, it is perhaps a little odd to call this skip_signature given GCP doesn't sign requests, but this is consistent with the naming of the similar feature for Azure and AWS so :+1:

alamb:
Thanks @tustvold and @kylebarron

Issue #291, opened 2025-03-05
Expose list_paginated in object_store


HugoCasa:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
I'm building a bucket explorer and would like to paginate objects at the top-level so basically using both delimiter and offset at the same time. 

**Describe the solution you'd like**
Expose a function with both offset and delimiter.
I see that the function is already implemented for AWS/GCP/Azure in the ListClientExt trait but not exposed.

**Describe alternatives you've considered**
I've considered list_with_offset and list_with_delimiter but using those would be inefficient in my case.

tustvold:
#371 Adds a PaginatedListStore with support for this

PR #324, opened 2025-04-07
feat: support list with ListOpts which contains the offset, max_keys,…

### LINKED ISSUE #295: Add ObjectStore::list_opts

We should add an ObjectStore::list_opts to allow

* Specifying both a start offset and delimiter apache/arrow-rs-object-store#291
* Providing `Extensions` apache/arrow-rs#7213 

I _think_ the following API should work and eventually allow providing default implementations of `list` and `list_with_delimiter` that call through to `list_opts`

```
struct ListOptions {
    pub offset: Option<Path>,
    pub delimiter: bool,
    pub extensions: Extensions
}

fn list_opts(&self, opts: ListOptions) -> BoxStream<'static, Result<ListResult>>
```

We could expose an API that more explicitly exposes the paginated nature of the underlying APIs, however, such APIs are quite cumbersome to use, and wouldn't obviously translate to LocalFileSystem.

**Additional context**

stayrascal:
… delimiter

Closes #295 and #291.

# Rationale for this change
 

- add a `list_opts` method in `ObjectStore` traits and implement the `list_opts` method in each implementation.
- the ListOptions support offset, delimiter, max_keys, extension(maybe we can include the prefix as a option into ListOptions, but in order to keep the style from other methods, keep prefix as a alone parameter currently)
- add a list_with_composite_conditions test case by compose prefix, offset, delimiter, max_keys

- TBD: whether mark these methods: list, list_with_offset, list_with_delimiter as `deprecated`

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -173,6 +173,8 @@ pub struct AmazonS3Builder {
     request_payer: ConfigValue<bool>,
     /// The [`HttpConnector`] to use
     http_connector: Option<Arc<dyn HttpConnector>>,
+    /// Max list size per request, by default 1000.
+    max_list_size_per_request: Option<ConfigValue<usize>>,
 }
 
 /// Configuration keys for [`AmazonS3Builder`]
@@ -891,6 +893,12 @@ impl AmazonS3Builder {
         self
     }
 
+    /// Set the max keys per list request. It's almost used for test paginated listing.
+    pub fn with_max_list_size_per_request(mut self, max_list_size_per_request: usize) -> Self {
+        self.max_list_size_per_request = Some(ConfigValue::Parsed(max_list_size_per_request));
+        self
+    }
+
     /// Create a [`AmazonS3`] instance from the provided values,
     /// consuming `self`.
     pub fn build(mut self) -> Result<AmazonS3> {
@@ -1029,6 +1037,10 @@ impl AmazonS3Builder {
             S3EncryptionHeaders::default()
         };
 
+        let max_list_size_per_request = self
+            .max_list_size_per_request
+            .unwrap_or(ConfigValue::Parsed(1000))
+            .get()?;
         let config = S3Config {
             region,
             endpoint: self.endpoint,
@@ -1046,6 +1058,7 @@ impl AmazonS3Builder {
             conditional_put: self.conditional_put.get()?,
             encryption_headers,
             request_payer: self.request_payer.get()?,
+            max_list_keys_per_request: max_list_size_per_request,
         };
 
         let http_client = http.connect(&config.client_options)?;
--- a/src/client/list.rs
+++ b/src/client/list.rs
@@ -70,6 +70,12 @@ impl<T: ListClient + Clone> ListClientExt for T {
             self.clone(),
             (prefix, offset, max_keys, extensions),
             move |client, (prefix, offset, max_keys, extensions), token| async move {
+                if let Some(remaining) = max_keys {
+                    if remaining == 0 {
+                        return Ok((ListResult::empty(), (prefix, offset, max_keys, extensions), None));
+                    }
+                }
+
                 let (r, next_token) = client
                     .list_request(
                         prefix.as_deref(),
@@ -81,7 +87,7 @@ impl<T: ListClient + Clone> ListClientExt for T {
                     )
                     .await?;
                 let key_count = r.common_prefixes.len() + r.objects.len();
-                let remaining = max_keys.map(|x| x - key_count);
+                let remaining = max_keys.map(|x| (x - key_count).max(0));
                 let next_token = match remaining {
                     None => next_token,
                     Some(remaining) if remaining > 0 => next_token,
--- a/src/client/list.rs
+++ b/src/client/list.rs
@@ -80,7 +80,8 @@ impl<T: ListClient + Clone> ListClientExt for T {
                         extensions.clone(),
                     )
                     .await?;
-                let remaining = max_keys.map(|x| x - r.key_count);
+                let key_count = r.common_prefixes.len() + r.objects.len();
+                let remaining = max_keys.map(|x| x - key_count);
                 let next_token = match remaining {
                     None => next_token,
                     Some(remaining) if remaining > 0 => next_token,
--- a/src/client/list.rs
+++ b/src/client/list.rs
@@ -17,12 +17,11 @@
 
 use crate::client::pagination::stream_paginated;
 use crate::path::Path;
-use crate::Result;
-use crate::{ListResult, ObjectMeta};
+use crate::ListResult;
+use crate::{ListOpts, Result};
 use async_trait::async_trait;
 use futures::stream::BoxStream;
-use futures::{StreamExt, TryStreamExt};
-use std::collections::BTreeSet;
+use futures::StreamExt;
 
 /// A client that can perform paginated list requests
 #[async_trait]
@@ -33,6 +32,8 @@ pub(crate) trait ListClient: Send + Sync + 'static {
         delimiter: bool,
         token: Option<&str>,
         offset: Option<&str>,
+        max_keys: Option<usize>,
+        extensions: ::http::Extensions,
     ) -> Result<(ListResult, Option<String>)>;
 }
 
@@ -42,20 +43,8 @@ pub(crate) trait ListClientExt {
     fn list_paginated(
         &self,
         prefix: Option<&Path>,
-        delimiter: bool,
-        offset: Option<&Path>,
+        opts: ListOpts,
     ) -> BoxStream<'static, Result<ListResult>>;
-
-    fn list(&self, prefix: Option<&Path>) -> BoxStream<'static, Result<ObjectMeta>>;
-
-    #[allow(unused)]
-    fn list_with_offset(
-        &self,
-        prefix: Option<&Path>,
-        offset: &Path,
-    ) -> BoxStream<'static, Result<ObjectMeta>>;
-
-    async fn list_with_delimiter(&self, prefix: Option<&Path>) -> Result<ListResult>;
 }
 
 #[async_trait]
@@ -63,9 +52,15 @@ impl<T: ListClient + Clone> ListClientExt for T {
     fn list_paginated(
         &self,
         prefix: Option<&Path>,
-        delimiter: bool,
-        offset: Option<&Path>,
+        opts: ListOpts,
     ) -> BoxStream<'static, Result<ListResult>> {
+        let ListOpts {
+            delimiter,
+            offset,
+            max_keys,
+            extensions,
+        } = opts;
+
         let offset = offset.map(|x| x.to_string());
         let prefix = prefix
             .filter(|x| !x.as_ref().is_empty())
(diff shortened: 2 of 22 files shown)


stayrascal:
Hi @tustvold, I have submitted a PR about #295, but it seems that the CI/workflow doesn't been triggered, need ti be approved, could you please help to take a look, and help to review this PR? thx.

tustvold:
I'll try to find some time to review this over the weekend

src/aws/builder.rs
```diff
@@ -891,6 +893,12 @@ impl AmazonS3Builder {
         self
     }
 
+    /// Set the max keys per list request. It's almost used for test paginated listing.
```
tustvold, line 896: This functionality never appears to be used or tested, is it necessary?

src/lib.rs
```diff
@@ -720,7 +721,22 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
+            .map_ok(|r| futures::stream::iter(r.objects.into_iter().map(Ok)))
+            .try_flatten()
+            .boxed()
+    }
+
+    /// List all the objects with given options defined in [`ListOpts`]
+    ///
+    /// Prefixes are evaluated on a path segment basis, i.e. `foo/bar` is a prefix of `foo/bar/x` but not of
+    /// `foo/bar_baz/x`. List is recursive, i.e. `foo/bar/more/x` will be included.
+    fn list_opts(
```
tustvold, line 735: This formulation is a breaking change, if we want to get this in any time in the next few months we probably want to instead provide a default implementation that calls through to list - erroring if not provided

src/prefix.rs
```diff
@@ -165,6 +165,36 @@ impl<T: ObjectStore> ObjectStore for PrefixStore<T> {
...
+        let offset = options.offset.map(|p| self.full_path(&p));
+        let opts = ListOpts {
+            offset,
+            delimiter: options.delimiter,
+            max_keys: options.max_keys,
+            extensions: options.extensions,
+        };
+        let s = self.inner.list_opts(Some(&prefix), opts);
+
+        s.map_ok(move |lst| ListResult {
```
tustvold, line 183: Perhaps we could encapsulate this, it is identical to list_with_delimiter below

src/memory.rs
```diff
@@ -311,52 +311,70 @@ impl ObjectStore for InMemory {
...
+                if common_prefixes.len() + objects.len() >= may_keys {
+                    break;
+                }
+            }
+
             if !k.as_ref().starts_with(prefix.as_ref()) {
                 break;
             }
 
+            if k <= offset {
```
tustvold, line 374: Could we instead just start listing from `offset` instead of `prefix` if provided?

src/memory.rs
```diff
@@ -311,52 +311,70 @@ impl ObjectStore for InMemory {
...
         let prefix = prefix.unwrap_or(&root);
+        let offset = offset.unwrap_or(&root);
 
         let mut common_prefixes = BTreeSet::new();
-
         // Only objects in this base level should be returned in the
         // response. Otherwise, we just collect the common prefixes.
         let mut objects = vec![];
         for (k, v) in self.storage.read().map.range((prefix)..) {
+            if let Some(may_keys) = max_keys {
```
tustvold, line 364: You could lift the conditional out by doing `let max_keys = max_keys.unwrap_or(usize::MAX)`.

src/lib.rs
```diff
@@ -745,7 +767,29 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
+                ..ListOpts::default()
+            },
+        );
+
+        let mut common_prefixes = BTreeSet::new();
+        let mut objects = Vec::new();
+
+        while let Some(result) = stream.next().await {
+            let response = result?;
+            common_prefixes.extend(response.common_prefixes.into_iter());
```
tustvold, line 784: Will this potentially result in the same common_prefix appearing multiple times?

src/memory.rs
```diff
@@ -383,31 +401,57 @@ impl ObjectStore for InMemory {
...
-        let entry = self.entry(from)?;
-        let mut storage = self.storage.write();
-        if storage.map.contains_key(to) {
-            return Err(Error::AlreadyExists {
-                path: to.to_string(),
+        let objects = match max_keys {
+            Some(max_keys) if max_keys < values.len() => {
+                values.truncate(max_keys);
+                values
             }
```
tustvold, line 444: Could we instead use `Stream::take`?

tustvold (commented):
So I have done a first pass of a review, but unfortunately have run out of time for today.

I wonder if we might be able to find someway to reduce the size of this diff, its really quite hard to review at the moment... Perhaps we could split adding list_opts and adding the max-keys?

src/client/list.rs
```diff
@@ -33,6 +32,8 @@ pub(crate) trait ListClient: Send + Sync + 'static {
         delimiter: bool,
         token: Option<&str>,
         offset: Option<&str>,
+        max_keys: Option<usize>,
```
tustvold, line 35: Perhaps this could use ListOpts?

src/aws/builder.rs
stayrascal: yeah, i'm trying to test paginated listing, for example, limit n objects per list request instead of preparing more than 1k objects before testing in ut, but i found it's a little bit hard to add a general test case in `integration.rs` because they used a common object store, e.g. AmazonS3, otherwise I need to create separate store and test cover it. It can be removed.

src/lib.rs
stayrascal: Sorry for I didn't get "this is a breaking change" here, do you mean maybe some other crates import the object_store crate and implement the ObjectStore traits, they have to implement the `list_opts` method if they upgrade the object_store version, right? If it is, it might be problem, provide a default implementation is a better option, but it also means we cannot remove other list methods to make this trait more concise. 

I planed marked other list methods as deprecated, it seems that it's better to do it in a big version.

src/lib.rs
stayrascal: it could be, but the BTreeSet should deduplicate the common_prefixes.

src/memory.rs
stayrascal: sure, maybe start listing from the bigger one between `offset` and `prefix`, because technically `offset` could be less than `prefix`.

src/memory.rs
stayrascal: will try

stayrascal:
> So I have done a first pass of a review, but unfortunately have run out of time for today.
> 
> I wonder if we might be able to find someway to reduce the size of this diff, its really quite hard to review at the moment... Perhaps we could split adding list_opts and adding the max-keys into separate PRs?
> 
> I also think we should try to make this not a breaking change, as it is unlikely we're going to look to make a breaking release in the next 4 or so months.

Thanks a lot for got lots of useful suggestions from you, i will split the this diff and also avoid the breaking change, thanks again.

stayrascal:
Hi @tustvold, I tried to reduce the scope of this PR, could you please help to review it when you free, thanks a lot.

src/lib.rs
```diff
@@ -722,6 +722,64 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
+                        objects,
+                    })
+                } else {
+                    Ok(ListResult {
+                        common_prefixes: vec![],
+                        objects: batch,
+                    })
+                }
+            })
+            .boxed()
```
tustvold, line 780:
```suggestion
        // Add the default implementation via calling list_with_offset methods avoid bring
        // a breaking change, the default implementation will be removed, and default
        // implementations added for the other list methods using it
        futures::stream::once(futures::future::ready(Err(Error::NotImplemented))).boxed()
```
I don't think there is a coherent way to provide a default implementation here, the below ignores `extensions` and doesn't push down the delimiter.

src/lib.rs
```diff
@@ -1283,6 +1359,22 @@ pub struct PutResult {
     pub version: Option<String>,
 }
 
+/// Options for [`ObjectStore::list_opts`]
+#[derive(Debug, Clone, Default)]
+pub struct ListOpts {
```
tustvold, line 1364:
```suggestion
pub struct ListOptions {
```

src/lib.rs
```diff
@@ -1283,6 +1359,22 @@ pub struct PutResult {
...
+    /// The listed objects whose locations are greater than `offset`
+    pub offset: Option<Path>,
+    /// True means returns common prefixes (directories) in addition to object in [`ListResult`]
+    pub delimiter: bool,
+    /// Implementation-specific extensions. Intended for use by [`ObjectStore`] implementations
+    /// that need to pass context-specific information (like tracing spans) via trait methods.
+    ///
+    /// These extensions are ignored entirely by backends offered through this crate.
+    ///
+    /// They are also eclused from [`PartialEq`] and [`Eq`].
```
tustvold, line 1374:
```suggestion
```
I appreciate this is copied, but it is also not true

src/client/list.rs
```diff
@@ -42,86 +42,44 @@ pub(crate) trait ListClientExt {
...
                     )
                     .await?;
-                Ok((r, (prefix, offset), next_token))
+                Ok((r, (prefix, offset, extensions), next_token))
             },
         )
         .boxed()
     }
-
-    fn list(&self, prefix: Option<&Path>) -> BoxStream<'static, Result<ObjectMeta>> {
```
tustvold, line 92: Can we revert this change please, the purpose of ListClientExt is that the ListClient implementations just implement list_request and all the ObjectStore list methods are provided for them.

Once list_opts is required we can remove ListClientExt, with the contents of these methods moving into the ObjectStore default impls, but until then...

src/client/list.rs
```diff
@@ -42,86 +42,44 @@ pub(crate) trait ListClientExt {
...
-        &self,
-        prefix: Option<&Path>,
-        offset: &Path,
-    ) -> BoxStream<'static, Result<ObjectMeta>>;
-
-    async fn list_with_delimiter(&self, prefix: Option<&Path>) -> Result<ListResult>;
 }
 
 #[async_trait]
 impl<T: ListClient + Clone> ListClientExt for T {
```
tustvold, line 50: Can we add a list_opts impl here, to avoid duplicating the plumbing logic in the 3 cloud provider stores

src/http/mod.rs
```diff
@@ -157,6 +158,93 @@ impl ObjectStore for HttpStore {
         .boxed()
     }
 
+    fn list_opts(
```
tustvold, line 161: This implementation seems over complicated given that webdav doesn't support pagination.

Could we perhaps reuse more of the existing logic?

src/integration.rs
```diff
@@ -960,6 +960,102 @@ pub async fn list_with_delimiter(storage: &DynObjectStore) {
     assert!(content_list.is_empty());
 }
 
+/// Tests listing with composite conditions by [`ObjectStore::list_opts`]
+pub async fn list_with_composite_conditions(storage: &DynObjectStore) {
```
tustvold, line 964: This doesn't appear to be wired into all of the integration tests?

src/lib.rs
```diff
@@ -1283,6 +1359,22 @@ pub struct PutResult {
     pub version: Option<String>,
 }
 
+/// Options for [`ObjectStore::list_opts`]
+#[derive(Debug, Clone, Default)]
+pub struct ListOpts {
+    /// The listed objects whose locations are greater than `offset`
```
tustvold, line 1365:
```suggestion
    /// Only list objects greater than `offset`
```

src/lib.rs
```diff
@@ -1283,6 +1359,22 @@ pub struct PutResult {
...
+    pub offset: Option<Path>,
+    /// True means returns common prefixes (directories) in addition to object in [`ListResult`]
```
tustvold, line 1367:
```suggestion
    /// Only list up to the next path delimiter `/`  [`ListResult`]
    ///
    /// See [`ObjectStore::list_with_delimiter`]
```

tustvold (commented):
I've done another pass, but again run out of time I have to spend on this.

As it stands this PR is really quite hard to review, it makes lots of changes to lots of places without it being very easy to connect things together.

I'll see if I can't find some time to bash out a smaller more incremental version of this PR in the next couple of weeks, but unfortunately my time is very limited at the moment.

src/client/list.rs
stayrascal: do you mean add a `list_opts` and then calling `list_paginated` within `list_opts` impl, and the 3 clould provider stores call `self.client.list_opts(prefix, options)`? like this:
```
fn list_opts(
    &self,
    prefix: Option<&Path>,
    opts: ListOpts,
) -> BoxStream<'static, Result<ListResult>> {
    let client = self.clone();
    client.list_paginated(prefix, opts)
}
```
The 3 clould provider are calling `list_paginated` method directly during implementing `list_opts` of ObjectStore, because the `list_opts` of `ListClientExt` will be very simple, just transfer request.

src/client/list.rs
stayrascal: sure

src/client/list.rs
tustvold: I'm not entirely sure I follow what you are saying, but if list_paginated ends up with the same signature as list_opts, it might as well just get renamed to list_opts.

src/lib.rs
stayrascal: sure, throw `NotImplemented` error is a simple way,

src/http/mod.rs
stayrascal: yeah, i considered reuse the logic of `list` and `list_with_delimiter`, but since the response of `list_opts` is a BoxStream with static lifecycle, so cannot reuse  `list_with_delimiter` directly(please correct me if misunderstand something).

stayrascal:
> I've done another pass, but again run out of time I have to spend on this.
> 
> As it stands this PR is really quite hard to review, it makes lots of changes to lots of places without it being very easy to connect things together.
> 
> I'll see if I can't find some time to bash out a smaller more incremental version of this PR in the next couple of weeks, but unfortunately my time is very limited at the moment.

I'm so sorry that this PR has taken up too much of your time. At the same time, I'm extremely grateful for your very constructive suggestions. I have already revised the code and made the entire PR simplify as much as possible. I hope this can reduce the difficulty of the review. Thanks again for your help.

tustvold:
https://github.com/apache/arrow-rs-object-store/pull/371 contains an alternative proposal to add support for this

tustvold:
Closing in favour of https://github.com/apache/arrow-rs-object-store/pull/371

Issue #327, opened 2025-04-09
Some Inconsistencies in the Path and List


iodeal:
**Describe the bug**
I have found some inconsistencies in the path and list:
For path conversion, the tailing '/' will be stripped, which is problematic in some distributed file systems. As in S3, it is sensitive to the tailing `/`, so it is necessary to keep the tailing `/`.
It is precisely because the path of the distributed file system has a significant impact on the tail of the `/`, whether or not to add the tail of the `/` in the "list", especially in "list_with_delimiter". When the project is currently in the list, `/` will be directly added to the tail of the path, which leads to some problems, such as
When querying objects with the prefix "path/child" in S3, the expected result is objects such as child1, child2, etc. However, if we follow the previous logic and query objects with the prefix "path/child/", the result will be empty.

**To Reproduce**
upload path/child1,path/child2 to s3
`
store.list(Some("path/child"));
`

**Expected behavior**
Regarding the handling of Path:
Modify Path: Parse, From<&str>and From<String>related code, keeping the last `/` at the end
Processing of List
Modify the tail addition logic of the prefix in list_paginated. If the delimiter is true and the tail does not contain `/`, add `/` to the tail. Otherwise, leave it as it is.

tustvold:
This is intended behaviour and not a bug, if you describe your use-case I can perhaps advise on how to achieve it

iodeal:
> This is intended behaviour and not a bug, if you describe your use-case I can perhaps advise on how to achieve it.
> 
> We don't support paths with trailing / as they cause compatibility problems across stores, in particular those that are directory oriented.
Thanks! Perhaps I didn't describe it very clearly.
First of all, my use case is that in a file system similar to S3, if we want to list all objects with a prefix key, we cannot add '/' after this prefix key. Currently, it seems that this project will automatically add tailing `/`, which may not achieve the expected effect.
Secondly, for native file systems, the path being a directory does not depend on whether the given path has a tail `/`. For some distributed systems, whether the path is a directory or not is closely related to the tail `/`. So, for a path, whether the tail contains'/'depends on the caller, rather than automatically strip the tail, which I think is reasonable.

iodeal:
As documented on [ObjectStore::list](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#tymethod.list)
>List all the objects with the given prefix.
>
[...]

PR #328, opened 2025-04-09
fix bug: #327

iodeal:
fix bug: #327


tustvold:
As stated on linked ticket, this is not a bug

PR #329, opened 2025-04-10
Support `object_store` with wasm: Default wasm32-unknown-unknown HttpConnector

H-Plus-Time:

- Closes https://github.com/apache/arrow-rs/issues/7227.

Target-gated wasm32-unknown-unknown reqwest-based HttpConnector (with explicit handling surrounding non-Send, non-Sync futures).

wasm32-unknown-unknown users will need to add the `getrandom = { version = "*", features = ["js"] }` cargo entry if they aren't already using it. The alternative would be to feature flag in a similar manner to chrono, getrandom, zstd (i.e. 'js').

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,9 @@ jobs:
         run: cargo build --all-features --target wasm32-wasip1
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
       - name: Run wasm32-unknown-unknown tests (via Node)
         run: wasm-pack test --node --features http --no-default-features
 
--- a/src/client/body.rs
+++ b/src/client/body.rs
@@ -49,6 +49,14 @@ impl HttpRequestBody {
         }
     }
 
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    pub(crate) fn into_reqwest(self) -> reqwest::Body {
+        match self.0 {
+            Inner::Bytes(b) => b.into(),
+            Inner::PutPayload(_, payload) => Bytes::from(payload).into(),
+        }
+    }
+
     /// Returns true if this body is empty
     pub fn is_empty(&self) -> bool {
         match &self.0 {
--- a/src/client/connection.rs
+++ b/src/client/connection.rs
@@ -228,18 +228,18 @@ impl HttpService for reqwest::Client {
 #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
 impl HttpService for reqwest::Client {
     async fn call(&self, req: HttpRequest) -> Result<HttpResponse, HttpError> {
-        let (parts, body) = req.into_parts();
-        let url = parts.uri.to_string().parse().unwrap();
-        let mut req = reqwest::Request::new(parts.method, url);
-        *req.headers_mut() = parts.headers;
-        *req.body_mut() = Some(body.into_reqwest());
-
         use futures::{
             channel::{mpsc, oneshot},
             SinkExt, StreamExt, TryStreamExt,
         };
         use http_body_util::{Empty, StreamBody};
         use wasm_bindgen_futures::spawn_local;
+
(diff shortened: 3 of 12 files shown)


src/client/body.rs
```diff
@@ -49,6 +49,14 @@ impl HttpRequestBody {
         }
     }
 
+    #[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+    pub(crate) fn into_reqwest(self) -> reqwest::Body {
+        match self.0 {
+            Inner::Bytes(b) => b.into(),
+            Inner::PutPayload(_, payload) => Into::<Bytes>::into(payload).into(),
```
kylebarron, line 56: nit: Is this the same as `Bytes::from(payload).into()`? IMO that might read slightly easier.

H-Plus-Time: :facepalm: ah, of course.

alamb (commented):
Thank you @H-Plus-Time  and @kylebarron  -- this looks really neat

I think it is important to have some sort of test for this feature otherwise it is quite likely it will be broken in some future reactor

How did you test this locally? Perhaps we can adapt such a test to run in our CI?

H-Plus-Time:
I tested through object-store-wasm (which provided all the wasm-bindgen bindings to create object store instances and so on from JS), but a reasonable approach would be:
1. A single wasm-bindgen'd method that sets up a simple `HttpStore` pointing to localhost, and does a selection of simple operations (head, full get, ranged get) on a pre-defined object.
2. Spin up a http server on localhost (serving static files, specifically that pre-defined object).
3. Import, instantiate and run that wasm-bindgen'd module in a NodeJS script.

wasm-bindgen-test essentially does 1 and 3, a simple startup and teardown script for the server component would be all that's needed for 2.

It shouldn't be too much trouble, I'll get something basic together in the next few days.

alamb:
Thanks @H-Plus-Time  🙏 

(Note I will be out next week so I may be delated in responding)

H-Plus-Time:
That should do it I think - that one test covers a sizable chunk of the wasm-specific code; put/post requests would be the only other aspect worth testing I think.

.github/workflows/ci.yml
```diff
@@ -192,6 +192,10 @@ jobs:
         run: rustup target add wasm32-wasip1
       - name: Build wasm32-wasip1
         run: cargo build --all-features --target wasm32-wasip1
+      - name: Install wasm-pack
+        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
+      - name: Run wasm32-unknown-unknown tests (via Node)
```
alamb, line 197: nice

alamb:
kicked off the tests

alamb:
Looks like there are some CI failures: https://github.com/apache/arrow-rs-object-store/actions/runs/14495449031/job/40672770867?pr=329

alamb (approved):
Looks good to me -- thank you very much @H-Plus-Time

src/client/connection.rs
```diff
@@ -224,6 +224,60 @@ impl HttpService for reqwest::Client {
...
+        let mut req = reqwest::Request::new(parts.method, url);
+        *req.headers_mut() = parts.headers;
+        *req.body_mut() = Some(body.into_reqwest());
+
+        use futures::{
+            channel::{mpsc, oneshot},
+            SinkExt, StreamExt, TryStreamExt,
+        };
+        use http_body_util::{Empty, StreamBody};
+        use wasm_bindgen_futures::spawn_local;
```
kylebarron, line 242: Nit: move imports to the top of this scope? E.g. before the `req.into_parts` line?

kylebarron (approved):
Channels always confuse me, but overall looks good.

src/client/connection.rs
H-Plus-Time: Ah, yep - I must have started from a copy-paste of the non-wasm impl, I'll move them.

alamb:
Thanks again everyone. From my perspective this PR is ready to merge but I want to leave it open for another day or two to allow more time for comments. I'll be away next week so if another committer doesn't merge it before I'll merge it when I get back

tustvold:
Thank you for this, looks good to me :+1:

PR #333, opened 2025-04-15
feat: add EKS Pod Identity support (#282)

### LINKED ISSUE #282: Support EKS Pod Identity (alternative to IRSA)

I have a rust binary that makes use of object_store and I'd like to use it in AWS EKS with Pod Identity to access objects in S3. Maybe I'm missing something, but it seems like object_store's AWS S3 setup can't make use of the credentials provided by Pod Identity. Looking [here](https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html), I see the reference to these environment variables:

```
AWS_ACCESS_KEY_ID -> access_key_id
AWS_SECRET_ACCESS_KEY -> secret_access_key
AWS_DEFAULT_REGION -> region
AWS_ENDPOINT -> endpoint
AWS_SESSION_TOKEN -> token
AWS_CONTAINER_CREDENTIALS_RELATIVE_URI -> https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html
AWS_ALLOW_HTTP -> set to “true” to permit HTTP connections without TLS
```

Pod Identity makes use of different variables [per this documentation](https://docs.aws.amazon.com/eks/latest/userguide/pod-id-how-it-works.html):

```
AWS_CONTAINER_CREDENTIALS_FULL_URI=http://169.254.170.23/v1/credentials
AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE=/var/run/secrets/pods.eks.amazonaws.com/serviceaccount/eks-pod-identity-token
```

Am I missing something or is it currently not possible to make use of these credentials with object_store?

I'd like to be able to use the Pod Identity credentials with object_store without having to do any additional workarounds.

I would like to avoid putting an access key and secret in a Kubernetes secret and mounting that to the containers. Other than that, I'm looking for a way to minimally adapt my code with something that can use the container credentials URI and auth token file to authenticate with object_store's S3 store.

**Additional context**
I'm happy to contribute to the project, although I would need a bit of guidance to understand how to adapt the library for this issue.

andreasbros:

Closes #282 

# Rationale for this change
 
This PR extends the `AmazonS3Builder` so that it recognises and supports EKS Pod Identity credentials using the two environment variables:

- `AWS_CONTAINER_CREDENTIALS_FULL_URI`

- `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`

Previously, the builder only considered ECS task credentials (`AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`), instance metadata, static credentials, or web identity tokens. Adding EKS Pod Identity support aligns it with modern Kubernetes IRSA setups, allowing pods to retrieve AWS credentials from an EKS endpoint without needing to mount AWS credentials directly.

1. **New Config Keys**
Adds `AmazonS3ConfigKey::ContainerCredentialsFullUri` and `AmazonS3ConfigKey::ContainerAuthorizationTokenFile` to the config-based approach, for parsing EKS Pod Identity settings. The builder picks these keys from environment variables `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`.

2. **EKSPodCredentialProvider**
Introduces an `EKSPodCredentialProvider`, which is constructed when both config keys are set. It uses a bearer token (read from the specified file) to fetch short-lived credentials from the EKS credential endpoint.

3. **Builder Logic**
Adjusts `AmazonS3Builder::build` to give priority to EKS credentials if both the full URI and token file are specified. It checks environment variables in `from_env` or direct calls to `with_config`, falling back to ECS or instance metadata if EKS variables are absent.

4. **Tests**
Adds tests to confirm EKS credentials build and provider.

- **New EKS Credential Support**
Users in EKS can now set `AWS_CONTAINER_CREDENTIALS_FULL_URI` and `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`, and the builder automatically fetches credentials.

- **Configuration Keys**
Two new config keys are recognised by the builder:

  - `AmazonS3ConfigKey::ContainerCredentialsFullUri`
  - `AmazonS3ConfigKey::ContainerAuthorizationTokenFile`

These changes are backwards-compatible: existing ECS, static credentials, or IMDS-based setups continue to work unchanged. No additional user steps are required unless they specifically opt to use EKS Pod Identity.

### DIFF
--- a/src/aws/credential.rs
+++ b/src/aws/credential.rs
@@ -759,8 +759,17 @@ async fn eks_credential(
     url: &str,
     token_file: &str,
 ) -> Result<TemporaryToken<Arc<AwsCredential>>, StdError> {
-    let token = std::fs::read_to_string(token_file)
-        .map_err(|e| format!("Failed to read EKS token file '{token_file}': {e}"))?;
+    // Spawn IO to blocking tokio pool if running in tokio context
+    let token = match tokio::runtime::Handle::try_current() {
+        Ok(runtime) => {
+            let path = token_file.to_string();
+            runtime
+                .spawn_blocking(move || std::fs::read_to_string(&path))
+                .await?
+        }
+        Err(_) => std::fs::read_to_string(token_file),
+    }
+    .map_err(|e| format!("Failed to read EKS token file '{token_file}': {e}"))?;
 
     let mut req = client.request(Method::GET, url);
     req = req.header("Authorization", token);
--- a/src/aws/credential.rs
+++ b/src/aws/credential.rs
@@ -719,6 +719,64 @@ async fn task_credential(
     })
 }
 
+/// EKS Pod Identity credential provider.
+///
+/// Uses the endpoint in `AWS_CONTAINER_CREDENTIALS_FULL_URI`
+/// and the bearer token in `AWS_CONTAINER_AUTHORIZATION_TOKEN_FILE`
+/// to fetch ephemeral AWS credentials from an EKS pod.
+#[derive(Debug)]
+pub(crate) struct EKSPodCredentialProvider {
+    pub url: String,
+    pub token_file: String,
+    pub retry: RetryConfig,
+    pub client: HttpClient,
+    pub cache: TokenCache<Arc<AwsCredential>>,
+}
+
+#[async_trait]
+impl CredentialProvider for EKSPodCredentialProvider {
+    type Credential = AwsCredential;
+
+    async fn get_credential(&self) -> Result<Arc<AwsCredential>> {
+        self.cache
+            .get_or_insert_with(|| {
+                eks_credential(&self.client, &self.retry, &self.url, &self.token_file)
+            })
+            .await
+            .map_err(|source| crate::Error::Generic {
+                store: STORE,
+                source,
+            })
+    }
+}
+
+/// Performs the actual credential retrieval and parsing for `EKSPodCredentialProvider`.
+///
+/// <https://mozillazg.com/2023/12/security-deep-dive-into-aws-eks-pod-identity-feature-en.html>
+async fn eks_credential(
+    client: &HttpClient,
+    retry: &RetryConfig,
+    url: &str,
+    token_file: &str,
+) -> Result<TemporaryToken<Arc<AwsCredential>>, StdError> {
+    let token = std::fs::read_to_string(token_file)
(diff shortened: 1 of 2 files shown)


alamb (commented):
Thank you @andreasbros 

I am not familiar with this AWS feature, but this PR looks very well commented, tested and follows the existing patterns in the crate. It looks good to me 👏 

It seems there is a small error when compiling for wasm that the CI found, but otherwise this PR is good to merge from my perspective

src/aws/credential.rs
```diff
@@ -719,6 +720,68 @@ async fn task_credential(
...
+
+/// Performs the actual credential retrieval and parsing for `EKSPodCredentialProvider`.
+///
+/// <https://mozillazg.com/2023/12/security-deep-dive-into-aws-eks-pod-identity-feature-en.html>
+async fn eks_credential(
+    client: &HttpClient,
+    retry: &RetryConfig,
+    url: &str,
+    token_file: &str,
+) -> std::result::Result<TemporaryToken<Arc<AwsCredential>>, Box<dyn std::error::Error + Send + Sync>>
```
tustvold, line 762:
```suggestion
) -> Result<TemporaryToken<Arc<AwsCredential>>, StdError>
```

tustvold (dismissed):
This looks good to me, I think it just needs the WASM build to be fixed (likely by just disabling this functionality on WASM platforms)

andreasbros:
Thanks @alamb and @tustvold for checking, and indeed WASM build was failing, it was because I used async `tokio::fs`, I now changed it to `std::fs` and this should resolve the issue, just waiting for GHA workflows to get approved and executed.

andreasbros:
@tustvold I changed to `std::fs` based on existing `web_identity` implementation, but I agree it should be async ideally:
https://github.com/apache/arrow-rs-object-store/blob/main/src/aws/credential.rs#L639

> it just needs to be appropriately gated
My understanding from the README doc is that entire feature `aws` is not available in WASM.
Can you point me to how to gate it?

tustvold:
I took the liberty of pushing a quick fix to avoid doing blocking IO within the credential provider, using the same trick we use for LocalFilesystem (which is also what tokio::fs does under the hood).

I don't have a means to test this, but if you are happy it is working I'm happy to get this merged

andreasbros:
> I took the liberty of pushing a quick fix
thanks @tustvold

PR #334, opened 2025-04-15
Fix query parameter signing in Azure

### LINKED ISSUE #320: Azure Gen2 broken on latest
**Describe the bug**

I've been investigating some Vortex queries on Azure Blob Store Gen2. I've created a simple storage account, populated some files, and was just trying to list the bucket on 0.12

Cargo.toml:

```toml
[package]
name = "rustazuretest"
version = "0.1.0"
edition = "2024"

[dependencies]
tokio = { version = "1", features = ["full"] }
object_store = { version = "=0.12", features = ["azure"] }
futures = "0.3"
```

main.rs
```rust
use futures::StreamExt;
use object_store::ObjectStore;

#[tokio::main]
async fn main() {
    let account = "vortexicebergsummit25";
    let access_key = "REDACTED";
    let container_name = "tpch";

    let azure = MicrosoftAzureBuilder::new()
        .with_account(account)
        .with_access_key(access_key)
        .with_container_name(container_name)
        .build()
        .expect("azure object store");

    let mut objects = azure.list(None);
    while let Some(next) = objects.next().await {
        println!("Object: {:?}", next.unwrap());
    }
}
```

On 0.12 this results in an error:

```
called `Result::unwrap()` on an `Err` value: Generic { store: "MicrosoftAzure", source: ListRequest { source: RetryError { method: GET, uri: Some(https://vortexicebergsummit25.blob.core.windows.net/tpch?&restype=container&comp=list), retries: 0, max_retries: 10, elapsed: 61.271375ms, retry_timeout: 180s, inner: Status { status: 403, body: Some("\u{feff}<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.\nRequestId:1fe66e1c-a01e-0000-2a10-a476a8000000\nTime:2025-04-02T20:45:18.2401813Z</Message><AuthenticationErrorDetail>The MAC signature found in the HTTP request 'REDACTED' is not the same as any computed signature. Server used following string to sign: 'GET\n\n\n\n\n\nWed, 02 Apr 2025 20:45:18 GMT\n\n\n\n\n\nx-ms-version:2023-11-03\n/vortexicebergsummit25/tpch\n:\ncomp:list\nrestype:container'.</AuthenticationErrorDetail></Error>") } } } }
stack backtrace:
...
```

However, if I forcibly pin to 0.11.2, it works

```
[package]
name = "rustazuretest"
version = "0.1.0"
edition = "2024"

[dependencies]
tokio = { version = "1", features = ["full"] }
object_store = { version = "=0.11.2", features = ["azure"] }
futures = "0.3"
```

Image

AdamGS:

- Closes #320 

# Rationale for this change
https://github.com/apache/arrow-rs/pull/7183 Changed the way the query string is serialized from:
```
azurebase.com/path?comp=list&restype:container
```
to:
```
azurebase.com/path?&comp=list&restype:container # Note the leading & after the ?
```

That data is encoded into `application/x-www-form-urlencoded` form, which is parsed according to [this](https://url.spec.whatwg.org/#urlencoded-parsing) spec. The `Url` crate ignores empty values (correctly) when parsing the underlying string in `Url::query_pairs`, but that means we sign the "wrong" string. The fix here basically tells is to serialize the query params as a suffix to the path, which is assumed to be encoded correctly.

I really think this behavior is mostly an Azure oddity/bug, but its much easier to fix it here than it is to get to the right person there. 

 

1. Fixes a bug in the request builder when adding query params to a request that already has a path component.
2. Some minor touchups to the `CONTRIBUTING.md` which seemed to reference some env variables that are no longer in use.

No changes to public APIs, but if a user relayed on this pretty obscure behavior, their code will break. Its important to note that this change brings the crate back to the same behavior it had in `0.11.2`.

### DIFF
--- a/src/client/builder.rs
+++ b/src/client/builder.rs
@@ -178,7 +178,7 @@ impl HttpRequestBuilder {
         self
     }
 
-    #[cfg(any(feature = "aws", feature = "gcp", feature = "azure"))]
+    #[cfg(any(test, feature = "aws", feature = "gcp", feature = "azure"))]
     pub(crate) fn query<T: serde::Serialize + ?Sized>(mut self, query: &T) -> Self {
         let mut error = None;
         if let Ok(ref mut req) = self.request {
@@ -252,12 +252,18 @@ where
 
     let mut out = match parts.path_and_query {
         Some(p) => match p.query() {
-            Some(x) => format!("{}?{}", p.path(), x),
+            Some(query) => format!("{}?{}", p.path(), query),
             None => format!("{}?", p.path()),
         },
         None => "/?".to_string(),
     };
-    let mut serializer = form_urlencoded::Serializer::new(&mut out);
+    let mut serializer = if out.ends_with('?') {
+        let start_position = out.len();
+        form_urlencoded::Serializer::for_suffix(&mut out, start_position)
+    } else {
+        form_urlencoded::Serializer::new(&mut out)
+    };
+
     serializer.extend_pairs(query_pairs);
 
     parts.path_and_query = Some(out.try_into().unwrap());
@@ -270,7 +276,10 @@ mod tests {
 
     #[test]
     fn test_add_query_pairs() {
-        let mut uri = Uri::from_static("https://foo@example.com/bananas?foo=1");
+        let mut uri = Uri::from_static("https://foo@example.com/bananas");
+
+        add_query_pairs(&mut uri, [("foo", "1")]);
+        assert_eq!(uri.to_string(), "https://foo@example.com/bananas?foo=1");
 
         add_query_pairs(&mut uri, [("bingo", "foo"), ("auth", "test")]);
         assert_eq!(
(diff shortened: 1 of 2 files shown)


src/client/builder.rs
```diff
@@ -183,7 +183,8 @@ impl HttpRequestBuilder {
         let mut error = None;
         if let Ok(ref mut req) = self.request {
             let mut out = format!("{}?", req.uri().path());
-            let mut encoder = form_urlencoded::Serializer::new(&mut out);
+            let start_position = out.len();
```
alamb, line 186: > [apache/arrow-rs#7183](https://github.com/apache/arrow-rs/pull/7183) Changed the way the query string is serialized from:
> 
> ```
> azurebase.com/path?comp=list&restype:container
> ```
> 
> to:
> 
> ```
> azurebase.com/path?&comp=list&restype:container # Note the leading & after the ?
> ```

Does this change remove the  `&` after `?`. I am having a hard time reasoning about the change this behavior makes. 

Is it possible to write some unit tests or something that can show the before/after behavior (and prevent this code from being accidentally broken again)?

alamb (commented):
Thank you so much for debugging this @AdamGS 🕵️ 👌 

If we could figure out some way to add a test for this change to avoid future regressions that would be really great

alamb:
FYI @crepererum and @tustvold since you approved / authored https://github.com/apache/arrow-rs/pull/7183  I would appreciate it if you could also give this PR a look

src/client/builder.rs
AdamGS: yeah, after this change the URL will again be `azurebase.com/path?comp=list&restype:container`. 

Part of the problem here is that I realized that it seems like [Azurite behaves differently](https://github.com/Azure/Azurite/issues/2548) than Azure in this case, so tests that will fail on the real Azure (like @a10y's original repro) would pass in the repo's test suite.

AdamGS:
merged `main` so we can run tests here.

src/client/builder.rs
alamb: Maybe a unit test then?

src/client/builder.rs
```diff
@@ -252,12 +252,18 @@ where
...
         },
         None => "/?".to_string(),
     };
-    let mut serializer = form_urlencoded::Serializer::new(&mut out);
+    let mut serializer = if out.ends_with('?') {
+        let start_position = out.len();
+        form_urlencoded::Serializer::for_suffix(&mut out, start_position)
+    } else {
+        form_urlencoded::Serializer::new(&mut out)
+    };
```
AdamGS, line 265: This code path had the same kind of bug, modified the relevant test to catch it.

crepererum:
I might have missed this, but can someone test manually that this fixes the actual bug by using a real azure account? Maybe you already did that, in which case I just want a confirmation that this was the case.

crepererum (approved):
Alright, then I think we're good to go here.

Thanks for going the extra mile of fixing the contribution guideline and filing an azurite upstream ticket.

Issue #337, opened 2025-04-16
Incorrect token sent as part of url signing function.


jackm-mimica:
**Describe the bug**
Creating a signed url for a gcp blob throws the error
```
Generic GCS error: Error performing token request: Error performing GET https://oauth2.googleapis.com/tokeninfo?&access_token=<TOKEN>
```
The  <TOKEN> is prefixed 1%2F%2F which I believe means it is a refresh token not the expected actual token.

**To Reproduce**
[Added a test here:](https://github.com/jackm-mimica/rs-object-store-arrow/commit/431c232d3de3f1dad8b74eeae4b7952719db0e74)

**Expected behavior**
I would expect the library to get the access_token not the refresh token from my ADC

**Additional context**
I don't think this would be caught in the integration test against fake gcs as oauth is marked as disabled for the tests. Specific to GCP.

jackm-mimica:
Fix here https://github.com/jackm-mimica/rs-object-store-arrow/commit/6ba38fa367f7143c213ae4b1153507add56bd8d7

PR #338, opened 2025-04-16
Fix GCP signing token

jackm-mimica:
* refresh token replaced by access token

Closes [#](https://github.com/apache/arrow-rs-object-store/issues/337).

Changes how the GCP Signing Auth retrieves the email, originally the token info endpoint was used with a refresh token whether behavior has changed on googles side or that was never functional is out of scope but now it generates a access_token using ADC then uses the access token to retrieve the email.

No.

### DIFF
--- a/src/gcp/credential.rs
+++ b/src/gcp/credential.rs
@@ -222,6 +222,7 @@ struct TokenClaims<'a> {
 struct TokenResponse {
     access_token: String,
     expires_in: u64,
+    id_token: Option<String>,
 }
 
 /// Self-signed JWT (JSON Web Token).
@@ -603,6 +604,11 @@ struct EmailResponse {
     email: String,
 }
 
+#[derive(Debug, Deserialize)]
+struct IdTokenClaims {
+    email: String,
+}
+
 async fn get_token_response(
     client_id: &str,
     client_secret: &str,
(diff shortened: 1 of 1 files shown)


tustvold:
This feature was originally added in https://github.com/apache/arrow-rs/pull/5300 by @l1nxy, I tested it at the time and it worked as advertised. @l1nxy perhaps you might be able to test out this PR and confirm it doesn't break your use-case?

I'm in two minds as to whether to consider this a breaking change, before it didn't actually use the client_secret, etc... but I think we can probably get this in under a minor release.

l1nxy:
After reviewing the document, I'm unsure if it's correct. When the "grant_type" is "refresh_token", does the response include the access_token field? Could you provide some clarification or documentation?

doc:https://google.aip.dev/auth/4113

jackm-mimica:
Thanks for your question about the refresh_token grant type.

To clarify: The tokeninfo endpoint (https://oauth2.googleapis.com/tokeninfo) is designed to validate and return information about an access token, not a refresh token. When you use a refresh_token grant type, you would typically make a request to the token endpoint (https://oauth2.googleapis.com/token), which then returns a new access_token (and potentially a new refresh_token depending on configuration).

The document you referenced (google.aip.dev/auth/4113) covers guidelines for OAuth authentication flows, but doesn't specifically address the behavior of the tokeninfo endpoint with respect to refresh tokens.

The change I implemented maintains this distinction between endpoints - tokeninfo for validating access tokens, and the token endpoint for operations involving refresh tokens.

Does this address your question, or were you referring to a different aspect of the authentication flow?

l1nxy:
You are right, LGTM, thanks for your contribution!

jackm-mimica:
I am questioning the validity of the fix though, I should be able to use the id_token on the token response and just parse the email from the id_token jwt this would eliminate an extra api call.

tustvold:
IIRC the id_token is guaranteed to be present

tustvold:
Perhaps we could look at what the Golang SDK does here?

jackm-mimica:
I can't seem to unpick what is actually happening there is one line of documentation that refers to this problem

> [Detecting GoogleAccessID may not be possible if you are authenticated using a token source or using option.WithHTTPClient. In this case, you can provide a service account email for GoogleAccessID and the client will attempt to sign the URL or Post Policy using that service account. 
](https://pkg.go.dev/cloud.google.com/go/storage#hdr-Credential_requirements_for_signing)

I have added a commit that will parse the id_token if present otherwise will fall back to the original tokeninfo endpoint.

tustvold (approved):

PR #346, opened 2025-05-02
chore: Add anda_object_store to README

zensh:
[anda_object_store](https://github.com/ldclabs/anda-db/blob/main/rs/anda_object_store/README.md) is a Rust library that extends the functionality of the [object_store](https://docs.rs/object_store) crate with additional features like metadata management and encryption.

Integration Test: 

- MetaStore: https://github.com/ldclabs/anda-db/blob/main/rs/anda_object_store/src/lib.rs#L577
- EncryptedStore: https://github.com/ldclabs/anda-db/blob/main/rs/anda_object_store/src/encryption.rs#L902

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,13 +52,16 @@ cargo build -p object_store --target wasm32-unknown-unknown
 
 Here are several related crates in different repositories
 
-| Crate                    | Description                                 | Documentation                           |
-| ------------------------ | ------------------------------------------- | --------------------------------------- |
-| [`object_store_opendal`] | Use [`opendal`] as [`object_store`] backend | [(README)][object_store_opendal-readme] |
+| Crate                    | Description                                                           | Documentation                           |
+| ------------------------ | --------------------------------------------------------------------- | --------------------------------------- |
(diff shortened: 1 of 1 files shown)


alamb (commented):
Thank you @zensh the anda_object_store crate looks neat

Currently the crates we list on this page are from other Apache projects (in this case OpenDAL https://opendal.apache.org/). I think it is great to list other crates maintained by the community in general, but we need to make it clear they are not official Apache project

Can you please move the link from `anda_object_store` to a new table under a heading that makes this clear? Something like:

## Community Extensions
There following community maintained crates provide additional functionality for `object_store` and are NOT governed by the Apache Software Foundation. We list them below in the hope they may be useful, but they are not official Apache projects or endorsed by the Apache Arrow project.

(table here)



Perhaps you can also add a link to this project that provides HDFS connectivity as well:  https://github.com/datafusion-contrib/hdfs-native-object-store

zensh:
@alamb Thank you for your suggestion. I have updated this PR.

alamb (approved):
Thank you @zensh  -- this looks great to me

Issue #259, opened 2023-05-01
Ability to get actually full path/URL, e.g. from ObjectMeta


ZetaTwo:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

It would be nice to have a function/attribute to give me the full URL for an object. Currently, the ObjectMeta.location provides the path _within_ the object store but I would like to get the full URL such as "gs://bucket/dir/file.txt".

**Describe the solution you'd like**

Maybe add an attribute "uri" to the ObjectMeta struct or provide a new function `.get_uri` on the ObjectStore struct.

**Additional context**

The reason I am looking for this is that I need to pass the URL a different program and it would be nice to keep all the logic centralised in the object_store library instead of having a separate function with "if gcp: "gs://" + path elif aws: "s3://" + path" etc.

PR #348, opened 2025-05-06
Add `ObjectStoreRegistry`

### LINKED ISSUE #347: Is there a way to go from `ObjectStore` to `(URL, opts)`?

ObjectStore's parse_url.

**Describe your question**

We're looking at converting between `ObjectStore`s and URLs in SlateDB:

https://github.com/slatedb/slatedb/pull/533#issuecomment-2848730594

URL to `ObjectStore` seems pretty straight forward with `parse_url`, but it's  unclear how we might go from `ObjectStore` to URL. Is there some `get_url()` method we can call on the `ObjectStore` trait? I poked around but didn't find anything. I suspect I'm missing something obvious.

**Additional context**

criccomini:
This change introduces an `ObjectStoreRegistry`. The rationale is that there are some cases where `ObjectStore`s need to be resolved at runtime. There already exists `parse_url` to address this need, but it's fairly hard-coded. It doesn't support custom `ObjectStore` implementations and parsing logic, for example.

`ObjectStoreRegistry` can also act as a nice cache when multiple object stores are used across the codebase. Moreover, it can serve as a way to map URLs to object stores--something that was requested in #347 and #259. Since `object_store` does not currently expose a `get_url()` method, this is useful in cases where a user wants to go from an `ObjectStore` back to a URL.

Fixes #347

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -1,0 +1,302 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+//! ObjectStoreRegistry holds all the object stores at Runtime with a scheme for each store.
+//! This allows the user to extend DataFusion with different storage systems such as S3 or HDFS
+//! and query data inside these systems.
+
+use dashmap::DashMap;
+use datafusion_common::{exec_err, DataFusionError, Result};
+#[cfg(not(target_arch = "wasm32"))]
+use object_store::local::LocalFileSystem;
+use object_store::ObjectStore;
+use std::sync::Arc;
+use url::Url;
+
+/// A parsed URL identifying a particular [`ObjectStore`] instance
+///
+/// For example:
+/// * `file://` for local file system
+/// * `s3://bucket` for AWS S3 bucket
+/// * `oss://bucket` for Aliyun OSS bucket
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct ObjectStoreUrl {
+    url: Url,
+}
+
+impl ObjectStoreUrl {
+    /// Parse an [`ObjectStoreUrl`] from a string
+    ///
+    /// # Example
+    /// ```
+    /// # use url::Url;
+    /// # use datafusion_execution::object_store::ObjectStoreUrl;
+    /// let object_store_url = ObjectStoreUrl::parse("s3://bucket").unwrap();
+    /// assert_eq!(object_store_url.as_str(), "s3://bucket/");
+    /// // can also access the underlying `Url`
+    /// let url: &Url = object_store_url.as_ref();
+    /// assert_eq!(url.scheme(), "s3");
+    /// assert_eq!(url.host_str(), Some("bucket"));
+    /// assert_eq!(url.path(), "/");
+    /// ```
+    pub fn parse(s: impl AsRef<str>) -> Result<Self> {
+        let mut parsed =
+            Url::parse(s.as_ref()).map_err(|e| DataFusionError::External(Box::new(e)))?;
+
+        let remaining = &parsed[url::Position::BeforePath..];
+        if !remaining.is_empty() && remaining != "/" {
+            return exec_err!(
+                "ObjectStoreUrl must only contain scheme and authority, got: {remaining}"
+            );
+        }
+
+        // Always set path for consistency
+        parsed.set_path("/");
+        Ok(Self { url: parsed })
+    }
+
+    /// An [`ObjectStoreUrl`] for the local filesystem (`file://`)
+    ///
+    /// # Example
+    /// ```
+    /// # use datafusion_execution::object_store::ObjectStoreUrl;
+    /// let local_fs = ObjectStoreUrl::parse("file://").unwrap();
+    /// assert_eq!(local_fs, ObjectStoreUrl::local_filesystem())
+    /// ```
+    pub fn local_filesystem() -> Self {
+        Self::parse("file://").unwrap()
+    }
+
+    /// Returns this [`ObjectStoreUrl`] as a string
+    pub fn as_str(&self) -> &str {
+        self.as_ref()
+    }
+}
+
+impl AsRef<str> for ObjectStoreUrl {
+    fn as_ref(&self) -> &str {
+        self.url.as_ref()
+    }
+}
+
+impl AsRef<Url> for ObjectStoreUrl {
+    fn as_ref(&self) -> &Url {
+        &self.url
+    }
+}
+
+impl std::fmt::Display for ObjectStoreUrl {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        self.as_str().fmt(f)
+    }
+}
+
+/// [`ObjectStoreRegistry`] maps a URL to an [`ObjectStore`] instance,
+/// and allows DataFusion to read from different [`ObjectStore`]
+/// instances. For example DataFusion might be configured so that
+///
+/// 1. `s3://my_bucket/lineitem/` mapped to the `/lineitem` path on an
+///    AWS S3 object store bound to `my_bucket`
+///
+/// 2. `s3://my_other_bucket/lineitem/` mapped to the (same)
+///    `/lineitem` path on a *different* AWS S3 object store bound to
+///    `my_other_bucket`
+///
+/// When given a [`ListingTableUrl`], DataFusion tries to find an
+/// appropriate [`ObjectStore`]. For example
+///
+/// ```sql
+/// create external table unicorns stored as parquet location 's3://my_bucket/lineitem/';
+/// ```
+///
+/// In this particular case, the url `s3://my_bucket/lineitem/` will be provided to
+/// [`ObjectStoreRegistry::get_store`] and one of three things will happen:
+///
+/// - If an [`ObjectStore`] has been registered with [`ObjectStoreRegistry::register_store`] with
+///   `s3://my_bucket`, that [`ObjectStore`] will be returned
+///
+/// - If an AWS S3 object store can be ad-hoc discovered by the url `s3://my_bucket/lineitem/`, this
+///   object store will be registered with key `s3://my_bucket` and returned.
+///
+/// - Otherwise an error will be returned, indicating that no suitable [`ObjectStore`] could
+///   be found
+///
+/// This allows for two different use-cases:
+///
+/// 1. Systems where object store buckets are explicitly created using DDL, can register these
+///    buckets using [`ObjectStoreRegistry::register_store`]
+///
+/// 2. Systems relying on ad-hoc discovery, without corresponding DDL, can create [`ObjectStore`]
+///    lazily by providing a custom implementation of [`ObjectStoreRegistry`]
+///
+/// <!-- is in a different crate so normal rustdoc links don't work -->
+/// [`ListingTableUrl`]: https://docs.rs/datafusion/latest/datafusion/datasource/listing/struct.ListingTableUrl.html
+/// [`ObjectStore`]: object_store::ObjectStore
+pub trait ObjectStoreRegistry: Send + Sync + std::fmt::Debug + 'static {
+    /// If a store with the same key existed before, it is replaced and returned
+    fn register_store(
+        &self,
+        url: &Url,
+        store: Arc<dyn ObjectStore>,
+    ) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Get a suitable store for the provided URL. For example:
+    ///
+    /// - URL with scheme `file:///` or no scheme will return the default LocalFS store
+    /// - URL with scheme `s3://bucket/` will return the S3 store
+    /// - URL with scheme `hdfs://hostname:port/` will return the hdfs store
+    ///
+    /// If no [`ObjectStore`] found for the `url`, ad-hoc discovery may be executed depending on
+    /// the `url` and [`ObjectStoreRegistry`] implementation. An [`ObjectStore`] may be lazily
+    /// created and registered.
+    fn get_store(&self, url: &Url) -> Result<Arc<dyn ObjectStore>>;
+}
+
+/// The default [`ObjectStoreRegistry`]
+pub struct DefaultObjectStoreRegistry {
+    /// A map from scheme to object store that serve list / read operations for the store
+    object_stores: DashMap<String, Arc<dyn ObjectStore>>,
+}
+
+impl std::fmt::Debug for DefaultObjectStoreRegistry {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        f.debug_struct("DefaultObjectStoreRegistry")
+            .field(
+                "schemes",
+                &self
+                    .object_stores
+                    .iter()
+                    .map(|o| o.key().clone())
+                    .collect::<Vec<_>>(),
+            )
+            .finish()
+    }
+}
+
+impl Default for DefaultObjectStoreRegistry {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl DefaultObjectStoreRegistry {
+    /// This will register [`LocalFileSystem`] to handle `file://` paths
+    #[cfg(not(target_arch = "wasm32"))]
+    pub fn new() -> Self {
+        let object_stores: DashMap<String, Arc<dyn ObjectStore>> = DashMap::new();
+        object_stores.insert("file://".to_string(), Arc::new(LocalFileSystem::new()));
+        Self { object_stores }
+    }
+
+    /// Default without any backend registered.
+    #[cfg(target_arch = "wasm32")]
+    pub fn new() -> Self {
+        let object_stores: DashMap<String, Arc<dyn ObjectStore>> = DashMap::new();
+        Self { object_stores }
+    }
+}
+
+///
+/// Stores are registered based on the scheme, host and port of the provided URL
+/// with a [`LocalFileSystem::new`] automatically registered for `file://` (if the
+/// target arch is not `wasm32`).
+///
+/// For example:
+///
+/// - `file:///my_path` will return the default LocalFS store
+/// - `s3://bucket/path` will return a store registered with `s3://bucket` if any
+/// - `hdfs://host:port/path` will return a store registered with `hdfs://host:port` if any
+impl ObjectStoreRegistry for DefaultObjectStoreRegistry {
+    fn register_store(
+        &self,
+        url: &Url,
+        store: Arc<dyn ObjectStore>,
+    ) -> Option<Arc<dyn ObjectStore>> {
+        let s = get_url_key(url);
+        self.object_stores.insert(s, store)
+    }
+
+    fn get_store(&self, url: &Url) -> Result<Arc<dyn ObjectStore>> {
+        let s = get_url_key(url);
+        self.object_stores
+            .get(&s)
+            .map(|o| Arc::clone(o.value()))
+            .ok_or_else(|| {
+                DataFusionError::Internal(format!(
+                    "No suitable object store found for {url}. See `RuntimeEnv::register_object_store`"
+                ))
+            })
+    }
+}
+
+/// Get the key of a url for object store registration.
+/// The credential info will be removed
+fn get_url_key(url: &Url) -> String {
+    format!(
+        "{}://{}",
+        url.scheme(),
+        &url[url::Position::BeforeHost..url::Position::AfterPort],
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_object_store_url() {
+        let file = ObjectStoreUrl::parse("file://").unwrap();
+        assert_eq!(file.as_str(), "file:///");
+
+        let url = ObjectStoreUrl::parse("s3://bucket").unwrap();
+        assert_eq!(url.as_str(), "s3://bucket/");
+
+        let url = ObjectStoreUrl::parse("s3://username:password@host:123").unwrap();
+        assert_eq!(url.as_str(), "s3://username:password@host:123/");
+
+        let err = ObjectStoreUrl::parse("s3://bucket:invalid").unwrap_err();
+        assert_eq!(err.strip_backtrace(), "External error: invalid port number");
+
+        let err = ObjectStoreUrl::parse("s3://bucket?").unwrap_err();
+        assert_eq!(err.strip_backtrace(), "Execution error: ObjectStoreUrl must only contain scheme and authority, got: ?");
+
+        let err = ObjectStoreUrl::parse("s3://bucket?foo=bar").unwrap_err();
+        assert_eq!(err.strip_backtrace(), "Execution error: ObjectStoreUrl must only contain scheme and authority, got: ?foo=bar");
+
+        let err = ObjectStoreUrl::parse("s3://host:123/foo").unwrap_err();
+        assert_eq!(err.strip_backtrace(), "Execution error: ObjectStoreUrl must only contain scheme and authority, got: /foo");
+
+        let err =
+            ObjectStoreUrl::parse("s3://username:password@host:123/foo").unwrap_err();
+        assert_eq!(err.strip_backtrace(), "Execution error: ObjectStoreUrl must only contain scheme and authority, got: /foo");
+    }
+
+    #[test]
+    fn test_get_url_key() {
+        let file = ObjectStoreUrl::parse("file://").unwrap();
+        let key = get_url_key(&file.url);
+        assert_eq!(key.as_str(), "file://");
+
+        let url = ObjectStoreUrl::parse("s3://bucket").unwrap();
+        let key = get_url_key(&url.url);
+        assert_eq!(key.as_str(), "s3://bucket");
+
+        let url = ObjectStoreUrl::parse("s3://username:password@host:123").unwrap();
+        let key = get_url_key(&url.url);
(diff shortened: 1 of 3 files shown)


criccomini:
Couple of open questions:

- How do y'all feel about the docs right now? They're just copied from DF.
- WDYT about adding a `get_url` method that takes an object store and compares refs to find the URL for it in the registry.
- I used `dashmap` since that's what DF was using. Not sure how you feel about adding the dependency.
- One of the tests uses a clunk debug fmt thingy to verify the returned instance is of type LocalFileSystem. I couldn't figure out the right way to do this in rust. o4-mini-high failed me.

criccomini:
Looks like we would need to drop dashmap to 5.5.2 (https://github.com/xacrimon/dashmap/blob/v5.5.2/Cargo.toml) to maintain the MSRV you guys have set (1.64). 6.1.0 requires 1.6.5 rust.

tustvold (commented):
I suggest we gate this by a feature flag given the additional dependency.

I think we also probably want to see how DF might switch to using this formulation, perhaps file a DF issue requesting help if you don't have capacity? I think we should make sure what we're adding here can be used downstream.

rtyler:
Golly this code looks familiar! I don't know if this was inspired by what we do in [delta-rs](https://github.com/delta-io/delta-rs/blob/main/crates/core/src/logstore/factories.rs) or if this is [convergent evolution](https://en.wikipedia.org/wiki/Carcinisation! :crab: ! 

Either way, I think this is a great improvement and the shedding of the dashmap dependency shouldn't be necessary based on the code paths I have seen using our implementation of the same concept.  I think this would be a positive change and allow me to shed some code as well :smile:

criccomini:
I've replaced `dashmap` with an RWLock on a HashMap here: https://github.com/apache/arrow-rs-object-store/pull/348/commits/81a05b0f770009260b9bad7b54115eefccd2b712

criccomini:
@rtyler Does this PR look like it would meet your needs (i.e. that you could remove your own registry)?

criccomini:
I opened a DataFusion github issue to ask if they'll adopt this version: https://github.com/apache/datafusion/issues/15983

rtyler:
@criccomini I think the approach here could, but as written thus far it only could replace use of `parse_url` whereas we're pretty heavily reliant on `parse_url_opts`, since object store options need to be plumbed through. I think that minor extension to the trait would make our factories code redundant

criccomini:
@rtyler Gotcha. I was a bit hesitant to add `_opts` support since I figured folks would instead implement their own `ObjectStoreRegistry` rather than use the default. It's easy enough to add, though. And since `_opts` is part of the `object_store` API, I suppose it's not that bad to support. It just felt funny to me since you have to supply the opts every time (unless you know it's a store is already cached for a URL, and decide to skip passing in the opts).

criccomini:
@rtyler I've pushed an update that contains a `get_store_with_opts` method. The default implementation passes the `opts` to `parse_url_opts`. `get_store` calls `get_store_with_opts` with an empty iterator now.

src/registry.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    fn list_urls(&self) -> Vec<String> {
+        let stores = self.object_stores.read().unwrap();
+        stores.keys().cloned().collect()
+    }
+}
+
+/// Get the key of a url for object store registration.
+/// The credential info will be removed
+fn get_url_key(url: &Url) -> String {
```
rtyler, line 181: I recommend making this a public URL so that downstream consumers can utilize it for consistency as well

rtyler (commented):
I think I was misunderstanding how the client usage here might look, the registry contains _initialized_ and configured `ObjectStoreRef`s based on the provided `Url`.  I was crossing the streams in my brain and thinking about the Url _scheme_ too much. 

Upon further consideration this would be a bit more limiting than I initially considered. As such it would not replace the factories within delta-rs since there we are needing to register a factory/closure/whatever-you-wanna-call-it which will produce those `ObjectStoreRef` objects rather than caching initialized refs in some global state.  

I'm pondering this a bit more but the visage of an API isn't coming to me yet :thinking:

tustvold:
Yes, the default implementation relies on pre-registering stores. Originally DF also had factory abstractions but this became increasingly unwieldy, and their interactions hard to reason about. This was resolved by extracting the ObjectStoreRegistry as a trait, allowing people full control over how to convert URLs to stores, cached or otherwise.

criccomini:
SGTM! That was my initial line of thinking. I will revert the `get_store_with_opts` method :)

criccomini:
> the registry contains initialized and configured ObjectStoreRefs based on the provided Url

@kylebarron yes and no. The registry contains initialized/configured `ObjectStoreRefs`. It does not necessarily require them to be generated from the Url since you may register the ObjectStore directly using `register_store`. That method allows you to initialize the stores however you want (for example, using `parse_url_opts`). The default implementation simply uses `parse_url`.

criccomini:
@kylebarron Looking at the delta-rs code a bit more. It seems like the `ObjectStoreFactoryRegistry` stuff is perhaps a bit closer to what the `ObjectStoreRegistry` does in this PR:

https://github.com/delta-io/delta-rs/blob/3d2bbef5cb7c5f6b7e5d1b778edbfd098e993bc3/crates/core/src/logstore/factories.rs#L64-L94

In particular `store_for` looks pretty close. I wonder if something like this would work:

```rust
/// Simpler access pattern for the [ObjectStoreFactoryRegistry] to get a single store
pub fn store_for<K, V, I>(url: &Url, options: I) -> DeltaResult<ObjectStoreRef>
where
    I: IntoIterator<Item = (K, V)>,
    K: AsRef<str> + Into<String>,
    V: AsRef<str> + Into<String>,
{
    if let Some(store) = object_store_registry.get_store(&url) {
        return store;
    } else { // TODO would need to validate scheme is memory/url
        let (store, _) = object_store::parse_url_opts(url, options);
        Ok(Arc::new(store))
    } else {
        Err(DeltaTableError::InvalidTableLocation(url.clone().into()))
    }
}
```

Admittedly, I didn't dig into the storage config stuff. Just sort of spit-balling to pressure test this PR's API.

criccomini:
@tustvold One thing I'm not clear on right now: the `parse_url` method returns the store and path separately. So if I have two URLs:

- `s3://foo/bar/baz`
- `s3://foo/blah`

I think they'll both get the same object store right now. I'm not so sure that's the desired behavior. Wouldn't we want to return an ObjectStore that has a root relative to the path?

src/registry.rs
alamb: It seems like an implementation detail of the `DefaultObjectStoreRegistry` to me 🤔  so I am not sure about making it public

If it needs to be public, perhaps it could a method on `DefaultObjectStoreRegistry` like:

```rust
impl DefaultObjectStoreRegistry {
  pub fn url_key(url: &Url) -> String { ... }
}
```

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+        use crate::local::LocalFileSystem;
+        let registry = DefaultObjectStoreRegistry::new();
+        let url = Url::parse("file:///foo/bar").unwrap();
+        let store = Arc::new(LocalFileSystem::new()) as Arc<dyn ObjectStore>;
+        let old_store = registry.register_store(&url, Arc::clone(&store));
+        assert!(
+            old_store.is_none(),
+            "Should not return a previous store when registering a new one"
+        );
+        let retrieved_store = registry
```
alamb, line 195: I think it would be good to test that a sub-prefix of this worked too -- aka test that `file:///foo/bar/baz` also resolves to the store

Also it would be good to test that `file://bar` did not resolve

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+///
+/// 2. Systems relying on ad-hoc discovery, without corresponding DDL, can create [`ObjectStore`]
+///    lazily by providing a custom implementation of [`ObjectStoreRegistry`]
+pub trait ObjectStoreRegistry: Send + Sync + std::fmt::Debug + 'static {
+    /// If a store with the same key existed before, it is replaced and returned
+    fn register_store(
+        &self,
+        url: &Url,
+        store: Arc<dyn ObjectStore>,
+    ) -> Option<Arc<dyn ObjectStore>>;
```
alamb, line 65:
```suggestion
    /// Register a new store for any url's that begin with `prefix` 
    /// 
    /// If a store with the same key existed before, it is replaced and returned
    fn register_store(
        &self,
        prefix: &Url,
        store: Arc<dyn ObjectStore>,
    ) -> Option<Arc<dyn ObjectStore>>;
```

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+/// - If an AWS S3 object store can be ad-hoc discovered by the url `s3://my_bucket/lineitem/`, this
+///   object store will be registered with key `s3://my_bucket` and returned.
+///
+/// - Otherwise `None` will be returned, indicating that no suitable [`ObjectStore`] could
+///   be found
+///
+/// This allows for two different use-cases:
+///
+/// 1. Systems where object store buckets are explicitly created using DDL, can register these
+///    buckets using [`ObjectStoreRegistry::register_store`]
```
alamb, line 55:
```suggestion
/// 1. Systems where object store buckets are explicitly created using DDL, can use the provided [`DefaultObjectStoreRegistry`] to register these
///    buckets using [`ObjectStoreRegistry::register_store`]
```

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// If no [`ObjectStore`] found for the `url`, ad-hoc discovery may be executed depending on
+    /// the `url` and [`ObjectStoreRegistry`] implementation. An [`ObjectStore`] may be lazily
+    /// created and registered.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// List all registered URLs
+    fn list_urls(&self) -> Vec<String>;
+}
+
+/// The default [`ObjectStoreRegistry`]
```
alamb, line 82: Having a default registry that opens files makes sense for DataFusion but I think is less clearly the right behavior for the generic `object_store` crate.

We should either:
1.  Document the automatic creation of `LocalFileSystem`  (if that is indeed the behavior we want).
2. Remove the automatic creation of `LocalFileSystem` (or at least make it a config flag)
4. Remove `DefaultObjectStoreRegistry` from this PR

Perhaps something like

```suggestion
/// A  default  in memory [`ObjectStoreRegistry`] that maps a prefix to an ObjectStore
```

alamb (commented):
Thank you @criccomini  -- I think this PR is looking quite good - I left some comments

One question I have is how this PR supports the requested usecase:

> Since object_store does not currently expose a get_url() method, this is useful in cases where a user wants to go from an ObjectStore back to a URL.

I don't understand how you would go from `ObjectStore` to URL. It is pretty clear how to go from `url` --> `ObjectStore` however

alamb:
> @tustvold One thing I'm not clear on right now: the `parse_url` method returns the store and path separately. So if I have two URLs:
> 
> * `s3://foo/bar/baz`
> * `s3://foo/blah`
> 
> I think they'll both get the same object store right now. I'm not so sure that's the desired behavior. Wouldn't we want to return an ObjectStore that has a root relative to the path?

I think it depends on what store was registered. It seems to me like if you had registered 2 stores:
* `s3://foo/bar` --> S1
* `s3://foo/blah` --> S2

I would expect the two URLs to map to:
* `s3://foo/bar/baz` --> S1, `path = `baz`
* `s3://foo/blah` -> S2, path = ``

criccomini:
> I don't understand how you would go from ObjectStore to URL. It is pretty clear how to go from url --> ObjectStore however

I was thinking we'd:

1. `list_urls`
2. For each URL, get_store
3. if Arc::ptr_eq(retrieved_store, store_i_want_url_for) then return URL

Not ideal, but this was (very roughly) the suggestion @tustvold suggested in https://github.com/apache/arrow-rs-object-store/issues/347#issuecomment-2849055301.

tustvold:
If you want to provide an ObjectStoreRegistry that keys stores based on more than just the host and scheme, you certainly can. This implementation could even wrap the DefaultObjectStoreRegistry to ensure you don't create multiple ObjectStore for the same bucket which makes things like connection pooling less effective.

That being said I'm not sure I really understand the advantage of doing this over just consistently using the "absolute" paths and not relying on PrefixStore...

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+                stores.insert(s, Arc::clone(&store));
+                Some(store)
+            }
+            Err(_) => None,
+        }
+    }
+
+    fn list_urls(&self) -> Vec<String> {
+        let stores = self.object_stores.read().unwrap();
+        stores.keys().cloned().collect()
```
tustvold, line 150: This doesn't return URLs, but I believe host and port

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// - URL with scheme `s3://bucket/` will return the S3 store
+    /// - URL with scheme `hdfs://hostname:port/` will return the hdfs store
+    ///
+    /// If no [`ObjectStore`] found for the `url`, ad-hoc discovery may be executed depending on
+    /// the `url` and [`ObjectStoreRegistry`] implementation. An [`ObjectStore`] may be lazily
+    /// created and registered.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// List all registered URLs
+    fn list_urls(&self) -> Vec<String>;
```
tustvold, line 79: This API doesn't really make sense to me, URLs of what? Stores aren't keyed by URL...

src/registry.rs
criccomini: IIUC, the keys in `object_stores` are the returns from `get_url_key`. `get_url_key` currently returns a URL:

```rust
    format!(
        "{}://{}",
        url.scheme(),
        &url[url::Position::BeforeHost..url::Position::AfterPort],
    )
```

src/registry.rs
```diff
@@ -0,0 +1,243 @@
...
+///   be found
+///
+/// This allows for two different use-cases:
+///
+/// 1. Systems where object store buckets are explicitly created using DDL, can register these
+///    buckets using [`ObjectStoreRegistry::register_store`]
+///
+/// 2. Systems relying on ad-hoc discovery, without corresponding DDL, can create [`ObjectStore`]
+///    lazily by providing a custom implementation of [`ObjectStoreRegistry`]
+pub trait ObjectStoreRegistry: Send + Sync + std::fmt::Debug + 'static {
```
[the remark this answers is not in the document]
criccomini: Will do! I think it's fine as part of this PR.

src/registry.rs
criccomini: Oops! I think I actually did remove the local filesystem registration. I must have missed some docs. I'll take a pass and make sure the docs reflect reality. My intention was to remove the default registration for local FS

src/registry.rs
criccomini: Yea, IMO it should be private since it's just used for the default registry. Other registries might want to cache differently.

src/registry.rs
criccomini: @alamb I want to make sure I understand what you're saying. Right now, the default registry does not automatically create a local filesystem. It starts empty. It _does_, however, auto-create an ObjectStore using `parse_url` for _any_ `get_store` call that results in a cache miss. So if you say `get_store("file:///foo/bar")` and there is no store for `file:///foo`, it will create one.

Are you suggesting that we fully remove the `parse_url` behavior from the `DefaultObjectStoreRegistry` to make it as dumb as possible (essentially just a map)?

src/registry.rs
criccomini: Renamed to `get_store_prefixes` in line with `prefix` param in `register_store`.

src/registry.rs
criccomini: Added here: 005c9f2

src/registry.rs
criccomini: Thinking about this more: I think we should just make `DefaultObjectStoreRegistry` really dumb by default. Least surprises and all. So I'm inclined to remove `get_url_key` from the `DefaultObjectStoreRegistry` and just have it return `None` if nothing matches a URL get_store key.

src/registry.rs
```diff
@@ -0,0 +1,258 @@
...
+    /// - URL with scheme `s3://bucket/` will return the S3 store
+    /// - URL with scheme `hdfs://hostname:port/` will return the hdfs store
+    ///
+    /// If no [`ObjectStore`] found for the `url`, ad-hoc discovery may be executed depending on
+    /// the `url` and [`ObjectStoreRegistry`] implementation. An [`ObjectStore`] may be lazily
+    /// created and registered.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// List all registered store prefixes
+    fn get_store_prefixes(&self) -> Vec<String>;
```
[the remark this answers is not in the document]
criccomini: Switched to `Url` for everything.

src/registry.rs
criccomini: Switched to return `Vec<Url>`

criccomini:
Ok, I have simplified things quite a bit. The `DefaultObjectStoreRegistry` is now pretty dumb. It just does exact matching, so all prefixes are treated as exact URLs.

I also added a method `get_url` to the registry trait. The default imnplementation scans the map looking for a matching URL and returns the `ObjectStore` if it finds one. I played with using a bimap, but I didn't want to add a dependency and I couldn't find a nice way to make `Arc<dyn ObjectStore>` play nicely as a key. I figure a scan is nice and simple for a default impl.

src/registry.rs
```diff
@@ -0,0 +1,249 @@
...
+    /// If no [`ObjectStore`] is found for the `url`, an [`ObjectStore`] may be lazily be
+    /// created and registered. The logic for doing so is left to each [`ObjectStoreRegistry`]
+    /// implementation.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Given one of the `Arc<dyn ObjectStore>`s you registered, return its URL.
+    fn get_url(&self, store: Arc<dyn ObjectStore>) -> Option<Url>;
+
+    /// List all registered store prefixes
+    fn get_store_prefixes(&self) -> Vec<Url>;
```
alamb, line 68: one final pedantic question -- should we make it `Vec<&Url>`? Owned is probably fine though

alamb (approved):
Thank you @criccomini  I think this looks really nice and we have now bikeshed enough on the API

src/registry.rs
criccomini: Eh, I kinda prefer keeping it Url just so we don't have to deal with lifetimes in the traits. If you feel strongly, I don't mind updating it though. :)

criccomini:
I opted to add a `PrefixObjectStoreRegistry` that uses the same `get_url_key` logic as data fusion. It builds on top of the dumb `DefaultObjectStoreRegistry` and takes a pluggable parameter if users want to provide their own function. I feel like it's a reasonable compromise. It's not invasive and it does what DF does. It still doesn't address the `parse_url_opts` usecase, but IMO that's beyond the scope of this PR.

src/registry.rs
alamb: I think it is fine

criccomini:
@alamb @tustvold Ok, I think this is ready for (another) review. I've added a few implementations of `ObjectStoreRegistry` that (I think) should get most users what they want. There are now three:

- `DefaultObjectStoreRegistry`: A very dumb registry that simply treats URLs in `get_store` calls as the prefix.
- `PrefixObjectStoreRegistry`: A registry that wraps `DefaultObjectStoreRegistry` and calls a `prefix_fn` on all `get_store` URLs. The default implementation uses DataFusion's `get_url_key` implementation style.
- `ParserObjectStoreRegistry`: A registry that wraps `PrefixObjectStoreRegistry` and calls a `parser_fn` on any `get_store` calls that aren't set in the wrapped (prefix) registry. The `parser_fn` defaults to using the `parse_url` method. If the `parser_fn` returns an object store, it's cached using the prefix that was resolved in `get_store`. If a user wants to use custom `opts` for different URL prefixes, they can supply their own `parser_fn`. See the tests and docs for examples that use `parse_url_opts`. This should address @rtyler's use case; it's also something we'll need for SlateDB.

Please take a pass and let me know how things look. A few notes for reviewers:

1. I'm not 100% on my Rust foo, so pay attention to silly things I might have done with Boxes, Arcs, and so on.
2. If y'all feel like this is too much, let me know. I can remove the prefix and/or parser implementations.
3. I found that `test_url_http` wasn't working as expected when I copied its code over to test the `ParserObjectStoreRegistry` on an `http` URL. I decided to fix the test as part of this PR.

tustvold:
I'll try to find some time to review this at the weekend, however, my initial reaction is that this has gotten a lot more complicated than I expected.

For example, IMO if people want to be providing custom parser_fn they should just implement the trait... 

I'm also not sure about some of the additions to the trait.

src/registry.rs
```diff
@@ -0,0 +1,672 @@
...
+    /// Get a store for the provided URL. The definition of a "suitable store" depends on
+    /// the [`ObjectStoreRegistry`] implementation. See implementation docs for more details.
+    ///
+    /// If no [`ObjectStore`] is found for the `url`, an [`ObjectStore`] may be lazily be
+    /// created and registered. The logic for doing so is left to each [`ObjectStoreRegistry`]
+    /// implementation.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Given one of the `Arc<dyn ObjectStore>`s you registered, return its URL.
+    fn get_prefix(&self, store: Arc<dyn ObjectStore>) -> Option<Url>;
```
tustvold, line 49: I'm not a massive fan of this interface, it is hard for stores to implement it efficiently, and is just a bit peculiar. 

The pattern used in DF instead is to pass around the URLs and only "materialize" the store at the point of usage

src/registry.rs
```diff
@@ -0,0 +1,672 @@
...
+
+    /// List all registered store prefixes
+    fn get_store_prefixes(&self) -> Vec<Url>;
```
tustvold, line 52: Similar to the above, I'm not sure about this method. It isn't well defined what this means, as it is up to the implementation how it wishes to key stores, which makes this abstraction leaky

tustvold (commented):
Did a quick pass, I'd like to suggest removing the new trait methods as they have quite confusing semantics and result in a very leaky abstraction as this is a lossy translation. 

As stated above, implementations should only "materialize" the URL to the store at the point of usage, with plans etc... in terms of URLs not `Arc<dyn ObjextStore>` avoiding the need for this functionality.

I think this will also remove the need for the additional impls, as the trait is now trivial for people to just implement as required for their use-case

src/registry.rs
```diff
@@ -0,0 +1,672 @@
...
+use crate::{parse_url, ObjectStore};
+use std::collections::HashMap;
+use std::sync::{Arc, RwLock};
+use url::Url;
+
+/// [`ObjectStoreRegistry`] maps a URL prefix to an [`ObjectStore`] instance. The definition of
+/// a URL prefix depends on the [`ObjectStoreRegistry`] implementation. See implementation docs
+/// for more details.
+pub trait ObjectStoreRegistry: Send + Sync + std::fmt::Debug + 'static {
+    /// Register a new store for any URL that begins with `prefix`
```
tustvold, line 31: This has subtly changed the behaviour from DF, which delegates it to the store to determine the keying portion of the URL. I think this needs to be changed back, it was that way for a reason.

tustvold (changes requested):
I think this needs changes prior to merge, in particular it makes changes to the way that URLs behave that would make this abstraction not usable by DF. The goal of ObjectStoreRegistry is effectively to act as cache for [parse_url](https://docs.rs/object_store/latest/object_store/fn.parse_url.html), it is not intended for implementations to be keyed on the full path to an object - rather some implementation-defined subset of that URL.

src/registry.rs
```diff
@@ -0,0 +1,672 @@
...
+        store: Arc<dyn ObjectStore>,
+    ) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Get a store for the provided URL. The definition of a "suitable store" depends on
+    /// the [`ObjectStoreRegistry`] implementation. See implementation docs for more details.
+    ///
+    /// If no [`ObjectStore`] is found for the `url`, an [`ObjectStore`] may be lazily be
+    /// created and registered. The logic for doing so is left to each [`ObjectStoreRegistry`]
+    /// implementation.
+    fn get_store(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
```
tustvold, line 46:
```suggestion
    fn get_store(&self, url: &Url) -> Result<(Arc<dyn ObjectStore>, Path)>;
```

Perhaps, to mirror parse_url and allow for more complex relationships between URLs and paths within stores.

ion-elgreco:
> I think I was misunderstanding how the client usage here might look, the registry contains _initialized_ and configured `ObjectStoreRef`s based on the provided `Url`. I was crossing the streams in my brain and thinking about the Url _scheme_ too much.
> 
> Upon further consideration this would be a bit more limiting than I initially considered. As such it would not replace the factories within delta-rs since there we are needing to register a factory/closure/whatever-you-wanna-call-it which will produce those `ObjectStoreRef` objects rather than caching initialized refs in some global state.
> 
> I'm pondering this a bit more but the visage of an API isn't coming to me yet 🤔

We do have this actually: https://github.com/delta-io/delta-rs/blob/3d2bbef5cb7c5f6b7e5d1b778edbfd098e993bc3/crates/core/src/logstore/storage/mod.rs#L23 (I grabbed that from datafusion codebase some time ago for the lakefs integration), so good to see it will be part of object-store :)

criccomini:
🔥 I will update the PR to simplify things. I suspected this might be the outcome, but I wanted to experiment/pressure test the APIs anyway. :)

criccomini:
@tustvold Ok, I think I've incorporated all of your feedback.

- Removed `get_store_url`
- Removed `PrefixObjectStoreRegistry`/`ParserObjectStoreRegistry`
- Updated `DefaultObjectStoreRegistry` to use `parse_opts` and added a `map_url_to_key` method.

Two other things:

1. The `map_url_to_key` adheres to similar rules to `ObjectStoreScheme::parse` to make it work with memory/file schemes. This seemed right to me for the time being. Once #356 is done, my code should probably use `ObjectStoreScheme::parse` directly.
2. I did not incorporate any of the environment parsing stuff you mentioned in https://github.com/apache/arrow-rs-object-store/pull/348#pullrequestreview-2835811212. TBH, I feel that `parse_url_opts` is a better place for that (rather than using `std::iter::empty::<(&str, &str)>()` when it invokes `parse_url_opts`).

tustvold:
I have an alternate proposal here, PTAL - https://github.com/apache/arrow-rs-object-store/pull/375

criccomini:
Closing in favor of #375

Issue #287, opened 2025-02-06
Release object_store `0.12.1` (non breaking) Around April 30 2025


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Previous release:
- https://github.com/apache/arrow-rs-object-store/issues/28

**Describe the solution you'd like**
Let's plan for a minor release

- [x] Update version + make release notes: https://github.com/apache/arrow-rs-object-store/pull/349
- [x] Prepare release candidate + begin vote thread: https://lists.apache.org/thread/w4v08jfjogcqxz1kcnrg44km65jkf9sh
- [x] Release to crates.io: https://crates.io/crates/object_store/0.12.1
- [x] Create ticket for next release: https://github.com/apache/arrow-rs-object-store/issues/307

**Describe alternatives you've considered**

**Additional context**
See discussion on https://github.com/apache/arrow-rs-object-store/issues/306

ion-elgreco:
@alamb do you reckon we could land this https://github.com/apache/arrow-rs-object-store/pull/332 on 0.12.x? Then it would be compatible with DF47

alamb:
> [@alamb](https://github.com/alamb) do you reckon we could land this [#332](https://github.com/apache/arrow-rs-object-store/pull/332) on 0.12.x? Then it would be compatible with DF47

Yes for sure -- I think that is very reasonable.

jkosh44:
If not already planned, it would be great to see https://github.com/apache/arrow-rs-object-store/pull/257 included in this release!

alamb:
> If not already planned, it would be great to see [#257](https://github.com/apache/arrow-rs-object-store/pull/257) included in this release!

that will be included for sure -- I don't think we need to do anything else additional here.

alamb:
I have made a PR with a version update and changelog: 
- https://github.com/apache/arrow-rs-object-store/pull/349

alamb:
I have made a RC and begun voting: https://lists.apache.org/thread/w4v08jfjogcqxz1kcnrg44km65jkf9sh

alamb:
I found a small minor issue during publishing however the release has now been completed
- https://github.com/apache/arrow-rs-object-store/issues/357

Release is available on https://crates.io/crates/object_store/0.12.1

ion-elgreco:
Woohoo! 🥳

PR #349, opened 2025-05-08
Prepare for 0.12.1 release

alamb:

- Part of https://github.com/apache/arrow-rs-object-store/issues/287

# Rationale for this change
 
Time to release

Note this will be the first release after the object store code was moved into its own repo
- https://github.com/apache/arrow-rs/issues/6183

1. Update Changelog. See rendered version here: [prepare_0.12.1/CHANGELOG.md](https://github.com/alamb/arrow-rs-object-store/blob/alamb/prepare_0.12.1/CHANGELOG.md)
2. Update version
3. Update scripts to work (hopefully) in new repo

### DIFF
--- a/dev/release/create-tarball.sh
+++ b/dev/release/create-tarball.sh
@@ -53,8 +53,7 @@ fi
 
 object_store_version=$1
 rc=$2
-
-tag=object_store_${object_store_version}
+tag=v${object_store_version}
 
 release=apache-arrow-object-store-rs-${object_store_version}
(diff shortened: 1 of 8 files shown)


dev/release/create-tarball.sh
```diff
@@ -53,8 +53,7 @@ fi
 
 object_store_version=$1
 rc=$2
-
-tag=object_store_${object_store_version}
+tag=v${object_store_version}
```
alamb, line 56: I haven't tested this -- I expect I will do so as part of actually creating the RC

PR #351, opened 2025-05-09
fix: treat TCP reset as a retryable error

### LINKED ISSUE #350: No retries when connection closes abruptly (i.e TCP-RST)
**Describe the bug**

When the underlying connection is abruptly closed to due to TCP reset, object store does not attempt to retry the request.
Here is the error I'm getting:
```
Generic {
    store: "S3",
    source: RetryError {
        method: GET,
        uri: Some(
            https://s3.us-east-1.amazonaws.com/<redacted>,
        ),
        retries: 0,
        max_retries: 1000,
        elapsed: 25.769196ms,
        retry_timeout: 180s,
        inner: Http(
            HttpError {
                kind: Unknown,
                source: reqwest::Error {
                    kind: Request,
                    source: hyper_util::client::legacy::Error(
                        SendRequest,
                        hyper::Error(
                            Io,
                            Os {
                                code: 104,
                                kind: ConnectionReset,
                                message: "Connection reset by peer",
                            },
                        ),
                    ),
                },
            },
        ),
    },
}
```
IIRC, connection reset by peer is basically when the other side closes the connection using RST.
I'm running on Linux, mind you.

**To Reproduce**

I'm not sure how to reproduce this, would probably require to write a small HTTP server which drops the connection while serving (?).
Either that or write a unit-test for the relevant function which converts the reqwest error to object store error. But I'm unsure whether the relevant types can be constructed publicly like this.

**Expected behavior**

I expect object_store to retry in this situation.

**Additional context**

This happens because [here](https://github.com/apache/arrow-rs-object-store/blob/36752c975d4f29e20b57c91f81a10872dcd48ae7/src/client/http/connection.rs#L104) we try to downcast the reqwest error to classify it better so we can act accordingly. However, my guess is that since the transition to hyper 1.0, the reqwest crate started relying on `hyper_util::client::legacy::Client` and thus the error type is `hyper_util::client::legacy::Error` rather than `hyper::Error`.

I'm willing to contribute a PR to fix this

OmriSteiner:

Closes #350.

# Rationale for this change

We want object_store to retry the request in case the connection closed abruptly.

Map `std::io::ErrorKind::ConnectionReset` -> `HttpErrorKind::Interrupted`

No.

### DIFF
--- a/src/client/http/connection.rs
+++ b/src/client/http/connection.rs
@@ -114,6 +114,7 @@ impl HttpError {
                 match e.kind() {
                     std::io::ErrorKind::TimedOut => kind = HttpErrorKind::Timeout,
                     std::io::ErrorKind::ConnectionAborted
+                    | std::io::ErrorKind::ConnectionReset
                     | std::io::ErrorKind::BrokenPipe
                     | std::io::ErrorKind::UnexpectedEof => kind = HttpErrorKind::Interrupted,
                     _ => {}


tustvold (approved):

berkaysynnada:
Failure on the main is related with this?

OmriSteiner:
> Failure on the main is related with this?

That same test passed on this branch.

The only reason I can think of is that statistically the delete request in the test fails - and this change causes it to both (!):
* Not retry
* Not fail the request

I can get behind not retry - maybe there was an error which was previously mapped to `Timeout` / `Connect` / `Decode` and now it is mapped to `Interrupted` - which is retried in case the request is not idempotent (delete isn't).

But I would expect the test to panic at `src/integration.rs:955` because the request fails.


-----------------------

To be fair, it's possible that TCP-RST on connect actually caused this to convert `Connect` errors to `Interrupted` errors.
WDYT about only doing the error refinement in case `kind == Unknown`?

tustvold:
I think it is just a flaky test unrelated to this PR

Issue #18, opened 2025-02-19
Remove file-handle from object store GET operations


crepererum:
# What
Remove [`GetResultPayload::File`](https://docs.rs/object_store/latest/object_store/enum.GetResultPayload.html#variant.File) (and the entire enum).

# Why
The file handle has several drawbacks as it does NOT specify the following properties:

- **mode:** Is the file opened read-only or does it also have write access (which would be bad)?
- **seek position:** At which position is the current file "read head"? Are users expected to seek to the file start if they want to read the entire file?
- **direct IO:** Is the file opened using direct IO? (in which case the user would have to read in aligned blocks)
- **exclusive access:** Does the API user have exclusive access to the said file handle (and hence can use seek-based operations) or not (in which case only non-seeking operations like [`read_at`](https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#tymethod.read_at) would be required)

At the same time I suspect that most users copy the file data into a buffer anyways and don't use any kernel-based IO operations like [`sendfile`](https://man.archlinux.org/man/sendfile.2).

All object store users and integrating libraries potentially have to deal with streams AND file handles.

# How
Remove the entire [`GetResultPayload`](https://docs.rs/object_store/latest/object_store/enum.GetResultPayload.html) enum and just use streams.

AdamGS:
I would also love this, I think a file-like API makes a lot of sense for some perf-sensetive stuff locally, and works perfectly well remotely.

PR #353, opened 2025-05-09
Move payload methods from `GetResult` onto `GetResultPayload`

### LINKED ISSUE #352: Move payload helpers from `GetResult` to `GetResultPayload`

Calling `GetResult::bytes` or `GetResult::into_stream` consumes all of `GetResult` even though the only part that really needs to be consumable is the `GetResultPayload`. Instead, having the helpers live on `GetResultPayload` means that downstream users can store their own `GetResult` and have the `GetResult::meta` and `GetResult::attributes` be accessible even after consuming the payload.

Move payload helpers from `GetResult` to `GetResultPayload`

Alternatively, we could just clone the `range`, `meta`, and `attributes` fields on `GetResult`. It's probably not that much overhead.

**Additional context**

kylebarron:

Closes #352.

# Rationale for this change
 

Move payload helpers from `GetResult` to `GetResultPayload`.

This is **not** a breaking change because the helpers still exist on `GetResult` too.

Only added methods.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1102,6 +1102,10 @@ impl GetResultPayload {
     ///
     /// If not called from a tokio context, this will perform IO on the current thread with
     /// no additional complexity or overheads
+    #[cfg_attr(
+        not(all(feature = "fs", not(target_arch = "wasm32"))),
+        allow(unused_variables)
+    )]
     pub fn into_stream(self, range: Range<u64>) -> BoxStream<'static, Result<Bytes>> {
         match self {
             #[cfg(all(feature = "fs", not(target_arch = "wasm32")))]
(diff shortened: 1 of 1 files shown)


tustvold (commented):
I think we should probably better document how and when the range is used, i.e. that if the payload is a stream it is simply ignored

kylebarron:
I added a note that the `range` parameter is only used when the payload is a file.

tustvold:
I wonder if this potentially conflicts with https://github.com/apache/arrow-rs-object-store/issues/18 ?

kylebarron:
> I wonder if this potentially conflicts with #18 ?

Probably, yes.

crepererum:
Since I'm gonna implement #18 for version 0.13 anyways, I think this breaking change here is kinda redundant.

Issue #357, opened 2025-05-13
Error running `cargo publish`:  wildcard (`*`) dependency constraints are not allowed on crates.io.


alamb:
**Describe the bug**
I tried to release 0.12.1 on crates io and go tan error
- https://github.com/apache/arrow-rs-object-store/issues/287

**To Reproduce**
```shell
cargo publish

error: failed to publish to registry at https://crates.io

Caused by:
  the remote server responded with an error (status 400 Bad Request): wildcard (`*`) dependency constraints are not allowed on crates.io. Crate with this problem: `wasm-bindgen-test` See https://doc.rust-lang.org/cargo/faq.html#can-libraries-use--as-a-version-for-their-dependencies for more information
```

**Expected behavior**
Should publish successfully

**Additional context**
I fixed it locally like this:

```diff
--- Cargo.toml.orig	2025-05-13 06:35:15
+++ Cargo.toml	2025-05-13 06:35:50
@@ -89,7 +89,7 @@
 reqwest = { version = "0.12", features = ["gzip"] }

 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
-wasm-bindgen-test = "*"
+wasm-bindgen-test = "0.3.50"

 [dev-dependencies.getrandom_v03]
 package = "getrandom"
```

ion-elgreco:
Docs build failed as well, probably due to that failure

alamb:
> Docs build failed as well, probably due to that failure

- Filed https://github.com/apache/arrow-rs-object-store/issues/360 to track

PR #358, opened 2025-05-13
Fix `cargo publish` by specifying version for wasm-bindgen-test

alamb:

- Closes https://github.com/apache/arrow-rs-object-store/issues/357

# Rationale for this change
 
I can't run `cargo publish` without doing this first

See subject

No, this affects testing and release, not users

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,7 +89,7 @@ regex = "1.11.1"
 reqwest = { version = "0.12", features = ["gzip"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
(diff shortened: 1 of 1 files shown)


tustvold (approved):
I thought we did `cargo publish --dry-run` when testing a tarball, guessing that doesn't catch this?

alamb:
> I thought we did `cargo publish --dry-run` when testing a tarball, guessing that doesn't catch this?

🤔  it seems that we don't (or no longer do):

https://github.com/apache/arrow-rs-object-store/blob/main/dev/release/create-tarball.sh

tustvold:
It looks like we do it as part of verify-release.sh, and it appears it doesn't catch this sadly

Issue #343, opened 2025-04-28
Error in rust 1.87: integer out of range for `u16` in format string


alamb:
**Describe the bug**
- Reported by @cuviper on  https://github.com/rust-lang/rust/pull/136932

In Rust 1.87 some of the object store tests will fail to compile, due to a change in formatting

```
out] error: invalid format string: integer `5300000` does not fit into the type `u16` whose range is `0..=65535`
[INFO] [stdout]     --> src/integration.rs:1123:45
[INFO] [stdout]      |
[INFO] [stdout] 1123 |         .put_part(Bytes::from(format!("1:{:05300000},", 0)).into())
[INFO] [stdout]      |                                             ^^^^^^^ integer out of range for `u16` in format string
[INFO] [stdout] 
```
**To Reproduce**

**Expected behavior**
We should rewrite this test not to use a giant format string

**Additional context**

Issue #307, opened 2025-03-24
[Discuss] Release object store 0.13.0 (breaking) or 0.12.2 (non breaking API) June, 2025


alamb:
**Which part is this question about**

Last breaking release was in March 2025
- https://github.com/apache/arrow-rs-object-store/issues/28

Next scheduled release is minor at end of April 2025
- https://github.com/apache/arrow-rs-object-store/issues/287

Per [release schedule](https://github.com/apache/arrow-rs-object-store?tab=readme-ov-file#release-schedule) we aim for2 months release: 

**Describe your question**
What type of release should we plan for in June 2025?
1. Another incremental release
2. A major release including breaking API changes (e.g. to include https://github.com/apache/arrow-rs-object-store/pull/303)

**Additional context**
- See prior discussion https://github.com/apache/arrow-rs-object-store/issues/306

alamb:
I don't know of any major API changes waiting, so I propose making this a 0.12.2 release

alamb:
@crepererum suggests a `0.13.0` release here:
- https://github.com/apache/arrow-rs-object-store/issues/367

alamb:
In order to get a fix for the documentation out, we are going to accelerate this release. Please see discussion here
- https://github.com/apache/arrow-rs-object-store/issues/360#issuecomment-2892160772

I'll try and make one near the end of this week

alamb:
I plan to make the release candidate later today

alamb:
PR to prepare the release
- https://github.com/apache/arrow-rs-object-store/pull/391

alamb:
I have begun voting for the release candidate: https://lists.apache.org/thread/0kcpv3kxow50np845j3bm4owmfx6z653

alamb:
The release is approved and available

The release is available here:
  https://dist.apache.org/repos/dist/release/arrow/arrow-object-store-rs-0.12.2

I have also released it to crates.io:
  https://crates.io/crates/object_store/0.12.2

PR #359, opened 2025-05-13
Update integration test to avoid long format strings

alamb:

- Closes https://github.com/apache/arrow-rs-object-store/issues/343

# Rationale for this change
 
The new rust release will limit the length of formatting identifiers, so we need to update the tests to create test data using a different mechanism

Rewrite test to avoid soon to be unsupported `format!` strings

### DIFF
--- a/src/integration.rs
+++ b/src/integration.rs
@@ -24,8 +24,6 @@
 //!
 //! They are intended solely for testing purposes.
 
-use core::str;
-
 use crate::multipart::MultipartStore;
 use crate::path::Path;
 use crate::{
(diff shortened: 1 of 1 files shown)


src/integration.rs
```diff
@@ -1119,48 +1117,64 @@ pub async fn multipart_race_condition(storage: &dyn ObjectStore, last_writer_win
     let mut multipart_upload_1 = storage.put_multipart(&path).await.unwrap();
     let mut multipart_upload_2 = storage.put_multipart(&path).await.unwrap();
 
+    /// Create a string like `"1:"`  followed by `part` padded to 5,300,000 places
+    ///
+    /// equivalent of format!("{prefix}:{part:05300000}"), which is no longer suppoted
```
alamb, line 1122:
```suggestion
    /// equivalent of format!("{prefix}:{part:05300000}"), which is no longer supported
```

tustvold (approved):

zensh:
I am facing the same issue. May I ask when a new version will be released to fix this problem?

https://github.com/ldclabs/ic-oss/actions/runs/15457272105/job/43511721430

alamb:
> I am facing the same issue. May I ask when a new version will be released to fix this problem?
> 
> https://github.com/ldclabs/ic-oss/actions/runs/15457272105/job/43511721430

I am hoping shortly but there is some sort of integration test that needs to be debugged. See the status of the release here
- https://github.com/apache/arrow-rs-object-store/issues/307

If you can spare some help debugging whatever is wrong that would be super helpful
- https://github.com/apache/arrow-rs-object-store/issues/395

PR #362, opened 2025-05-14
Add ObjectStoreUrl

### LINKED ISSUE #356: Add ObjectStoreUrl

Some object store URL conventions encode the bucket name in the path of the URL, e.g. `https://ACCOUNT_ID.r2.cloudflarestorage.com/<bucket>/<path>`, instead of the host e.g. `s3://<bucket>/<path>`

`ObjectStoreScheme::parse` returns both the `ObjectStoreScheme` and the `Path` within that store to handle this, however, it is insufficient when implementing an ObjectStoreRegistry (#348) where we need something to key the store based on.

I'd like a mechanism given a URL to determine the store and path components.

```
/// Path to an object in an ObjectStore
pub struct ObjectStoreUrl {
    scheme: ObjectStoreScheme,
    store: Url,
    path: Path,
}

impl ObjectStoreUrl {
    pub fn parse(url: &Url) -> Result<Self> {...}

    /// Returns the scheme of this URL
    pub fn scheme(&self) -> ObjectStoreScheme { ... }

    /// Returns the URL of the store root
    pub fn store(&self) -> &Url { ... }
    
    /// Returns the path to the object within the store
    pub fn path(&self) -> &Path { ... }
}
```

**Additional context**

### LINKED ISSUE #361: Why doesn't list_with_delimiter yield a BoxStream?

I was curious why the list and list_with_delimiter functions on the `ObjectStore` don't have the same return type 🤔 

I have some ... storage with literally billions of objects  The response of a ListResult from list_with_delimiter I don't understand while `list` provides a BoxStream, which I presume to be yielding results from paginated calls to the underlying storage provider.

criccomini:
I took a crack at adding an `ObjectStoreUrl`, which makes it easier to track store URLs in addition to paths when calling parse_url and when using `ObjectStoreRegistry` in #348. I updated `ObjectStoreScheme::parse` to use `ObjectStore::url` under the hood, and deprecated the old parse method.

Note: I opted to leave usernames, but strip passwords for URLs in the `ObjectStoreUrl::parse` method.

Fixes #356

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -82,8 +82,13 @@ impl ObjectStoreUrl {
         };
 
         let (scheme, store_url, path) = match (url.scheme(), url.host_str()) {
-            ("file", _) | ("memory", _) => (
+            ("file", _) => (
                 ObjectStoreScheme::Local,
+                url[url::Position::BeforeScheme..url::Position::AfterHost].to_string(),
+                &url[url::Position::BeforeHost..url::Position::AfterPath],
+            ),
+            ("memory", _) => (
+                ObjectStoreScheme::Memory,
                 url[url::Position::BeforeScheme..url::Position::AfterHost].to_string(),
                 &url[url::Position::BeforeHost..url::Position::AfterPath],
             ),
@@ -166,7 +171,7 @@ impl ObjectStoreUrl {
         Ok(Self {
             scheme,
             store: store_url,
-            path: path.into(),
+            path: Path::from_url_path(path)?,
         })
     }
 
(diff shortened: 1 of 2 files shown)


src/parse.rs
```diff
@@ -43,6 +43,162 @@ impl From<Error> for super::Error {
...
+                ObjectStoreScheme::AmazonS3,
+                url[url::Position::BeforeScheme..url::Position::AfterHost].to_string(),
+                url.path(),
+            ),
+            ("gs", Some(_)) => (
+                ObjectStoreScheme::GoogleCloudStorage,
+                url[url::Position::BeforeScheme..url::Position::AfterHost].to_string(),
+                url.path(),
+            ),
+            ("az" | "adl" | "azure" | "abfs" | "abfss", Some(_)) => (
```
tustvold, line 109: IIRC there is a pre-existing bug where the bucket is part of the path for some of these, check AzureBuilder

src/parse.rs
```diff
@@ -43,6 +43,162 @@ impl From<Error> for super::Error {
...
+                &url[url::Position::BeforeScheme..url::Position::AfterPort],
+                bucket,
+            ))
+            .map_err(|_| Error::Unrecognised { url: url.clone() })
+        };
+
+        let (scheme, store_url, path) = match (url.scheme(), url.host_str()) {
+            ("file", None) => (
+                ObjectStoreScheme::Local,
+                url[url::Position::BeforeScheme..url::Position::AfterHost].to_string(),
```
tustvold, line 91: Is there a reason we don't just use BeforePath? I worry we're discarding part of the URL that could matter

tustvold (commented):
Did a quick scan

criccomini:
> I need to think a bit more about how this should interplay with parse_url and ObjectStoreRegistry. I plan to take some time at the weekend

Figured. I'll leave this up anyway in case it's useful. :) Feel free to close it, though.

tustvold:
As I somewhat suspected might happen, in the process of working out how ObjectStoreRegistry should work I devised a formulation that doesn't require this - see #375

PR #364, opened 2025-05-15
fix: mask the aws credential info

yanghua:

Closes #.

# Rationale for this change
 


src/aws/credential.rs
```diff
@@ -77,6 +77,16 @@ pub struct AwsCredential {
     pub token: Option<String>,
 }
 
+impl std::fmt::Debug for AwsCredential {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("AwsCredential")
+            .field("key_id", &"******")
```
tustvold, line 83: This isn't secret and is safe to print

yanghua: @tustvold reasonable, let's only mask `secret_key ` and `token `.

rtyler (approved):

PR #366, opened 2025-05-15
Return Non-Generic Errors from HttpStore

### LINKED ISSUE #365: Error handling of HTTP storage backend not utilizing retry::RetryError::error when possible
**Describe the bug**
When using the HTTP/WebDAV storage backend many HTTP error statuses result in `crate::Error:Generic` making it harder to handle errors, especially compared to Azure/AWS/GCP backends which properly utilize `retry::RetryError::error`

**To Reproduce**
Use the HTTP storage backend, run a query that would result in a 304/401/403 response.

**Expected behavior**
Requests should return the relevant crate::Error based on the RetryError to easily handle the separate cases.

**Additional context**

PR available: #366

Rynoxx:

Closes #365.

# Rationale for this change

Allows for better error handling of HTTP statuses, and makes it consistent with how the S3/Azure/GCP clients handles RetryError.

Adjust implementation of `impl From<http::client::Error> for crate::Error` to use RetryError::error when possible.

Most RetryError cases just returned `crate::Error::GenericError` before, now they'll return the crate::Error that matches the RetryError

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -292,17 +292,7 @@ impl Client {
             .delete(url)
             .send_retry(&self.retry_config)
             .await
-            .map_err(|source| match source.status() {
-                Some(StatusCode::NOT_FOUND) => crate::Error::NotFound {
-                    source: Box::new(source),
-                    path: path.to_string(),
-                },
-                _ => Error::Request {
-                    source,
-                    path: path.to_string(),
-                }
-                .into(),
-            })?;
+            .map_err(|source| source.error(STORE, path.to_string()))?;
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


src/http/client.rs
```diff
@@ -275,7 +297,11 @@ impl Client {
                     source: Box::new(source),
                     path: path.to_string(),
                 },
-                _ => Error::Request { source }.into(),
+                _ => Error::Request {
```
tustvold, line 300: I think we can now remove this match block entirely

Rynoxx: Good point, cleaned that up

tustvold (approved):

PR #369, opened 2025-05-16
Fix clippy for Rust 1.87

alamb:

Closes #.

# Rationale for this change
 
Rust 1.87 was released with newly stringent clippy

Make clippy happy

No

### DIFF
--- a/src/gcp/credential.rs
+++ b/src/gcp/credential.rs
@@ -123,6 +123,7 @@ pub struct ServiceAccountKey(RsaKeyPair);
 
 impl ServiceAccountKey {
     /// Parses a pem-encoded RSA key
+    #[expect(clippy::result_large_err)]
     pub fn from_pem(encoded: &[u8]) -> Result<Self> {
         use rustls_pemfile::Item;
         use std::io::Cursor;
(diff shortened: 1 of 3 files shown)


tustvold:
Lol snap - https://github.com/apache/arrow-rs-object-store/pull/370

I think it would be better to just do what it is saying to do

src/gcp/credential.rs
```diff
@@ -123,6 +123,7 @@ pub struct ServiceAccountKey(RsaKeyPair);
 
 impl ServiceAccountKey {
     /// Parses a pem-encoded RSA key
+    #[expect(clippy::result_large_err)]
```
alamb, line 126: I think the root cause of this is that the `TokenRequest` error is large according to Clippy:

```
error: the `Err`-variant returned from this function is very large
   --> src/gcp/credential.rs:126:40
    |
86  | /     TokenRequest {
87  | |         source: crate::client::retry::RetryError,
88  | |     },
    | |_____- the largest variant contains at least 192 bytes
...
126 |       pub fn from_pem(encoded: &[u8]) -> Result<Self> {
    |                                          ^^^^^^^^^^^^

```

We could probably box the `Error::TokenRequest` variant and make clippy happier, but that would be a breaking API change. For now, let's leave the code as is an tell clippy it is ok

alamb:
> Lol snap - #370
> 
> I think it would be better to just do what it is saying to do

LOL -- but it seems like #370 is a breaking API change, perhaps?

alamb:
Superceded by https://github.com/apache/arrow-rs-object-store/pull/370

PR #370, opened 2025-05-16
Fix 1.87 Clippy Lints

tustvold:


src/client/retry.rs
```diff
@@ -33,7 +33,11 @@ use web_time::{Duration, Instant};
 
 /// Retry request error
 #[derive(Debug, thiserror::Error)]
-pub struct RetryError {
+pub struct RetryError(Box<RetryErrorImpl>);
```
alamb, line 36: Isn't this a breaking public API change?

src/client/retry.rs
```diff
@@ -33,7 +33,11 @@ use web_time::{Duration, Instant};
...
+
+/// Box error to avoid large error variant
+#[derive(Debug)]
+struct RetryErrorImpl {
```
alamb, line 40: Doesn't this mean the fields are now not public on the error 🤔

src/client/retry.rs
tustvold: I'm not sure RetryError is actually public, but regardless all the fields are crate private?

alamb: I double checked and indeed I agree this error is not public
```shell
cargo doc --all-features --open
```

Screenshot 2025-05-16 at 2 16 11 PM

alamb (approved):

PR #371, opened 2025-05-17
Add PaginatedListStore

### LINKED ISSUE #291: Expose list_paginated in object_store

### LINKED ISSUE #327: Some Inconsistencies in the Path and List

tustvold:

Closes #291
Closes #327.
Relates to #361
Relates to #295 

# Rationale for this change
 

There is a desire to expose listing APIs even closer to the object stores themselves, either to support custom pagination or non-path-segment based listing.

Adds a PaginatedListStore in a similar vein to MultipartStore

### DIFF
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -1018,7 +1018,7 @@ impl ListClient for Arc<AzureClient> {
         let mut response: ListResultInternal = quick_xml::de::from_reader(response.reader())
             .map_err(|source| Error::InvalidListResponse { source })?;
 
-        let token = response.next_marker.take();
+        let token = response.next_marker.take().filter(|x| !x.is_empty());
 
         Ok(PaginatedListResult {
             result: to_list_result(response, prefix)?,
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -24,8 +24,8 @@ use crate::client::header::{get_put_result, HeaderConfig};
 use crate::client::list::ListClient;
 use crate::client::retry::RetryExt;
 use crate::client::{GetOptionsExt, HttpClient, HttpError, HttpRequest, HttpResponse};
+use crate::list::{PaginatedListOptions, PaginatedListResult};
 use crate::multipart::PartId;
-use crate::path::DELIMITER;
 use crate::util::{deserialize_rfc1123, GetRange};
 use crate::{
     Attribute, Attributes, ClientOptions, GetOptions, ListResult, ObjectMeta, Path, PutMode,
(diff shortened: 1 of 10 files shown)


src/list.rs
```diff
@@ -0,0 +1,85 @@
...
+use crate::ListResult;
+use async_trait::async_trait;
+use std::borrow::Cow;
+
+/// Options for a paginated list request
+#[derive(Debug, Default, Clone)]
+pub struct PaginatedListOptions {
+    /// Path to start listing from
+    ///
+    /// Note: Not all stores support this
```
tustvold, line 30: The combination of this, and the fact not all stores return objects in lexicographical order means that implementing pagination on top of the current APIs is not possible

src/list.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// Perform a paginated list request
+    ///
+    /// Note: the order of returned objects is not guaranteed and
+    /// unlike [`ObjectStore::list`] a trailing delimiter is not
+    /// automatically added to `prefix`
+    ///
+    /// [`ObjectStore::list`]: crate::ObjectStore::list
+    async fn list_paginated(
+        &self,
+        prefix: Option<&str>,
```
tustvold, line 82: Note that this takes `&str` instead of `&Path` to allow choosing whether to include the trailing `/` or not

tustvold:
Will look into azurite failure later this afternoon, looks like a quick fix

rtyler (approved):

src/azure/client.rs
```diff
@@ -1018,7 +1018,7 @@ impl ListClient for Arc<AzureClient> {
         let mut response: ListResultInternal = quick_xml::de::from_reader(response.reader())
             .map_err(|source| Error::InvalidListResponse { source })?;
 
-        let token = response.next_marker.take();
+        let token = response.next_marker.take().filter(|x| !x.is_empty());
```
tustvold, line 1021: I think this might have been a bug, which would potentially lead us to make a redundant extra list request, but this could also be an azurite specific quirk

PR #375, opened 2025-05-17
Add ObjectStoreRegistry (#347)

### LINKED ISSUE #347: Is there a way to go from `ObjectStore` to `(URL, opts)`?

### LINKED ISSUE #356: Add ObjectStoreUrl

tustvold:

Closes #347
Closes #356 

# Rationale for this change
 

See tickets, but the TLDR is we want to provide a flexible way to allow people to register and manage ObjectStore based on URL.

Adds an ObjectStoreRegistry and a default implementation


src/parse.rs
```diff
@@ -286,6 +286,15 @@ mod tests {
...
             ),
+            (
+                "s3://bucket/foo bar",
+                (ObjectStoreScheme::AmazonS3, "foo bar"),
+            ),
+            ("s3://bucket/😀", (ObjectStoreScheme::AmazonS3, "😀")),
+            (
+                "s3://bucket/%F0%9F%98%80",
+                (ObjectStoreScheme::AmazonS3, "😀"),
+            ),
```
tustvold, line 297: This is an unrelated test I wrote whilst playing around

src/registry.rs
```diff
@@ -0,0 +1,238 @@
...
+    ///
+    /// If a store with the same URL existed before, it is replaced and returned
+    fn register(&self, url: Url, store: Arc<dyn ObjectStore>) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Unregister a store for the provided store URL
+    fn unregister(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>;
+
+    /// Resolve an object URL
+    ///
+    /// If [`ObjectStoreRegistry::register`] has been called with a URL with the same
```
tustvold, line 61: The original formulation in DF simply uses scheme and host. This works well for bucket specific URLs like `s3://bucket/path` but it falls apart for URLs like `https://s3.region.amazonaws.com/bucket`.

The challenge is to come up with a mechanism to support keying on more than scheme and host, without requiring the registry to actually understand the components of that URL. This is the formulation I came up with

src/registry.rs
```diff
@@ -0,0 +1,238 @@
...
+    fn unregister(&self, store: &Url) -> Option<Arc<dyn ObjectStore>> {
+        Some(self.map.write().remove(store.as_str())?.store)
+    }
+
+    fn resolve(&self, url: &Url) -> crate::Result<(Arc<dyn ObjectStore>, Path)> {
+        {
+            let map = self.map.read();
+            let start = Bound::Included(&url[..url::Position::BeforePath]);
+
+            let longest = map
```
tustvold, line 179: This is the downside of this formulation, we have to use a BTreeMap which is slower than a HashMap, and if there are multiple (scheme, host) matches we have to scan through them

src/registry.rs
```diff
@@ -0,0 +1,255 @@
...
+    fn resolve(&self, to_resolve: &Url) -> crate::Result<(Arc<dyn ObjectStore>, Path)> {
+        {
+            let map = self.map.read();
+            let start = &to_resolve[..url::Position::BeforePath];
+
+            let candidates = map
+                .range::<str, _>((Bound::Included(start), Bound::Unbounded))
+                .take_while(|&(base, _)| &base.0[..url::Position::BeforePath] == start);
+
+            let mut longest_len = 0;
```
tustvold, line 183: This is the downside of this formulation, we have to use a BTreeMap which is slower than a HashMap, and if there are multiple (scheme, host) matches we have to scan through them.

tustvold:
I actually plan to tweak this to do path matching on a path segment basis

src/registry.rs
```diff
@@ -0,0 +1,274 @@
...
+struct PathEntry {
+    store: Option<Arc<dyn ObjectStore>>,
+    children: HashMap<String, Self>,
+}
+
+impl PathEntry {
+    /// Lookup a store based on URL path
+    ///
+    /// Returns the store and its path segment depth
+    fn lookup(&self, to_resolve: &Url) -> Option<(&Arc<dyn ObjectStore>, usize)> {
```
tustvold, line 123: This does mean that if people register really long paths, this process will be rather inefficient, however, in most cases I think this should be acceptable.

criccomini (commented):
This looks excellent. I think it will work for our SlateDB use case!

src/registry.rs
```diff
@@ -0,0 +1,274 @@
...
+            let path = path_suffix(to_resolve, depth)?;
+            return Ok((store, path));
+        }
+
+        Err(Error::NotFound.into())
+    }
+}
+
+/// Returns the non-empty segments of a path
+fn path_segments(s: &str) -> impl Iterator<Item = &str> {
```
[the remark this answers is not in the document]
tustvold: We just want the non-empty path segments (as Path strips them)

src/registry.rs
```diff
@@ -0,0 +1,274 @@
...
+            if let Some((store, depth)) = map.get(key).and_then(|entry| entry.lookup(to_resolve)) {
+                let path = path_suffix(to_resolve, depth)?;
+                return Ok((Arc::clone(store), path));
+            }
+        }
+
+        if let Ok((store, path)) = parse_url_opts(to_resolve, std::env::vars()) {
+            let depth = num_segments(to_resolve.path()) - num_segments(path.as_ref());
+
+            let mut map = self.map.write();
```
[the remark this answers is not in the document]
tustvold: Good point, I forgot to handle the race which is why this code is separate

src/registry.rs
```diff
@@ -0,0 +1,328 @@
...
+            return Ok((store, path));
+        }
+
+        Err(Error::NotFound.into())
+    }
+}
+
+/// Extracts the scheme and authority of a URL (components before the Path)
+fn url_key(url: &Url) -> &str {
+    &url[..url::Position::AfterPort]
```
tustvold, line 216: It should be noted this will consider credentials unlike the version in DF - IMO this keeps things simple, and ultimately people shouldn't really be putting credentials in URLs

src/registry.rs
```diff
@@ -0,0 +1,274 @@
...
+    map: RwLock<HashMap<String, PathEntry>>,
+}
+
+#[derive(Debug, Default)]
+struct PathEntry {
+    store: Option<Arc<dyn ObjectStore>>,
+    children: HashMap<String, Self>,
+}
+
+impl PathEntry {
```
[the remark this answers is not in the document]
tustvold: I've added some docs

src/registry.rs
criccomini: This just makes me nervous because what people should do and what they will do are two different things. I worry this will leaked to leaked credentials and security incidents. Could we at least .warn() if we see it?

tustvold: Ok, so imagine we do mask out passwords, now imagine a user fat fingers the password the first time, they then try again with the fixed password, but because the URL has been cached with the incorrect password it will continue to use the old version.

Whilst I accept passwords in URLs are bad, I also can't help feeling masking them introduces somewhat surprising behaviour for relatively limited security improvement - ultimately the user is typing a credential into a SQL query or similar, it's gonna get logged somewhere :sweat_smile:

src/registry.rs
```diff
@@ -0,0 +1,274 @@
...
+        let mut ret = self.store.as_ref().map(|store| (store, 0));
+        let mut depth = 0;
+        for segment in path_segments(to_resolve.path()) {
+            if let Some(e) = current.children.get(segment) {
+                current = e;
+                depth += 1;
+                if let Some(store) = &current.store {
+                    ret = Some((store, depth))
+                }
+            }
```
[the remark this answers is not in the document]
criccomini: To clarify, I'm not saying when we find a child with a store. I'm saying when we find that there's no child for the given segment. Doesn't this code iterate over path segments? So can't we break as soon as we find that a path entry has no child for the segment?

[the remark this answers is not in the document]
tustvold: Oh, yes we can :+1:

src/registry.rs
criccomini: I'm suggesting leaving the code as you have it but calling a WARN telling the user they shouldn't be doing this.

tustvold: IMO a utility crate like object_store is the wrong level for such a warning to be printed, we will get people complaining about it and then wanting to disable it, e.g. if they're user provided URLs that are being used or something... I'd expect such a warning to be implemented as part of whatever frontend users are interacting with

criccomini (approved):
:shipit: looks good!

src/registry.rs
```diff
@@ -0,0 +1,337 @@
...
+    /// assert_eq!(path.as_ref(), "to/object");
+    /// assert!(Arc::ptr_eq(&ret, &bucket3));
+    /// ```
+    fn resolve(&self, url: &Url) -> crate::Result<(Arc<dyn ObjectStore>, Path)>;
+}
+
+/// Error type for [`DefaultObjectStoreRegistry`]
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+enum Error {
```
alamb, line 90: Should this be `pub`? Otherwise people can't match on the error type, only on a string message. Maybe that is ok

src/registry.rs
alamb: I think given this is a trait too people can potentially implement their own version if they have special needs

alamb (approved):
This is looking good to me. I am happy to have @criccomini 's review (made my review strightforward)

Thank you @tustvold and @criccomini

src/registry.rs
tustvold: No it is intentionally crate private/opaque, it gets converted into `crate::Error` a few lines down. It purely exists to make the error handling code slightly more ergonomic.

alamb:
```suggestion
///
/// Crate private/opaque type to make the error handling code  more ergonomic. 
/// Always converted into `crate::Error` when reported externally. 
enum Error {
```

tustvold: I don't think this is necessary, this pattern is used throughout this crate

alamb: 😢  -- I think code that can be understood locally as much as possible is preferable (aka not have to research existing patterns in codebase). I realize this is an opinion. In my mind if the pattern wasn't clear to one reader, others will have the same question

tustvold: I've incorporated this in https://github.com/apache/arrow-rs-object-store/pull/375/commits/a3110f50177cd75b2acdbac85ee8a1e967f657f2

PR #376, opened 2025-05-18
Add option to Azure client to ignore unparsable paths

### LINKED ISSUE #341: Allow ignoring checks for empty segments in paths (double slash)
I was using obstore library which uses this package and encountered an issue with weird paths in our object storage (azure) ( https://github.com/developmentseed/obstore/issues/429 for reference ). We had some paths that contained double slash which is allowed by azure service. These kinds of paths throw an exception when listing contents of a container.

It would be nice to be able to catch this exception or just ignore this and print warning as right now it stops the iterator from yielding paths. Maybe a flag? Or to have implementations for specific services allow these kinds of paths if the service allows them (like azure).

Unfortunately I can't always fix these paths as its coming from other teams so have to deal with messy data lake.

**Additional context**
_(issue linked above)_

ttomasz:

Closes #341.

# Rationale for this change
 

Add config option to Azure client that when listing files will ignore paths that do not match to Path rules i.e. files that would fail `Path::parse`. Currently with path like `foo/bar//baz.txt` we get a panic.

New boolean config option for Azure client: `ignore_unparsable_paths`.

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -383,10 +383,10 @@ pub enum AzureConfigKey {
     FabricClusterIdentifier,
 
     /// When listing objects, ignore paths that do not conform to the Path class assumptions
-    /// 
+    ///
     /// e.g. Azure allows to have paths like `foo/bar//baz` which would throw an error
     /// when trying to parse them into a Path.
-    /// 
+    ///
     /// When set to true, these paths will be ignored.
     ///
     /// Supported keys:
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -698,7 +698,10 @@ impl MicrosoftAzureBuilder {
             "https" => match host.split_once('.') {
                 Some((a, "dfs.core.windows.net")) | Some((a, "blob.core.windows.net")) => {
                     self.account_name = Some(validate(a)?);
-                    if let Some(container) = parsed.path_segments().unwrap().next() {
+                    let container = parsed.path_segments().unwrap().next().expect(
+                        "iterator always contains at least one string (which may be empty)",
+                    );
+                    if !container.is_empty() {
                         self.container_name = Some(validate(container)?);
                     }
                 }
@@ -709,10 +712,11 @@ impl MicrosoftAzureBuilder {
                     // - https://onelake.dfs.fabric.microsoft.com/<workspace>/<item>.<itemtype>/<path>/<fileName>
                     //
                     // See <https://learn.microsoft.com/en-us/fabric/onelake/onelake-access-api>
-                    if let Some(workspace) = parsed.path_segments().unwrap().next() {
-                        if !workspace.is_empty() {
-                            self.container_name = Some(workspace.to_string())
-                        }
+                    let workspace = parsed.path_segments().unwrap().next().expect(
+                        "iterator always contains at least one string (which may be empty)",
+                    );
+                    if !workspace.is_empty() {
+                        self.container_name = Some(workspace.to_string())
                     }
                     self.use_fabric_endpoint = true.into();
                 }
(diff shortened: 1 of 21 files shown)


src/azure/client.rs
```diff
@@ -1540,4 +1565,135 @@ Time:2018-06-14T16:46:54.6040685Z</Message></Error>\r
...
+        };
+        let result = to_list_result(fake_result, None, false).unwrap();
+        assert_eq!(result.common_prefixes.len(), 0);
+        assert_eq!(result.objects.len(), 2);
+        assert_eq!(result.objects[0].location, Path::from("blob0.txt"));
+        assert_eq!(result.objects[1].location, Path::from("blob1.txt"));
+    }
+
+    #[tokio::test]
+    #[should_panic]
```
crepererum, line 1615:
```suggestion
    #[should_panic(expected = "???")]
```

I find it helpful to check the panic message (or a substring of it) to ensure it's providing some useful user error.

crepererum (commented):
Apart from the notes I've left in-line, I'm wondering if we should even panic or should rather return an error. This would be in line with all kinds of unexpected server answers that we handle so far (e.g. weird status codes, broken bodies, TLS failures). The flag would then toggle between "error returned" and "error ignored" instead of "panic" and "no panic". WDYT?

src/azure/client.rs
ttomasz: Added expected error message

src/azure/client.rs
```diff
@@ -1058,6 +1063,9 @@ fn to_list_result(value: ListResultInternal, prefix: Option<&str>) -> Result<Lis
             !matches!(blob.properties.resource_type.as_ref(), Some(typ) if typ == "directory")
                 && blob.name.len() > prefix.len()
         })
+        .map(BlobInternal::try_from)
+        .filter(|parsed| !ignore_unparsable_paths || parsed.is_ok())
+        .map(|parsed| parsed.unwrap())
```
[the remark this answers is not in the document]
ttomasz: Added suggested code

ttomasz:
> Apart from the notes I've left in-line, I'm wondering if we should even panic or should rather return an error. This would be in line with all kinds of unexpected server answers that we handle so far (e.g. weird status codes, broken bodies, TLS failures). The flag would then toggle between "error returned" and "error ignored" instead of "panic" and "no panic". WDYT?

Thanks for the review.

In general it would probably be better to pass the error for user to handle instead of throwing exception/panicking (or as I mentioned in the linked issue there could be a separate interface to get paths as they come i.e. strings instead of parsing them into Path objects). Then the flag wouldn't even be needed.

But I don't know how that could be implemented without backwards incompatible changes.

src/azure/builder.rs
```diff
@@ -180,6 +180,8 @@ pub struct MicrosoftAzureBuilder {
     fabric_cluster_identifier: Option<String>,
     /// The [`HttpConnector`] to use
     http_connector: Option<Arc<dyn HttpConnector>>,
+    /// When listing objects, ignore paths that throw and error when parsing
```
kylebarron:
```suggestion
    /// When listing objects, ignore paths that throw an error when parsing
```
Might also be good to say "See [`AzureConfigKey::IgnoreUnparseablePaths`] for more details

kylebarron (approved):

src/azure/client.rs
```diff
@@ -1058,6 +1063,12 @@ fn to_list_result(value: ListResultInternal, prefix: Option<&str>) -> Result<Lis
             !matches!(blob.properties.resource_type.as_ref(), Some(typ) if typ == "directory")
                 && blob.name.len() > prefix.len()
         })
+        .map(BlobInternal::try_from)
+        .filter_map(|parsed| match parsed {
+            Ok(parsed) => Some(parsed),
+            Err(_) if ignore_unparsable_paths => None,
+            Err(e) => panic!("cannot parse path: {e}"),
```
crepererum, line 1070: I read @kylebarron's :+1: on https://github.com/apache/arrow-rs-object-store/pull/376#issuecomment-2966380560 as "current behavior is there's an error, NOT a panic". So I think the flag should ignore the error, but even if the flag isn't set, you MUST NOT panic.

crepererum (changes requested):

src/azure/client.rs
kylebarron: Oh sorry I was just agreeing with you that it's a good question 😅. I don't recall what the current behavior is

kylebarron: In https://github.com/apache/arrow-rs-object-store/pull/376#issuecomment-2927448265 @ttomasz said returning an error might require a breaking change, though now that we've released 0.12.2 we can now merge breaking changes, right?

crepererum: > though now that we've released 0.12.2 we can now merge breaking changes, right?

right :+1:

TomNicholas:
hi @tustvold - why was this closed? As far as I can tell its still an unresolved issue.

tustvold:
This PR was closed due to inactivity, the issue is still open

Issue #15, opened 2025-03-06
object store: retry / recover after partially reading a streaming response ( fix timeout errors / `error decoding response body` )


alamb:
# Problem Description
if a request fails mid-stream (after we begin to read data) it is not retried and instead an error is returned. The error message, somewhat confusingly, often says `error decoding response body`

Some examples:

> ExternalError(General("ParquetObjectReader::get_byte_ranges error: Generic MicrosoftAzure error: error decoding response body"))

>  Generic S3 error: error decoding response body

# Workaround
You can often work around this error by increasing the network timeout to something longer than the 30s default

### Related Tickets
- there are many related discussions / report https://github.com/apache/arrow-rs-object-store/issues/272
- we see this in DataFusion too https://github.com/apache/datafusion/issues/15067 
- https://github.com/apache/arrow-rs-object-store/issues/53
- https://github.com/apache/arrow-rs/pull/6519
- https://github.com/delta-io/delta-rs/issues/2595

As  @crepererum says on https://github.com/apache/arrow-rs-object-store/issues/272 :

> So long store short: People agree that this would be a good feature to have, but it requires a proper implementation.

# Background

## Streaming ✅ 
Some APIs like [`ObjectStore::get`](https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get) are "streaming" in the sense that they start returning data as soon as it comes back from the network (as opposed to buffering the response before returning to the caller)

This is great for performance as response processing can happen immediately and limits memory usage for large payloads 🏆 

## Retries ✅ 
In order to deal with the intermittent errors that occur processing object store requests, most ObjectStore implementations retry the request if they encounter error (see [retry.rs](https://github.com/apache/arrow-rs/blob/main/object_store/src/client/retry.rs))

## Retries + Streaming ❌ 

However, there is a problem when streaming is mixed with the existing retries. Specifically, if a request fails mid-stream (after some, but not all, of the data has been returned to the client), just retrying the entire request isn't enough because then the client would be potentially be given the same data from the start of the response that it had already been given

# Solution 

**Describe the solution you'd like**

Implementing retries for streaming reads would need something more complicated like retrying the request just for the bytes that hadn't been already read

Any solution for this I think needs:
1. Very good tests / clear documentation

**Describe alternatives you've considered**

@crepererum suggests on https://github.com/apache/arrow-rs-object-store/issues/272 :

[...]

PR #378, opened 2025-05-22
fix: TCP reset retry again and recursively iterates sources 

### LINKED ISSUE #368: Retry does not cover connection errors
According to the docs on [RetryConfig](https://docs.rs/object_store/0.12.0/object_store/struct.RetryConfig.html) the intent expressed seems to be that connection errors should be retried. However the [code](https://docs.rs/crate/object_store/0.12.0/source/src/client/retry.rs#374-403) does not implement this (if the error is not an HTTP 4xx or 5xx, it exits without retrying on L390).

A typical example of a non-retried error I saw in practice is:
```
Service(ObjectStore(Generic { store: "S3", source: RetryError { method: PUT, uri: Some(XXX-REDACTED-XXX), retries: 0, max_retries: 3, elapsed: 801.473µs, retry_timeout: 60s, inner: Http(HttpError { kind: Unknown, source: reqwest::Error { kind: Request, source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })) } }) } }))
```

Is there interest in making the behavior match the documentation? Otherwise it'd be good to edit the documentation to match the behavior. I prefer the former, but either way I'd be happy to chip in.

Jun-Yuan:

Closes #368 .

# Rationale for this change
 
https://github.com/apache/arrow-rs-object-store/issues/368, which was closed but we will reopen it because we believe the bug was not completely fixed.  

Based on the description of https://github.com/apache/arrow-rs-object-store/pull/351
> I didn't see it iterates sources recursively, trying to convert to hyper / io errors.
> So we just need to make sure std::io::ErrorKind::ConnectionReset gets converted to HttpErrorKind::Interrupted

I think this PR is really what the comment meant: recursively iterates sources and convert hyper/io errors, because currently `break;` causes pre-mature exit and the inner `io` errors are not getting converted, and the thus we are still experiencing no retry for ConnectionReset.

Removed `break`s.

No

### DIFF
--- a/src/client/http/connection.rs
+++ b/src/client/http/connection.rs
@@ -102,16 +102,12 @@ impl HttpError {
         // Reqwest error variants aren't great, attempt to refine them
         let mut source = e.source();
         while let Some(e) = source {
-            let mut is_hyper_error = false;
-            let mut is_io_error = false;
-
             if let Some(e) = e.downcast_ref::<hyper::Error>() {
                 if e.is_closed() || e.is_incomplete_message() || e.is_body_write_aborted() {
                     kind = HttpErrorKind::Request;
                 } else if e.is_timeout() {
                     kind = HttpErrorKind::Timeout;
                 }
-                is_hyper_error = true;
             }
             if let Some(e) = e.downcast_ref::<std::io::Error>() {
                 match e.kind() {
@@ -122,10 +118,6 @@ impl HttpError {
                     | std::io::ErrorKind::UnexpectedEof => kind = HttpErrorKind::Interrupted,
                     _ => {}
                 }
-                is_io_error = true;
-            }
-            if is_hyper_error || is_io_error {
-                break;
             }
             source = e.source();
         }
--- a/src/client/http/connection.rs
+++ b/src/client/http/connection.rs
@@ -102,13 +102,16 @@ impl HttpError {
         // Reqwest error variants aren't great, attempt to refine them
         let mut source = e.source();
         while let Some(e) = source {
+            let mut is_hyper_error = false;
+            let mut is_io_error = false;
+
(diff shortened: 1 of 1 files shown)


tustvold:
Perhaps you could come up with a reproducer, I don't follow why this would change the behaviour. If hyper is able to classify the error we use the more precise classification (which may allow it to retry in more cases).

I wonder if what you're running into is actually #15

Jun-Yuan:
> Perhaps you could come up with a reproducer, I don't follow why this would change the behaviour. If hyper is able to classify the error we use the more precise classification (which may allow it to retry in more cases).
> 
> I wonder if what you're running into is actually #15

No we are not running into #15. The error message we had is described in the issue #368
`Service(ObjectStore(Generic { store: "S3", source: RetryError { method: PUT, uri: Some(XXX-REDACTED-XXX), retries: 0, max_retries: 3, elapsed: 801.473µs, retry_timeout: 60s, inner: Http(HttpError { kind: Unknown, source: reqwest::Error { kind: Request, source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" })) } }) } }))`

The “unknown” indicates hyper is failing to classify more precisely so it’s getting swallowed and the proposed fix #351 doesn't completely fix it because of these early breaks, which, when successfully resolving `hyper::Error` it just simple breaks out the loop without even attempting to dive into `io` error. The rest of recursion is totally skipped, thus defeating the purpose of fix #351.

I am unfortunately too tied to write up a reproducer test until some time later.  If this brings any attention to the matter and helps with other users,  that would be great.

PR #383, opened 2025-05-26
Retry streaming get requests (#15)

### LINKED ISSUE #15: object store: retry / recover after partially reading a streaming response ( fix timeout errors / `error decoding response body` )

tustvold:
~_Draft as needs tests_~

Closes #15

# Rationale for this change
 

With this PR we retry streaming get requests. Technically it is still possible to get response body errors on the other methods, however, I chose to just do get requests as these are the highest impact as they have the largest payloads, and we can always do the others later.

### DIFF
--- a/src/client/get.rs
+++ b/src/client/get.rs
@@ -15,20 +15,28 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use std::ops::Range;
-
-use crate::client::header::{header_meta, HeaderConfig};
-use crate::client::HttpResponse;
+use crate::client::header::{get_etag, header_meta, HeaderConfig};
+use crate::client::retry::RetryContext;
+use crate::client::{HttpResponse, HttpResponseBody};
 use crate::path::Path;
-use crate::{Attribute, Attributes, GetOptions, GetRange, GetResult, GetResultPayload, Result};
+use crate::{
+    Attribute, Attributes, GetOptions, GetRange, GetResult, GetResultPayload, ObjectMeta, Result,
+    RetryConfig,
+};
 use async_trait::async_trait;
+use bytes::Bytes;
+use futures::stream::BoxStream;
 use futures::{StreamExt, TryStreamExt};
 use http::header::{
     CACHE_CONTROL, CONTENT_DISPOSITION, CONTENT_ENCODING, CONTENT_LANGUAGE, CONTENT_RANGE,
     CONTENT_TYPE,
 };
 use http::StatusCode;
+use http_body_util::BodyExt;
 use reqwest::header::ToStrError;
+use std::ops::Range;
+use std::sync::Arc;
+use tracing::info;
 
 /// A client that can perform a get request
 #[async_trait]
@@ -38,7 +46,14 @@ pub(crate) trait GetClient: Send + Sync + 'static {
     /// Configure the [`HeaderConfig`] for this client
     const HEADER_CONFIG: HeaderConfig;
 
-    async fn get_request(&self, path: &Path, options: GetOptions) -> Result<HttpResponse>;
+    fn retry_config(&self) -> &RetryConfig;
+
+    async fn get_request(
+        &self,
+        ctx: &mut RetryContext,
+        path: &Path,
+        options: GetOptions,
+    ) -> Result<HttpResponse>;
 }
 
 /// Extension trait for [`GetClient`] that adds common retrieval functionality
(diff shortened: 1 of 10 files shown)


src/client/retry.rs
```diff
@@ -248,81 +253,89 @@ fn body_contains_error(response_body: &str) -> bool {
...
-    backoff: Backoff,
-
-    sensitive: bool,
-    idempotent: Option<bool>,
-    retry_on_conflict: bool,
-    payload: Option<PutPayload>,
-
-    retry_error_body: bool,
+/// Combines a [`RetryableRequest`] with a [`RetryContext`]
+pub(crate) struct RetryableRequestBuilder {
```
tustvold, line 257: This largely exists to keep the changeset down

src/client/get.rs
```diff
@@ -145,40 +156,132 @@ enum GetResultError {
...
+                            let request = ctx
+                                .client
+                                .get_request(&mut ctx.retry_ctx, &ctx.location, options)
+                                .await
+                                .map_err(Self::err)?;
+
+                            let (parts, retry_body) = request.into_parts();
+                            let retry_etag = get_etag(&parts.headers).map_err(Self::err)?;
+
+                            if etag != &retry_etag {
```
crepererum, line 242: :+1:

src/client/get.rs
```diff
@@ -145,40 +156,132 @@ enum GetResultError {
...
-            .map_err(|source| GetResultError::InvalidContentRange { source })?;
+    fn retry_stream(
+        self,
+        body: HttpResponseBody,
+        etag: Option<String>,
+        range: Range<u64>,
+    ) -> BoxStream<'static, Result<Bytes>> {
+        futures::stream::try_unfold(
+            (self, body, etag, range),
+            |(mut ctx, mut body, etag, mut range)| async move {
```
crepererum, line 206: I think that code is rather elegant :+1:

crepererum (approved):

src/client/get.rs
```diff
@@ -145,40 +156,133 @@ enum GetResultError {
...
+                        (Ok(frame), _) => match frame.into_data() {
+                            Ok(bytes) => {
+                                range.start += bytes.len() as u64;
+                                return Ok(Some((bytes, (ctx, body, etag, range))));
+                            }
+                            Err(_) => continue, // Isn't data frame
+                        },
+                        // Retry all response body errors
+                        (Err(e), Some(etag)) if !ctx.retry_ctx.exhausted() => {
+                            let sleep = ctx.retry_ctx.backoff();
```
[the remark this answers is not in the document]
alamb: This strategy will make many classes of transient error less severe which is great

However, it is still possible to get timeout errors when actively reading a streaming response body as the retry mechanism kicks in after timeout, even though progress was made

At some point, it may make sense to detect when progress was made and adjust the remaining retries / policy accordingly. 
- I filed https://github.com/apache/arrow-rs-object-store/issues/386

alamb (commented):
Thank you @tustvold  and @crepererum  -- this is pretty amazing

alamb:
The CI appears to fail after this PR: 
- https://github.com/apache/arrow-rs-object-store/issues/395

Issue #388, opened 2025-05-30
LocalFileSystem: offset for `list_with_offset` can't be identified  / List results *must* be sorted


timwaizenegger:
**Describe the bug**

The `list` / `[list_with_offset](https://docs.rs/object_store/0.12.0/object_store/trait.ObjectStore.html#method.list_with_offset)` functions on LocalFileSystem stores do not (reliably) return results in sorted order.
Now the [interface docs](https://docs.rs/object_store/0.12.0/object_store/trait.ObjectStore.html#method.list_with_offset) for this interface state that:

> Note: the order of returned [ObjectMeta](https://docs.rs/object_store/0.12.0/object_store/struct.ObjectMeta.html) is not guaranteed

so it would be consistent with that but I believe this is not useful:
- offset filtering/listing works by skipping any input objects that are (lexicographically) smaller than the offset
- in order to use offsets to list the all objects in multiple iterations, each iteration must contain a slice of the sorted overall results
- if I have such a slice, the last element will be correct offset to the get the next slice/batch
- if results are in a random order overall, offset listing is impossible to use. I can never find any useful offset value out of the ones I received
  - unless I list everything and sort it then. But that defeats the purpose
- object stores typically list results in sorted order specifically to allow for offset listing
- Having a usable offset listing feature is key for object-sort to achieve its design goal of a stateless API; the only alternative for users is to use a stateful iterator.

**To Reproduce**

generate a set of files:
```
mkdir -p /tmp/manyfiles/
cd /tmp/manyfiles/
    for i in $(seq 1 5000); do
echo "hello world" > "hello world.txt.$i"
done
```
Run this code to show the (random) ordering and show that offset listing isn't usable:
```
#[tokio::main]
async fn main() -> Result<()> {
    let store = LocalFileSystem::new_with_prefix("/tmp/manyfiles")?;

    // list() returns a Stream of Result<ObjectMeta>
    let mut list_stream = store.list(None);

    // pull each ObjectMeta out of the stream
    while let Some(result) = list_stream.next().await {
        let meta = result?;
        // print its path
        println!("-> {} - {}", meta.location.to_string(), meta.last_modified.to_string());
    }



    println!("\nListing files in batches of 10 using list_with_offset...");
    let mut offset: Option<String> = None;
    loop {
        // Choose list or list_with_offset based on whether we have an offset
        let mut batch_stream = if let Some(ref off) = offset {
            store.list_with_offset(None, &Path::from(off.as_str()))
[...]

PR #389, opened 2025-05-30
LocalFileSystem: let WalkDir return sorted results

timwaizenegger:
addresses https://github.com/apache/arrow-rs-object-store/issues/388

Closes https://github.com/apache/arrow-rs-object-store/issues/388

# Rationale for this change
 

offset listing is only usable if the list results are sorted. Only then can I use a member of the results I already have, to make another call that continues my listing.

Adds `sort_by_file_name()` call to `WalkDir` in `list_with_maybe_offset`. As far as I can tell, this is the only change needed to make local file offset listing possible.

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -641,6 +641,7 @@ impl LocalFileSystem {
         let walkdir = WalkDir::new(root_path)
             // Don't include the root directory itself
             .min_depth(1)
+            .sort_by_file_name()
             .follow_links(true);
 
         let maybe_offset = maybe_offset.cloned();


tustvold:
Unfortunately this is a limitation that we can't reliably workaround, some stores such as S3 Express don't return data in sorted order, others don't support it at all. Ultimately list_with_offset is an optimisation for when you want everything after a given prefix.

FWIW #371 adds a way to do pagination, although it intentionally doesn't support filesystems.

src/local.rs
```diff
@@ -641,6 +641,7 @@ impl LocalFileSystem {
         let walkdir = WalkDir::new(root_path)
             // Don't include the root directory itself
             .min_depth(1)
+            .sort_by_file_name()
```
tustvold, line 644: This will list every file in the directory, sort it, and then return it - at the very least this should be explicitly opt-in as directories can get very large - the standard library APIs are intentionally lazy

timwaizenegger: That's true and you don't always need it. But sometimes you absolutely do; offset listing just makes no sense without sorting.

- offset listing: need sorting
- iterator listing: don't care about sorting, higher perf. without sorting

opt-out could be a good option too given those requirements

Issue #398, opened 2025-06-02
Incorrect prefix in `ObjectStoreScheme::parse` for Azure HTTP urls


kylebarron:
**Describe the bug**

In Azure, there is a URL format like:

```
https://overturemapswestus2.blob.core.windows.net/release/path/to/file.txt
```

this is 

```
https://{account_name}.blob.core.windows.net/{container_name}/{file_path}
```

However the container name is included in the file path in `ObjectStoreScheme::parse`. I'd argue this is a bug, but it's at least very confusing.

**To Reproduce**

```rust
#[test]
fn test_prefix() {
    let url = "https://overturemapswestus2.blob.core.windows.net/release/path/to/file.txt";
    let (scheme, path) = ObjectStoreScheme::parse(&url.parse().unwrap()).unwrap();
    assert_eq!(scheme, ObjectStoreScheme::MicrosoftAzure);
    assert_eq!(path, Path::from("path/to/file.txt"));
}
```

The `release/` should be taken off the resulting Path in my opinion.

**Expected behavior**

Expected the container name to be taken off the path before returning. I.e. I expect the path to be the path inside the container.

**Additional context**

https://github.com/developmentseed/obstore/issues/477

kylebarron:
I don't understand what you're suggesting.

> The https parts should be https://{account_name}.blob.core.windows.net/{container_name}/{file_path}

Isn't that what I said?


> In Azure, there is a URL format like:
> 
> ```
> https://overturemapswestus2.blob.core.windows.net/release/path/to/file.txt
> ```
> 
> this is
> 
> ```
> https://{account_name}.blob.core.windows.net/{container_name}/{file_path}
> ```

niltecedu:
Hey @kylebarron  Sorry about the formatting, fixed it now.

Let me come back with a reproducible steps and a test with azurite on #443, currently atleast in downstream libraries in object storage >= 0.12.3; parsing of certain endpoints is causing the GET Request URLs to drop the container name, that is what I observed based on datafusion and delta-rs. #399 and #403 are the only Azure changes I see between 0.12.2 and 0.12.3. 

Where based on the delta-rs debug logs I saw this behavior; 

> For "az://account/container/path" and some other endpoints
> 
The GET Operation ends up used with
> 
> https://storage_acount_name/storage_acount_name?restype=container&comp=list&prefix=test%2Ftesting_delta_table%2F_delta_log%2
> 
instead of 
> 
[...]

PR #399, opened 2025-06-02
Fix azure path parsing

kylebarron:

- Closes https://github.com/apache/arrow-rs-object-store/issues/398.

# Rationale for this change

Is there a spec for each of these URL formats? In my usage of Azure so far it seems like Azure URLs like `https://blob.core.windows.net/` _always_ have the container as the first element of the URL path. Is that true? 

Refer to the tests updated in this PR and ensure that you agree with the edited versions of them.

Fixed parsing paths from Azure URLs.

Yes. Up to consideration whether this is a bug fix or a breaking change.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -92,7 +92,7 @@ impl ObjectStoreScheme {
     /// assert_eq!(scheme, ObjectStoreScheme::Local);
     /// assert_eq!(path.as_ref(), "path/to/my/file");
     ///
-    /// let url: Url = "https://blob.core.windows.net/path/to/my/file".parse().unwrap();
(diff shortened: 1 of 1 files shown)


kylebarron:
The failing emulator tests seem unrelated?

alamb:
Here is a proposed fix for the emulator tests:
- https://github.com/apache/arrow-rs-object-store/pull/401

alamb (commented):
Thank you @kylebarron 

This seems like it does what it is designed to do. I don't know much about azure

Is the path format documented anywhere that I can double check?

alamb:
Thanks again @kylebarron

Issue #395, opened 2025-05-30
Emulator tests are broken on main


alamb:
**Describe the bug**
- After  https://github.com/apache/arrow-rs-object-store/pull/383 is merged, the emulator CI tests are broken

**To Reproduce**
Here is a failure on main: https://github.com/apache/arrow-rs-object-store/actions/runs/15319692559/job/43100487313

```
--- gcp::builder::tests::gcs_test_proxy_url stdout ----

thread 'gcp::builder::tests::gcs_test_proxy_url' panicked at src/gcp/builder.rs:[679](https://github.com/apache/arrow-rs-object-store/actions/runs/15319692559/job/43100487313#step:8:680):14:
called `Result::unwrap_err()` on an `Ok` value: GoogleCloudStorage { client: GoogleCloudStorageClient { config: GoogleCloudStorageConfig { base_url: "https://storage.googleapis.com", credentials: StaticCredentialProvider { credential: GcpCredential { bearer: "" } }, signing_credentials: StaticCredentialProvider { credential: GcpSigningCredential { email: "", private_key: None } }, bucket_name: "foo", retry_config: RetryConfig { backoff: BackoffConfig { init_backoff: 100ms, max_backoff: 15s, base: 2.0 }, max_retries: 10, retry_timeout: 180s }, client_options: ClientOptions { user_agent: None, root_certificates: [], content_type_map: {}, default_content_type: None, default_headers: None, proxy_url: Some("asdf://example.com"), proxy_ca_certificate: None, proxy_excludes: None, allow_http: Parsed(true), allow_insecure: Parsed(false), timeout: Some(Parsed(30s)), connect_timeout: Some(Parsed(5s)), pool_idle_timeout: None, pool_max_idle_per_host: None, http2_keep_alive_interval: None, http2_keep_alive_timeout: None, http2_keep_alive_while_idle: Parsed(false), http2_max_frame_size: None, http1_only: Parsed(true), http2_only: Parsed(false), randomize_addresses: Parsed(true) }, skip_signature: false }, client: HttpClient(Client { accepts: Accepts { gzip: false }, proxies: [Matcher], referer: true, default_headers: {"accept": "*/*", "user-agent": "object_store/0.12.1"}, reqwest::config::RequestTimeout: 30s }), bucket_name_encoded: "foo", max_list_results: None } }
stack backtrace:
   0: __rustc::rust_begin_unwind
[...]

PR #401, opened 2025-06-05
Chore: fix emulator tests due to changes in reqwest

alamb:

- Closes https://github.com/apache/arrow-rs-object-store/issues/395

# Rationale for this change
 
The CI tests started failing on main

I tracked the issue down to reqwests `0.12.16`: https://crates.io/crates/reqwest/0.12.16

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -1526,7 +1526,8 @@ mod tests {
             .with_region("region")
             .with_bucket_name("bucket_name")
             .with_allow_http(true)
-            .with_proxy_url("asdf://example.com")
+            // use invalid url
+            .with_proxy_url("dxx:ddd\\example.com")
             .build()
             .unwrap_err()
             .to_string();
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -622,7 +622,7 @@ mod tests {
         assert_eq!(r.status(), StatusCode::OK);
 
         // Handles redirect loop
-        for _ in 0..10 {
+        for _ in 0..11 {
             mock.push(
                 Response::builder()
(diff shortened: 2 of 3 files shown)


src/client/retry.rs
```diff
@@ -622,7 +622,7 @@ mod tests {
         assert_eq!(r.status(), StatusCode::OK);
 
         // Handles redirect loop
-        for _ in 0..10 {
+        for _ in 0..11 {
```
alamb, line 625: I believe this change is needed due to https://github.com/seanmonstar/reqwest/pull/2664

without this change the test fails because the redirect loop fails to trigger the limit

src/aws/builder.rs
```diff
@@ -1526,7 +1526,8 @@ mod tests {
             .with_region("region")
             .with_bucket_name("bucket_name")
             .with_allow_http(true)
-            .with_proxy_url("asdf://example.com")
+            // use invalid url
```
alamb, line 1529: I suspect, but don't know, that this change is required by 
- https://github.com/seanmonstar/reqwest/pull/2681

the test expects the proxy config to fail, but after the reqwest upgrade it was passing

src/gcp/builder.rs
```diff
@@ -674,7 +674,8 @@ mod tests {
         let err = GoogleCloudStorageBuilder::new()
             .with_service_account_path(service_account_path.to_str().unwrap())
             .with_bucket_name("foo")
-            .with_proxy_url("asdf://example.com")
+            // use invalid url
```
alamb, line 677: Same as above

tustvold (approved):

PR #403, opened 2025-06-09
feat (azure): support for account in `az://` URLs

### LINKED ISSUE #285: object_store: Support `container@account.dfs.core.windows.net/path` URL style for `az` protocol

In adlfs (fsspec), there are [two valid conventions](https://github.com/fsspec/adlfs#quickstart) for specifying a canonical path to Azure Data Lake Gen2 storage, including the account name:

    az://container@account.dfs.core.windows.net/path-part/file
    abfs://container@account.dfs.core.windows.net/path-part/file

While `object_store` supports both the `az` and `abfs` protocols, it does not support specifying the account in the URL for the `az` protocol.
However, it [does support](https://docs.rs/object_store/latest/object_store/azure/struct.MicrosoftAzureBuilder.html#method.with_url) this for the `abfs` protocol.

I would like to be able to use the following URL convention with `object_store`, which improves compatibility with `adlfs`:

    az://container@account.dfs.core.windows.net/path-part/file

The workaround is to remember that I need to use the `abfs` protocol in `object_store` based tools such as `polars`.
However, it would be nice to be able to use the `az` protocol for everything.

ByteBaker:

Closes #285 .

# Rationale for this change
 When using a URL of type `az://container@account.dfs.core.windows.net/path-part/file`, the account name wasn't parsed from the URL. It is now done.

The parser logic used for URL scheme `az://` is now the same as that used for `abfs://`

No. However, they can now use the scheme `az://` just as they used to use `abfs://`

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -658,8 +658,8 @@ impl MicrosoftAzureBuilder {
         };
 
         match parsed.scheme() {
-            "az" | "adl" | "azure" => self.container_name = Some(validate(host)?),
-            "abfs" | "abfss" => {
+            "adl" | "azure" => self.container_name = Some(validate(host)?),
+            "az" | "abfs" | "abfss" => {
                 // abfs(s) might refer to the fsspec convention abfs://<container>/<path>
                 // or the convention for the hadoop driver abfs[s]://<file_system>@<account_name>.dfs.core.windows.net/<path>
                 if parsed.username().is_empty() {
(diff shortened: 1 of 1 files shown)


src/azure/builder.rs
```diff
@@ -658,8 +658,8 @@ impl MicrosoftAzureBuilder {
         };
 
         match parsed.scheme() {
-            "az" | "adl" | "azure" => self.container_name = Some(validate(host)?),
-            "abfs" | "abfss" => {
+            "adl" | "azure" => self.container_name = Some(validate(host)?),
+            "az" | "abfs" | "abfss" => {
```
kylebarron, line 662: Can you add a test of the form `az://container@account.dfs.core.windows.net/path-part/file`?

ByteBaker: Sure. Done.

kylebarron (approved):

kylebarron:
If you merge main, CI should be fixed from https://github.com/apache/arrow-rs-object-store/pull/404

ByteBaker:
@kylebarron can you please run the workflows again?

kylebarron:
I'm not a maintainer myself and don't have permissions to enable workflows.

Issue #385, opened 2025-05-28
Cleanup `ObjectStore` trait: Split out convenience methods


crepererum:
# Abstract
The `ObjectStore` trait -- as designed currently -- is a middle ground of somewhat competing design goals. I think we can do better than that.

# Requirements
The trait serves two groups of API users:

## Object Store Users
Humans and machines that want to use an object store through a unified interface.

They usually like to write only the relevant parts, e.g.:

- "get object `foo/bar.txt`" ==> `store.get("foo/bar.txt").await?`
- "get first 10 bytes of `foo/bar.txt`" ==> `store.get("foo/bar.txt").with_range(..10).await?`

## Object Store Implementations
Parties that implement a new object store or wrap an existing one. They usually want to implement the methods that determine the behavior of the object store, ideally without surprises (like opinionated default implementations).

# Status Quo
Due to this dual nature the trait has accumulated an increasing number of methods, a bunch of them with default implementation. Let's have a look at the "get" methods:

https://github.com/apache/arrow-rs-object-store/blob/0c3152c709d5101bc2346c49fff5c94e033b8e71/src/lib.rs#L633-L662

All except for `get_ranges` basically map to `get_opts` and there is no reason for a store to override any of the defaults. And even  for `get_ranges` we could come up with a sensible mapping to `get_opts` if the range parameter would support multi-ranges similar to HTTP.

Now let's look at "rename":

https://github.com/apache/arrow-rs-object-store/blob/0c3152c709d5101bc2346c49fff5c94e033b8e71/src/lib.rs#L773-L782

https://github.com/apache/arrow-rs-object-store/blob/0c3152c709d5101bc2346c49fff5c94e033b8e71/src/lib.rs#L793-L799

I think it is out of question that a store implementation should definitely override these if there's any way to perform key renames without a full "get + put".

# Proposal
I propose to remove all default implementations from the trait and only have a single method per operation, so "get" would look like this:

```rust
async fn get(&self, location: &Path, options: GetOptions) -> Result<GetResult>;
```

Note that the `location` is NOT part of `options` so that `GetOptions` can still implement `Default` and a user only needs to specify the options of interest:

```rust
// get bytes of file
store.get(
	&location,
	Default::default(),
).await?.bytes().await?;

// get range
store.get(
	&location,
	GetOptions{
		range: (..10).into(),
		..Default::default()
	},
).await?.bytes().await?;
```

A similar mapping can be done for "rename".

[...]

PR #405, opened 2025-06-16
refactor: introduce `ObjectStoreExt` trait

crepererum:
- First part of #385.

# Rationale for this change
See #385.

- the trait
- `put` method migrated

1. users may need to import another trait
2. users no longer can / have to implement `put`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -252,11 +252,11 @@
 //!
 //! # Put Object
 //!
-//! Use the [`ObjectStore::put`] method to atomically write data.
+//! Use the [`ObjectStoreExt::put`] method to atomically write data.
 //!
 //! ```ignore-wasm32
 //! # use object_store::local::LocalFileSystem;
-//! # use object_store::{ObjectStore, PutPayload};
+//! # use object_store::{ObjectStore, ObjectStoreExt, PutPayload};
 //! # use std::sync::Arc;
 //! # use object_store::path::Path;
 //! # fn get_object_store() -> Arc<dyn ObjectStore> {
@@ -364,7 +364,7 @@
 //!
 //! ```ignore-wasm32
 //! # use object_store::local::LocalFileSystem;
-//! # use object_store::{ObjectStore, PutPayloadMut};
+//! # use object_store::{ObjectStore, ObjectStoreExt, PutPayloadMut};
 //! # use std::sync::Arc;
 //! # use bytes::Bytes;
 //! # use tokio::io::AsyncWriteExt;
@@ -613,19 +613,24 @@ pub type DynObjectStore = dyn ObjectStore;
 pub type MultipartId = String;
 
 /// Universal API to multiple object store services.
+///
+/// For more convience methods, check [`ObjectStoreExt`].
+///
+/// # Contract
+/// This trait is meant as a contract between object store implementations
+/// (e.g. providers, wrappers) and the `object_store` crate itself and is
+/// intended to be the minimum API required for an object store.
+///
+/// The [`ObjectStoreExt`] acts as an API/contract between `object_store`
+/// and the store users and provides additional methods that may be simpler to use but overlap
+/// in functionality with `ObjectStore`
 #[async_trait]
 pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
-    /// Save the provided bytes to the specified location
+    /// Save the provided `payload` to `location` with the given options
     ///
     /// The operation is guaranteed to be atomic, it will either successfully
     /// write the entirety of `payload` to `location`, or fail. No clients
     /// should be able to observe a partially written object
-    async fn put(&self, location: &Path, payload: PutPayload) -> Result<PutResult> {
-        self.put_opts(location, payload, PutOptions::default())
-            .await
-    }
-
-    /// Save the provided `payload` to `location` with the given options
     async fn put_opts(
         &self,
         location: &Path,
@@ -635,7 +640,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
 
     /// Perform a multipart upload
     ///
-    /// Client should prefer [`ObjectStore::put`] for small payloads, as streaming uploads
+    /// Client should prefer [`ObjectStoreExt::put`] for small payloads, as streaming uploads
     /// typically require multiple separate requests. See [`MultipartUpload`] for more information
     ///
     /// For more advanced multipart uploads see [`MultipartStore`](multipart::MultipartStore)
@@ -646,7 +651,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
 
     /// Perform a multipart upload with options
     ///
-    /// Client should prefer [`ObjectStore::put`] for small payloads, as streaming uploads
+    /// Client should prefer [`ObjectStore::put_opts`] for small payloads, as streaming uploads
     /// typically require multiple separate requests. See [`MultipartUpload`] for more information
     ///
     /// For more advanced multipart uploads see [`MultipartStore`](multipart::MultipartStore)
@@ -665,7 +670,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// ```ignore-wasm32
     /// # use object_store::local::LocalFileSystem;
     /// # use tempfile::tempdir;
-    /// # use object_store::{path::Path, ObjectStore};
+    /// # use object_store::{path::Path, ObjectStore, ObjectStoreExt};
     /// async fn get_example() {
     ///     let tmp = tempdir().unwrap();
     ///     let store = LocalFileSystem::new_with_prefix(tmp.path()).unwrap();
@@ -699,7 +704,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// ```ignore-wasm32
     /// # use object_store::local::LocalFileSystem;
     /// # use tempfile::tempdir;
-    /// # use object_store::{path::Path, ObjectStore, GetOptions};
+    /// # use object_store::{path::Path, ObjectStore, ObjectStoreExt, GetOptions};
     /// async fn get_opts_example() {
     ///     let tmp = tempdir().unwrap();
     ///     let store = LocalFileSystem::new_with_prefix(tmp.path()).unwrap();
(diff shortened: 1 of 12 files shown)


crepererum:
Technically the extension crate doesn't need `async_trait` because it is never dyn-dispatched (it may use a dyn-dispatched `ObjectStore` though). However our MSRV doesn't support that yet. So there are two options:

- bump MSRV to 1.75
- use `async_trait` for the extension crate as well, resulting in another indirect call. This is likely never going to be noticeable though, since it's not a hot path.

src/lib.rs
```diff
@@ -901,6 +894,30 @@ macro_rules! as_ref_impl {
...
+    /// Save the provided bytes to the specified location
+    ///
+    /// The operation is guaranteed to be atomic, it will either successfully
+    /// write the entirety of `payload` to `location`, or fail. No clients
+    /// should be able to observe a partially written object
+    fn put(
+        &self,
+        location: &Path,
+        payload: PutPayload,
+    ) -> impl Future<Output = Result<PutResult>> + Send;
```
alamb: If we changed this to return a `BoxFuture` would it avoid the need to change MSRV? 

Perhaps that is what you mean by avoids dyn dispatch.

src/lib.rs
```diff
@@ -901,6 +894,30 @@ macro_rules! as_ref_impl {
 as_ref_impl!(Arc<dyn ObjectStore>);
 as_ref_impl!(Box<dyn ObjectStore>);
 
+/// Extension trait for [`ObjectStore`] with convinience functions.
```
alamb, line 1205: To ease the transition I suggest we augment the documentation here with the following additional information

# 1. Explain the intended use / goal of this trait

As I understand it, the idea of this trait is to make it more clear what functions *must* be provided by an `ObjectStore` implementation and which are mostly implemented in terms of the others. 

Maybe we can also make it clear to implementers that the default implementations of `ObjectStoreExt` may not be optimal for their uscase

However, I may not understand the full subtely - perhaps we can add more info from https://github.com/apache/arrow-rs-object-store/issues/385 as appropriate

# 2. Add a note about the migration plan / tips to help on upgrade

For example, we could add a section on "upgrade notes" and explain "if you implemented `put` for your `ObjectStore`, if it has an implementation different than the default move that implementation to `ObjectStoreExt`. 

Also it would be good to highlight any planned changes that were forthcoming (like "we plan to move the X, Y and Z functions to ObjectStoreExt over time as well")

# 3. An example 

I know this sounds silly, but I think especially given `impl Future --> ...` signature it is non obvious to the causal Rust programmer that this means the impl should have an `async` fn

alamb (approved):
> Changes
> users may need to import another trait
> users no longer can / have to implement put

I think this makes sense, but will potentially be disruptive  to downstream crates (will have to figure out they need now to `use ObjectStoreExt`). 

I think the Rust compiler error messages will do a reasonable job of telling them what is going on, but adding some additional documentation could help make the transition easier. 


> use async_trait for the extension crate as well, resulting in another indirect call. This is likely never going to be noticeable though, since it's not a hot path

I think this is more consistent with the rest of the crate and would be preferred suggestion

> bump MSRV to 1.75

It seems like 1.75 was [released 1.5 years ago](https://blog.rust-lang.org/2023/12/28/Rust-1.75.0/), which is a long time in Rust terms but maybe we can avoid making the next upgrade too painful


cc @tustvold

crepererum:
I plan to migrate other methods in follow-up PRs. I just wanted to keep this PR rather simple.

src/lib.rs
crepererum: I've added a motivation text to both traits and also changed the implementation to `async_trait`. Does this look more useful to you?

src/lib.rs
```diff
@@ -587,19 +587,22 @@ pub type DynObjectStore = dyn ObjectStore;
 pub type MultipartId = String;
 
 /// Universal API to multiple object store services.
+///
+/// For more convience methods, check [`ObjectStoreExt`].
+///
+/// # Contract
+/// This trait is meant as a contract between object store implementations
+/// (e.g. providers, wrappers) and the `object_store` crate itself.
```
alamb: Is another key goal that `ObjectStore` is meant to be the minimal required API? If so perhaps we can add something like

```suggestion
/// This trait is meant as a contract between object store implementations
/// (e.g. providers, wrappers) and the `object_store` crate itself and is 
/// intended to be the minimum API required for an object store.
```

src/lib.rs
```diff
@@ -587,19 +587,22 @@ pub type DynObjectStore = dyn ObjectStore;
...
+///
+/// The [`ObjectStoreExt`] acts as an API/contract between `object_store`
+/// and the store users.
```
alamb:
```suggestion
/// and the store users and provides additional methods that may be simpler to use but overlap
/// in functionality with `ObjectStore`
```

alamb (approved):

tustvold (commented):
Taken a quick look, unfortunately very short on time this week.

I think the major question from me is if this is premature, in particular we don't currently have a list_opts or copy_opts method, and in the former case it isn't immediately obvious what this would look like. Similarly methods like get_ranges have specializations that make a material difference.

The reason I've held off doing something like this before is that some methods are not obvious how they'd fit into this, I'd feel more comfortable to have a plan for these before committing to this approach

crepererum:
@tustvold do you wanna put a veto in against this change? Otherwise @alamb and myself are going to push for this and follow-up clean-up for other methods for the upcoming 0.13 release. I also found a good prior art now which is [`rand::Rng`](https://docs.rs/rand/latest/rand/trait.Rng.html) VS [`rand::RngCore`](https://docs.rs/rand/latest/rand/trait.RngCore.html).

tustvold:
No objections from me, sorry didn't realise this was waiting on me

src/lib.rs
```diff
@@ -901,6 +900,40 @@ macro_rules! as_ref_impl {
 as_ref_impl!(Arc<dyn ObjectStore>);
 as_ref_impl!(Box<dyn ObjectStore>);
 
+/// Helper module to [seal traits](https://predr.ag/blog/definitive-guide-to-sealed-traits-in-rust/).
+mod private {
+    pub trait Sealed {}
+
+    impl<T> Sealed for T where T: super::ObjectStore + ?Sized {}
```
mbrobbel: Maybe I'm confused but my understanding is that we want to seal the `ObjectStoreExt` trait?
```suggestion
    impl<T> Sealed for T where T: super::ObjectStoreExt + ?Sized {}
```

src/lib.rs
```diff
@@ -901,6 +900,40 @@ macro_rules! as_ref_impl {
...
+    /// The operation is guaranteed to be atomic, it will either successfully
+    /// write the entirety of `payload` to `location`, or fail. No clients
+    /// should be able to observe a partially written object
+    async fn put(&self, location: &Path, payload: PutPayload) -> Result<PutResult>;
+}
+
+#[async_trait]
+impl<T> ObjectStoreExt for T
+where
+    T: ObjectStore + private::Sealed + ?Sized,
```
mbrobbel: However, with the blanket impl here, we don't even have to seal the trait?

src/lib.rs
crepererum: that's correct

src/lib.rs
crepererum: Because you could otherwise implement `ObjectStoreExt` for something that doesn't implement `ObjectStore` (because these impls don't overlap), like here:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ede4714a54f997aee78219597dc46b81

mbrobbel: `ObjectStore` should be a supertrait of `ObjectStoreExt`:

```rust
trait ObjectStoreExt: ObjectStore {}
```

```
error[E0277]: the trait bound `u8: ObjectStore` is not satisfied
  --> src/main.rs:12:25
   |
12 | impl ObjectStoreExt for u8 {}
   |                         ^^ the trait `ObjectStore` is not implemented for `u8`
   |
   = help: the trait `ObjectStore` is implemented for `Store`
note: required by a bound in `ObjectStoreExt`
  --> src/main.rs:7:23
   |
 7 | trait ObjectStoreExt: ObjectStore {}
   |                       ^^^^^^^^^^^ required by this bound in `ObjectStoreExt`
```

src/lib.rs
```diff
@@ -901,6 +900,40 @@ macro_rules! as_ref_impl {
...
+}
+
+/// Extension trait for [`ObjectStore`] with convinience functions.
+///
+/// See "contract" section within the [`ObjectStore`] documentation for more reasoning.
+///
+/// # Implementation
+/// You MUST NOT implement this trait yourself. It is automatically implemented for all [`ObjectStore`] implementations.
+#[async_trait]
+pub trait ObjectStoreExt: private::Sealed {
```
mbrobbel:
```suggestion
pub trait ObjectStoreExt: ObjectStore {
```

src/lib.rs
crepererum: that has the same effect but your solution is cleaner and provides better UX :+1:  . I'll change my PR.

src/lib.rs
kylebarron: I like the split here between implementors and users of the crate

src/lib.rs
```diff
@@ -901,6 +900,40 @@ macro_rules! as_ref_impl {
...
+    /// Save the provided bytes to the specified location
+    ///
+    /// The operation is guaranteed to be atomic, it will either successfully
+    /// write the entirety of `payload` to `location`, or fail. No clients
+    /// should be able to observe a partially written object
+    async fn put(&self, location: &Path, payload: PutPayload) -> Result<PutResult>;
```
kylebarron: I haven't followed discussion; is `put` the only method intended to put on this ext trait? for example get_ranges has a default implementation but isn't in this ObjectStoreExt trait

kylebarron:
> * bump MSRV to 1.75

https://github.com/apache/arrow-rs-object-store/pull/515 bumped msrv. Does that change this pr at all?

src/lib.rs
crepererum: there gonna be more methods that move (e.g. `get`), but I wanted to keep this PR kinda manageable.

crepererum:
> > * bump MSRV to 1.75
> 
> #515 bumped msrv. Does that change this pr at all?

no longer, will update

src/lib.rs
kylebarron: oh I see. How does it work to override methods in `ObjectStoreExt`? For example `get_ranges` currently hardcodes the coalesce size and in obstore I want to make that configurable. I guess it would be simplest to use my own helper function that calls `get_range`?

src/lib.rs
crepererum: MSRV got bumped in the meantime, so this is no longer relevant

src/lib.rs
crepererum: done

src/lib.rs
crepererum: no longer relevant as per https://github.com/apache/arrow-rs-object-store/pull/405#discussion_r2490803375

src/lib.rs
crepererum: I'm not sure how far I get with the `get_range(s)` family in this release but my long term plan is to fold them into `get_opts` as well. I think we need to somewhat rework the return type of `get` & Co to be more flexible, but that needs a bit more brain work.

Issue #339, opened 2025-04-21
Rename PutMultipartOpts to PutMultipartOptions


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

As pointed out on https://github.com/apache/arrow-rs-object-store/pull/324/files#r2052563762 by @stayrascal, PutMultipartOpts is inconsistent with the naming of other such options structs

**Describe the solution you'd like**

We should rename `PutMultipartOpts` to `PutMultipartOptions`.

To avoid this being a breaking change we should add

```
#[doc(hidden)]
#[deprecated(note = "Use PutMultipartOptions")]
pub type PutMultipartOpts = PutMultipartOptions;
```

**Describe alternatives you've considered**

**Additional context**

Issue #367, opened 2025-05-16
Release object store `0.13.0` (breaking) - Target Nov 2025


crepererum:
I wonder if after 0.12.2 (see #307) we should have a breaking release again.

Community Requested Items
- [ ] https://github.com/apache/arrow-rs-object-store/issues/295
- [ ] https://github.com/apache/arrow-rs-object-store/issues/18
- [x] https://github.com/apache/arrow-rs-object-store/issues/385

Release tasklist:
- [ ] Version + changelog: https://github.com/apache/arrow-rs-object-store/pull/581
- [ ] Test with arrow-rs: https://github.com/apache/arrow-rs/pull/9017
- [ ] Make release candidate / start vote thread
- [ ] Approval and release to crates.io

alamb:
I support a 0.13.0 release -- 0.12.0  was released on March 9, 2025: https://crates.io/crates/object_store/0.12.0

We are solidly in the realm of being able to make a new breaking release. Maybe we should do so in early June?

alamb:
Any thoughts @Xuanwo or @criccomini ?

criccomini:
SGTM!

alamb:
> I'm not sure about the schedule. Are there people currently working on those two issues?

I think @crepererum  is thinking of doing so (though I think he is out next week so probably the week after)

tustvold:
> Maybe we should do so in early June?

I think we should do at least one more patch release (perhaps in addition to one shortly to fix docs.rs) and then a breaking release late June to sync with arrow-rs' July breaking release.

alamb:
July for a 0.13 breaking release sounds good to me

tustvold:
Currently the only breaking change that has been merged is https://github.com/apache/arrow-rs-object-store/pull/407

This is hardly an urgent change, and so no objections to me with reverting it temporarily and doing another patch release

alamb:
@crepererum and I discussed this today, here is a ticket to track creating a new non breaking release
- https://github.com/apache/arrow-rs-object-store/issues/428

johnnyg:
Is this still the next planned release? Or will there be a 0.12.4 release?

alamb:
@johnnyg, are you waiting on anything in particular?

johnnyg:
[...]

PR #406, opened 2025-06-16
refactor: `PutMultiPartOpts` => `PutMultiPartOptions`

crepererum:
See #339. (not closed, because old alias is still around)

# Rationale for this change
Consistency.

Rename.

Users should use the new name, but the old name is kept as a deprecated alias for now.


src/lib.rs
```diff
@@ -1234,9 +1234,14 @@ impl From<Attributes> for PutOptions {
     }
 }
 
+// See <https://github.com/apache/arrow-rs-object-store/issues/339>.
+#[doc(hidden)]
+#[deprecated(note = "Use PutMultipartOptions")]
```
alamb: Can we also please add an `since` annotation so we know in the future when to remove this? 

For example like this
https://github.com/apache/arrow-rs/blob/639b5bb93e5a152a437b93a25ead8095a3866a9b/parquet/src/file/footer.rs#L47

I believe it will be released in `0.13.0`:
- https://github.com/apache/arrow-rs-object-store/issues/367

alamb (approved):
Thank you @crepererum  - this looks good to me

alamb:
FWIW I think we could also close the ticket (even though there is an alias) but I don't feel strongly
- https://github.com/apache/arrow-rs-object-store/issues/339

src/lib.rs
crepererum: done

PR #410, opened 2025-06-19
Retry requests when status code is 429

### LINKED ISSUE #309: Retry on 429s and equivalents
We are trying to do bulk ingestion on a LanceDB table in GCS, and hitting 429 "TooManyRequests" errors, which are not being retried.

I'd like 429 responses to always generate a retry with truncated exponential backoff.

This could theoretically be handled by users of this library like LanceDB or even application code. But it seems to me that the best place to address is here.

**Additional context**
GCS [guidance](https://cloud.google.com/storage/docs/request-rate):
>You should [retry your requests](https://cloud.google.com/storage/docs/retry-strategy) with exponential backoff when:
>  * Receiving errors with 408 and 429 response codes.
>  * Receiving errors with 5xx response codes.

In AWS this is a 503 with SlowDown as the reason: https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

### LINKED ISSUE #425: S3 store fails without retrying
**Describe the bug**
When doing a high amount of requests on Tigris (S3 clone), I see this sometimes:

```
called `Result::unwrap()` on an `Err` value: ObjectStoreError(Generic { store: "S3", source: RetryError(RetryErrorImpl { method: GET, uri: Some(https://t3.storage.dev/***/slatedb-bencher_20_128/compacted/01JZ9FMS2RJ801EXBKYCX43J95.sst), retries: 0, max_retries: 10, elapsed: 18.220024ms, retry_timeout: 180s, inner: Status { status: 429, body: Some("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<Error><Code>ErrTooManyRequest</Code><Message>Too many simultaneous request count</Message><Resource>/***/slatedb-bencher_20_128/compacted/01JZ9FMS2RJ801EXBKYCX43J95.sst</Resource><RequestId>1751591317331388824</RequestId><Key>slatedb-bencher_20_128/compacted/01JZ9FMS2RJ801EXBKYCX43J95.sst</Key><BucketName>***</BucketName></Error>") } }) })
```

@Barre reported the same issue. The confusing thing is it says `retries: 0` but `max_retries: 10`. Shouldn't it keep retrying until it hits the max?

**To Reproduce**

Get rate limited on an object store.

**Expected behavior**

I expect the S3ObjectSore to back off and retry.

**Additional context**

I tried both 0.12.0 and 0.12.2. Both exhibit the same behavior.

paraseba:
Closes #309
Closes #425

# Rationale for this change

Status code 429 is what HTTP uses to indicate the client has sent too many requests in a given amount of time. It may sound counter-intuitive to retry in this situation, but that's why we use an exponential backoff mechanism. It gives the server the opportunity to recover, without failing the requests immediately.

The retry mechanism already works in object stores like S3 because they return a server error. But without this change, we are not handling GCS properly. GCS returns a client error `429 Too Many Requests` instead. This change enables retries on this response too.

A more advanced retry mechanism would use the optional response header `Retry-After`, but that is beyond the scope of this PR.

### DIFF
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -396,6 +396,7 @@ impl RetryableRequest {
                         let status = r.status();
                         if ctx.exhausted()
                             || !(status.is_server_error()
+                                || status == StatusCode::TOO_MANY_REQUESTS
                                 || (self.retry_on_conflict && status == StatusCode::CONFLICT))
                         {
                             let source = match status.is_client_error() {


paraseba:
Is the unrelated Clippy error a blocker?

kylebarron:
you should be able to merge in main https://github.com/apache/arrow-rs-object-store/pull/418

paraseba:
@kylebarron 

> you should be able to merge in main #418

done

mbrobbel (approved):

Issue #48, opened 2024-09-10
Improve `reqwests` error display


crepererum:
Originally reported in https://github.com/apache/arrow-rs/issues/5882#issuecomment-2340037342 :

For apache/arrow-rs#5882 we only report `Generic S3 error: error decoding response body`. We should try to improve the error message. It seems that the `Display` impl. is not super helpful and we might wanna use `Debug` instead, see https://github.com/seanmonstar/reqwest/issues/2373

PR #422, opened 2025-07-02
fix: expose source of `RetryError`

crepererum:
\-

# Rationale for this change
This used to work once but got broken during refactoring. Added a test to catch regressions.

See https://github.com/apache/arrow-rs-object-store/issues/48 where -- I think -- we decided that instead of printing entire error chains in `Display` the API users shall walk the error chain. This is especially relevant for error type that we do NOT control, like upstream `request` types.

Manual `Error::source` implementation + test.

`RetryError` cannot be walked again.

### DIFF
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -32,7 +32,7 @@ use tracing::info;
 use web_time::{Duration, Instant};
 
 /// Retry request error
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug)]
 pub struct RetryError(Box<RetryErrorImpl>);
 
 /// Box error to avoid large error variant
@@ -63,6 +63,12 @@ impl std::fmt::Display for RetryError {
             )?;
         }
         write!(f, " - {}", self.0.inner)
+    }
+}
+
+impl std::error::Error for RetryError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        Some(&self.0.inner)
     }
 }
 
@@ -504,6 +510,7 @@ mod tests {
     use hyper::header::LOCATION;
     use hyper::Response;
     use reqwest::{Client, Method, StatusCode};
+    use std::error::Error;
     use std::time::Duration;
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/client/retry.rs
```diff
@@ -32,7 +32,7 @@ use tracing::info;
 use web_time::{Duration, Instant};
 
 /// Retry request error
-#[derive(Debug, thiserror::Error)]
+#[derive(Debug)]
```
alamb, line 35: this is technically a breaking API change, I think

However the next release appears to be slated to be a breaking one so that would be ok: https://github.com/apache/arrow-rs?tab=readme-ov-file#release-versioning-and-schedule

We could potentially put in a workaround like implement a non standard way to access the source if we want to put it in before the next breaking change

src/client/retry.rs
```diff
@@ -657,11 +664,15 @@ mod tests {
...
         }
 
-        let e = do_request().await.unwrap_err().to_string();
+        let e = do_request().await.unwrap_err();
         assert!(
-            e.contains(" after 2 retries, max_retries: 2, retry_timeout: 1000s  - Server returned non-2xx status code: 502 Bad Gateway"),
+            e.to_string().contains(" after 2 retries, max_retries: 2, retry_timeout: 1000s  - Server returned non-2xx status code: 502 Bad Gateway"),
             "{e}"
         );
+        assert_eq!(
```
alamb, line 673: I think it would help to add a comment here saying the use of `e.source()` was intentional

Perhaps something like

```suggestion
        // verify e.source() is available as well for users who need programatic access
        assert_eq!(
```

(also in the one below)

alamb (approved):
Thank you @crepererum

src/client/retry.rs
crepererum: I disagree because:

1. this type isn't public, hence it can only be used as via the trait `dyn Error`. You CANNOT downcast it.
2. All this derive does is to implement `std::error::Error`. Also see [thiserror docs](https://docs.rs/thiserror/latest/thiserror/#details) that state "Thiserror deliberately does not appear in your public API." I.e. removing this is a change of an implementation detail, NOT a public API.

src/client/retry.rs
crepererum: done

src/client/retry.rs
alamb: Makes sense to me -- thank you for the clarification

PR #427, opened 2025-07-08
Add conditional deletes

### LINKED ISSUE #298: add support for conditional deletes
Conditional deletes have been introduced to S3 Express now
https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-express-one-zone-conditional-deletes/

ankrgyl:

Closes #298

# Rationale for this change
 
It's useful to have conditional deletes while implementing concurrent primitives on object stores, like distributed locks.

Mirror the conditional put implementation to support conditional deletes

Yes a new `delete_opts` method that accepts deletion opts.


ankrgyl:
> As far as I know, aws only support delete with if match on s3 express.

Ah yes that is correct!

rbehjati:
It seems that as of September 16th, if-match is supported for S3 general purpose buckets ([source](https://aws.amazon.com/about-aws/whats-new/2025/09/amazon-s3-conditional-deletes-s3-general-purpose-buckets/)). Can the work on this PR be continued now?

PR #432, opened 2025-07-09
Add LoggingStore wrapper implementation

### LINKED ISSUE #380: [Question] Would contribution of logging wrapper and a "readahead" wrapper for ObjectStore be wanted?

We have developed two `ObjectStore` wrapper implementations in a similar style to `LimitStore`, `PrefixStore`, etc. which we would like to contribute to this repository if they are desired. Based on the description below, would there be enough of a use-case for the creation of two PRs for the implementations?

**Describe your question**
The two implementations are as follows:

*** `LoggingObjectStore`
Our implementation, which a PR would be based off is [here](https://github.com/gchq/sleeper/blob/3ffa40ba33172365e78fe138c7d4f477550848f0/rust/compaction/src/store.rs#L72).

`LoggingObjectStore` Wraps another `ObjectStore` implementation and writes all operations "GETs", "PUTs", "LIST", etc, to Rust's standard logger.

This was extremely helpful when debugging an application we had written that requested many files from Amazon S3 in small chunks. It was helpful to be able to see exactly when, where and what file parts were being requested by our application to gain a better understanding of what was going on deep inside some library code.

*** `ReadaheadStore`
Our implementation, which a PR would be based off is [here](https://github.com/gchq/sleeper/blob/develop/rust/compaction/src/readahead.rs).

This wraps another `ObjectStore` and attempts to re-use opened data streams on an object during GET operations. Inspired by the functionality in [Apache Hadoop's Hadoop-AWS module](https://hadoop.apache.org/docs/r3.4.1/hadoop-aws/tools/hadoop-aws/index.html), the readahead store keeps a data stream open once a client has finished reading from it. When a new GET operation is requested on the same object, if the starting read position is within a configurable distance of the last read, we can re-use the existing stream instead of opening a new stream.

This can drastically reduce the number of new GET operations that have to be started against the wrapped object store in sequential reads of objects. This can have a performance improvement due to fewer network requests being created.

If a new GET operation is requested that is before the position of a previous stream, or too far beyond the previous position, a new request is performed on the underlying store. The number of concurrent open streams per object, maximum time-to-live and maximum safe "readahead" are configurable.

**Additional context**
We would like to contribute these two implementations to the wider community.

m09526:

Closes #380.

# Rationale for this change
 

This adds an ObjectStore wrapping implementation that logs all calls being made to the wrapped implementation. This is to aid in debugging. It is particularly useful when object stores are used by 3rd party code such as Apache DataFusion so the developer can determine what remote object calls the 3rd party code is making.

LoggingStore wrapper is added.

Yes, the new implementation is part of the public API.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ futures = "0.3"
 http = "1.2.0"
 humantime = "2.1"
 itertools = "0.14.0"
+log = "0.4.27"
 parking_lot = { version = "0.12" }
 percent-encoding = "2.1"
 thiserror = "2.0.2"
@@ -83,10 +84,11 @@ integration = ["rand"]
 hyper = { version = "1.2", features = ["server"] }
 hyper-util = "0.1"
 rand = "0.9"
-tempfile = "3.1.0"
 regex = "1.11.1"
 # The "gzip" feature for reqwest is enabled for an integration test.
 reqwest = { version = "0.12", features = ["gzip"] }
+tempfile = "3.1.0"
+testing_logger = "0.1.1"    
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
 wasm-bindgen-test = "0.3.50"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -521,6 +521,7 @@ pub mod http;
 pub mod limit;
 #[cfg(all(feature = "fs", not(target_arch = "wasm32")))]
 pub mod local;
+pub mod logging;
 pub mod memory;
 pub mod path;
 pub mod prefix;
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -16,16 +16,17 @@
 // under the License.
 
 //! An object store that logs calls to the wrapped implementation.
-
 use crate::{
     path::Path, GetOptions, GetRange, GetResult, ListResult, MultipartUpload, ObjectMeta,
     ObjectStore, PutMultipartOptions, PutOptions, PutPayload, PutResult, Result, UploadPart,
 };
 use async_trait::async_trait;
 use futures::stream::BoxStream;
-use log::info;
+use log::debug;
 
-/// An [`ObjectStore`] wrapper that logs operations made to the wrapped store.
+/// An [`ObjectStore`] wrapper that logs operations made to the wrapped store. The logs are written using the ['log'] crate.
+/// 
+/// Logs are written at the "debug" logging level.
 #[derive(Debug)]
 pub struct LoggingStore<T: ObjectStore> {
     store: T,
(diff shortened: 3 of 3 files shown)


Cargo.toml
```diff
@@ -37,6 +37,7 @@ futures = "0.3"
 http = "1.2.0"
 humantime = "2.1"
 itertools = "0.14.0"
+log = "0.4.27"
```
tustvold: Can we use tracing instead given this crate already depends on it?

m09526: Thanks for the suggestion! The PR has been updated to reflect this.

src/trace.rs
```diff
@@ -0,0 +1,578 @@
...
+            }
+        }
+        self.store.get_opts(location, options).await
+    }
+
+    async fn head(&self, location: &Path) -> Result<ObjectMeta> {
+        debug!(
+            "{} head request for {}/{}",
+            self.prefix, self.path_prefix, location
+        );
```
mbrobbel: Instead of debug level events, maybe we should use spans?

asubiotto:
FWIW, https://github.com/datafusion-contrib/datafusion-tracing/tree/main/instrumented-object-store exists and seems to work well with our use of datafusion. Just wanted to point it out to reduce duplication of efforts.

geoffreyclaude:
> FWIW, https://github.com/datafusion-contrib/datafusion-tracing/tree/main/instrumented-object-store exists and seems to work well with our use of datafusion. Just wanted to point it out to reduce duplication of efforts. cc @geoffreyclaude

👋 Hi!
`instrumented-object-store` uses the `tracing` crate to wrap oject-store operations in spans, and is published as a standalone crate so as not to pull in yet another dependency into `arrow-rs-object-store` or `datafusion`.
If you want just basic debug logging, I'd say this PR is complimentary. All these wrappers are short and simple single file code anyways, so it's probably fine to have multiple different versions!

src/trace.rs
m09526: I've made some changes to use spans instead.

m09526:
How does that version look? I've made some modifications to avoid the use of single events and moved to using spans.

geoffreyclaude:
> I think this looks good to me -- thank you @m09526 
> 
> 
> @geoffreyclaude and @asubiotto  are you happy with this implementation as well?

@alamb I agree with @asubiotto that moving `instrumented-object-store` as is here would probably be simpler, if only because it is already validated in (our) production setup. Then if there are changes to do for better user experience or performance, we can iterate on the code.

Moreover, from a quick look at this PR it seems to be missing functionality of `instrumented-object-store`, e.g. tracing of operation results, and the `get` and `put` calls?

m09526:
The original implementation of this code which was based on the `log` crate has been tested in our code for nearly a year now, so we're confident it works as expected.

> Moreover, from a quick look at this PR it seems to be missing functionality of `instrumented-object-store`, e.g. tracing of operation results, and the `get` and `put` calls?

We deliberately only instrumented the non-provided trait functions since the plain versions of `get` and `put` are just wrappers around those functions.

geoffreyclaude:
> the plain versions of `get` and `put` are just wrappers around those functions.

Not necessarily: you can very well have a particular implementation that overrides the Trait's `get` and `put` defaults without calling `get_opts` or `put_opts`. In which case it won't get traced?

Issue #428, opened 2025-07-08
Release object store  `0.12.3` (non breaking API) Release July 2025


alamb:
**Which part is this question about**

- Part of https://github.com/apache/arrow-rs-object-store/issues/392
- Follow on to https://github.com/apache/arrow-rs-object-store/issues/307

Per [release schedule](https://github.com/apache/arrow-rs-object-store?tab=readme-ov-file#release-schedule) we aim for2 months release: 

**Additional context**
- See prior discussion https://github.com/apache/arrow-rs-object-store/issues/306

**Describe the solution you'd like**
- [x] Update version + make release notes: 
- [x] Test with DataFusion: https://github.com/apache/datafusion/pull/16753
- [ ] Prepare release candidate + begin vote thread: 
- [ ] Release to crates.io
- [ ] File tickets for next planned releases: 

Community requested content: 
- [x] https://github.com/apache/arrow-rs-object-store/pull/410 (https://github.com/apache/arrow-rs-object-store/issues/393#issuecomment-3042312313)
- [x] https://github.com/apache/arrow-rs-object-store/pull/426  (https://github.com/apache/arrow-rs-object-store/issues/367#issuecomment-3046011725)

alamb:
> I've created a [branch `release/0.12`](https://github.com/apache/arrow-rs-object-store/tree/release/0.12) based of the [tag `0.12.2`](https://github.com/apache/arrow-rs-object-store/releases/tag/v0.12.2). I'll file a "backports PR" into that branch with picks the relevant changes from [branch `main`](https://github.com/apache/arrow-rs-object-store/tree/main/). So the backport will look roughly like this:

Sounds like a good plan to me -- once you complete this process, I'll make a release candidate from that branch (and will update the documentation in this repo to reflect that process)

Would you like help creating the backports or do you plan to do it yourself?

crepererum:
I'm on it.

alamb:
I have made a release candidate and started a voting thead: https://lists.apache.org/thread/vtq1bz3wxbkb6y6n162jhqr0159b3hn0

```
---------- Forwarded message ---------
From: Andrew Lamb <alamb@influxdata.com>
Date: Fri, Jul 11, 2025 at 3:40 PM
Subject: [VOTE][RUST] Release Apache Arrow Rust Object Store 0.12.3 RC1
To: dev <dev@arrow.apache.org>


Hi,

I would like to propose a release of Apache Arrow Rust Object
[...]

PR #412, opened 2025-06-19
Prevent compilation error with all cloud features but fs turned on

jder:
# Which issue does this PR close?

Closes #411 .

# Rationale for this change

Allows the crate to compile with `--no-default-features --features=aws,azure,gcp,http`

# What changes are included in this PR?

Adds a missing feature test for including a fallback case in the relevant match statement. I could also add this set of feature flags to CI if desired but not sure it's worth it?

# Are there any user-facing changes?

Just the lack of compilation failure.

PR #407, opened 2025-06-16
refactor: remove AWS dynamo integration

crepererum:
# Which issue does this PR close?
Closes #373.

# Rationale for this change
Clean up obsolete code.

# What changes are included in this PR?
Remove dynamo integration.

# Are there any user-facing changes?
Previously deprecated functionality was removed.

PR #431, opened 2025-07-09
feat (azure): support for '.blob.core.windows.net' in "az://" scheme

vladidobro:
# Which issue does this PR close?

Closes #430

# What changes are included in this PR?

Add support for URLs
- "az://container@account.blob.core.windows.net"
- "az://container@account.blob.fabric.microsoft.com"
- "abfs://container@account.blob.core.windows.net"
- "abfs://container@account.blob.fabric.microsoft.com"
- "abfss://container@account.blob.core.windows.net"
- "abfss://container@account.blob.fabric.microsoft.com"

Added tests for 
- "az://container@account.blob.core.windows.net"
- "az://container@account.blob.fabric.microsoft.com"
The other ones seem unnecessary

# Are there any user-facing changes?

It seems like there is no documentation for the allowed URLs, so there is no docs to update.
The change is non-breaking, only previously unparsable URLs are now parsable

PR #433, opened 2025-07-10
0.12.3 backports

crepererum:
See #428.

# Content
| PR | Included |
| - | - |
| #404 | ✅ |
| #399 | ✅ |
| #407 | ❌ |
| #406 | ✅ |
| #418 | ✅ |
| #417 | ✅ |
| #422 | ✅ |
| #412 | ✅ |
| #410 | ✅ |
| #421 | ✅ |
| #426 | ✅ |
| #429 | ✅ |
| #403 | ✅ |

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -658,8 +658,8 @@ impl MicrosoftAzureBuilder {
         };
 
         match parsed.scheme() {
-            "az" | "adl" | "azure" => self.container_name = Some(validate(host)?),
-            "abfs" | "abfss" => {
+            "adl" | "azure" => self.container_name = Some(validate(host)?),
+            "az" | "abfs" | "abfss" => {
                 // abfs(s) might refer to the fsspec convention abfs://<container>/<path>
                 // or the convention for the hadoop driver abfs[s]://<file_system>@<account_name>.dfs.core.windows.net/<path>
                 if parsed.username().is_empty() {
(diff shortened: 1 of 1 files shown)


paraseba (approved):

src/lib.rs
```diff
@@ -1234,9 +1234,14 @@ impl From<Attributes> for PutOptions {
     }
 }
 
+// See <https://github.com/apache/arrow-rs-object-store/issues/339>.
+#[doc(hidden)]
+#[deprecated(note = "Use PutMultipartOptions", since = "0.13.0")]
```
mbrobbel, line 1239: Maybe we should use a different version here?

mbrobbel (approved):

vladidobro:
Hello!

Please excuse me interrupting the release flow, but do you think it could be possible to fit in this very small MR (not yet merged or approved)? https://github.com/apache/arrow-rs-object-store/pull/431
As far as I understand, it should be in line with object_store goals (sensible compatibility with fsspec URLs).

src/lib.rs
crepererum: Since this needs to be fixed on `main` too, I'm gonna fix that in a release notes PR which needs to land on `main` anyways.

crepererum:
I've manually merged the PR using git because GitHub insisted on squashing the commits -- which would have been a mess.

PR #435, opened 2025-07-10
Fix for URI generation issue #434

### LINKED ISSUE #434: URI builder starts query strings with ampersand
**Describe the bug**
We encountered a bug in DDN Infinia where list requests were failing. We tracked it down to an unusual URI in the format of `http://server/bucket?&list-type=2&prefix=foo/`. Starting a query string with an ampersand (before `list-type`) is unusual and broke our URI parser.

**To Reproduce**
Run a list objects request with the AWS plugin and trace the URI generated.

**Expected behavior**
The URI should not start a query string with an ampersand. The above URI should be:
`http://server/bucket?list-type=2&prefix=foo/`

**Additional context**
We have fixed our parser to handle this corner case, but have customers we cannot upgrade right now so we would like to fix it in the crate.

I have a PR I will submit shortly.

jasoncwik:
When issuing list requests using the AWS module, we noticed that list objects requests were failing due to a strange URI. We saw the generated URI was

`http://server/bucket?&list-type=2&prefix=foo/`

but it should be

`http://server/bucket?list-type=2&prefix=foo/`

i.e. the query string should not start with an ampersand before the list-type parameter.

Closes #434 

# Rationale for this change
Fixes URI query parameter building to be more standard.

One change to URI building.

None.

### DIFF
--- a/src/client/builder.rs
+++ b/src/client/builder.rs
@@ -349,6 +349,33 @@ mod tests {
                 .query(&[("list-type", "2")]),
             "http://example.com/bucket?existing=param&list-type=2",
         );
+
+        // Test multiple parameters
+        assert_request_uri(
+            HttpRequestBuilder::new(client.clone())
+                .uri("http://example.com/bucket")
+                .query(&[("list-type", "2"),
+                    ("prefix", "foo/")]),
+            "http://example.com/bucket?list-type=2&prefix=foo%2F",
+        );
+
+        // Test multiple parameters to existing URI
+        assert_request_uri(
+            HttpRequestBuilder::new(client.clone())
+                .uri("http://example.com/bucket?existing=param")
+                .query(&[("list-type", "2"),
+                    ("prefix", "foo/")]),
+            "http://example.com/bucket?existing=param&list-type=2&prefix=foo%2F",
+        );
+
+        // Test multiple appends
+        assert_request_uri(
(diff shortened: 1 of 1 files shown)


src/client/builder.rs
```diff
@@ -181,9 +181,28 @@ impl HttpRequestBuilder {
     pub(crate) fn query<T: serde::Serialize + ?Sized>(mut self, query: &T) -> Self {
         let mut error = None;
         if let Ok(ref mut req) = self.request {
-            let mut out = format!("{}?", req.uri().path());
-            let start_position = out.len();
-            let mut encoder = form_urlencoded::Serializer::for_suffix(&mut out, start_position);
+            // Build the query string, preserving any existing query parameters.
+            // This prevents malformed URIs like "/bucket?&list-type=2" when the URI
+            // already contains query parameters.
```
tustvold, line 186: Why do we need to handle this case, we didn't before - we solely used the path?

Was this AI generated or something?

flipfloptech: Not sure if it's really a bug in the library. A query string that starts with an ampersand works fine in php($_GET["var"]) and python3 urllib parse_qs. I am also unable to find any RFC or protocol specification that outlines that a query string should not/cannot begin with an ampersand.

tustvold:
Converting to draft as not waiting on review

PR #436, opened 2025-07-11
aws: downgrade credential provider info! log messages to debug!

asubiotto:
Given this is a small change, I did not think it was necessary to open an issue. I can do so if necessary.

# Rationale for this change
These log messages are very noisy and it seems like they can be downgraded to debug.

Just downgrading log messages from info to debug

Users that would previously see `Using * credential provider` log messages with info level will not see them any more.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -36,7 +36,7 @@ use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
-use tracing::info;
+use tracing::debug;
 use url::Url;
 
 /// Default metadata endpoint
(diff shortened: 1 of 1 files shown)


tustvold:
I don't feel strongly about this, although I do think these logs are useful. I would say if you are seeing lots of these logs it might suggest you are not reusing stores, which will cause serious performance problems. You should not create a store per request for example.

asubiotto:
The thing that creates these object stores are cloud run/lambda functions so I don't have much control over that. I also don't feel strongly since I can just turn them off, but the key question is should these log messages be info or debug?

src/aws/builder.rs
```diff
@@ -943,7 +943,7 @@ impl AmazonS3Builder {
         } else if self.access_key_id.is_some() || self.secret_access_key.is_some() {
             match (self.access_key_id, self.secret_access_key, self.token) {
                 (Some(key_id), Some(secret_key), token) => {
-                    info!("Using Static credential provider");
+                    debug!("Using Static credential provider");
```
alamb, line 946: Given these messages are just static strings, I agree that debug is a more appropriate level of logging in my opinion

alamb (approved):

src/aws/builder.rs
philjb: i agree with this.

alamb:
Thanks @asubiotto and @philjb

PR #440, opened 2025-07-14
fix: update links in release docs and script

mbrobbel:

None.

# Rationale for this change
 
Noticed some links were not working in the release mail.

Update links that were not updated when this crate was moved out of the `arrow-rs` repo.

No.

### DIFF
--- a/dev/release/README.md
+++ b/dev/release/README.md
@@ -57,14 +57,11 @@ labels associated with them.
 
 Now prepare a PR to update `CHANGELOG.md` and versions on `main` to reflect the planned release.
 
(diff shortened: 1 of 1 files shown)


dev/release/README.md
```diff
@@ -57,9 +57,9 @@ labels associated with them.
 
 Now prepare a PR to update `CHANGELOG.md` and versions on `main` to reflect the planned release.
 
-Note  this process is done in the `object_store` directory. See [#6227] for an example
+Note this process is done in the `object_store` directory. See [#437] for an example
```
tustvold: I think this line should just be removed, there is no such directory anymore

PR #441, opened 2025-07-19
feat: re-export HTTP types used in public API

### LINKED ISSUE #263: Types from http through request leak into object_store public interfaces but aren't re-exported
The http is meant to be utility types for http libraries to agree, reqwest re-exports the headers module, and these are in the public interface of object_store but AFAICT are not exported by the object_store crate.

This creates an unclear situation for consumers, I wound up importing a library that depends on object_store, they depended on reqwest solely for these types, but didn't do the steps y'all did  (thank you for that btw) to depend on rustls causing me to depend on both toolchains and creating a cross-compilation nightmare.

I'd submit a PR, but I'm not exactly sure how y'all want to handle re-exports– having access to HeaderMap and HeaderValue at the very least would be very helpful to construct ClientOptions.

Thanks!

RE: https://github.com/lancedb/lance/pull/1189 https://github.com/lancedb/lance/pull/1188

ByteBaker:

Closes #263 .

# Rationale for this change
 
User convenience.

Re-export `HeaderMap`, `HeaderValue`, and `Extensions` from http crate to avoid forcing users to add http dependency when using object_store public API.

The types `HeaderMap`, `HeaderValue`, and `Extensions` can now be accessed directly from object-store without adding http as a separate dependency.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -567,6 +567,9 @@ pub use payload::*;
 pub use upload::*;
 pub use util::{coalesce_ranges, collect_bytes, GetRange, OBJECT_STORE_COALESCE_DEFAULT};
 
+// Re-export HTTP types used in public API
+pub use ::http::{Extensions, HeaderMap, HeaderValue};
+
 use crate::path::Path;
 #[cfg(all(feature = "fs", not(target_arch = "wasm32")))]
 use crate::util::maybe_spawn_blocking;
@@ -987,7 +990,7 @@ pub struct GetOptions {
     /// that need to pass context-specific information (like tracing spans) via trait methods.
     ///
     /// These extensions are ignored entirely by backends offered through this crate.
-    pub extensions: ::http::Extensions,
+    pub extensions: Extensions,
 }
 
 impl GetOptions {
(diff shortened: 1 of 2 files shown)


ByteBaker:
Pushed again after resolving CI issues.

src/lib.rs
```diff
@@ -567,6 +567,16 @@ pub use payload::*;
 pub use upload::*;
 pub use util::{coalesce_ranges, collect_bytes, GetRange, OBJECT_STORE_COALESCE_DEFAULT};
 
+/// Type alias for HTTP extensions used in object store operations
```
tustvold: What's the motivation for using a type alias Vs re-exporting the http type?

ByteBaker: To be honest I didn't think of it much. At the moment it felt cleaner rather than to use `::http::Extensions` everywhere. But I guess I could put it next to the other two imports and it won't change a thing.

ByteBaker: Will update the code and push again.

tustvold: I think I would prefer a re-export, as I think it is potentially confusing otherwise - can the types be used interchangeably or should people use object_store::Extensions in case we redefine it?

PR #442, opened 2025-07-21
Dont unwrap on body send

cetra3:
# Rationale for this change

Spawn service was panicking when it was trying to write out the body to the channel.

This PR adjusts the spawn service so it doesn't panic, but instead just finishes off the task.

No

# Side Note

 I think you should really ban the use of `unwrap` in any code in this project (besides tests).  You can do so with clippy pretty easily


tustvold:
Do you have a test / way to reproduce this?

PR #444, opened 2025-07-21
feat: Add HDFS object store implementation

### LINKED ISSUE #424: Integrate HDFS object store

Code base

**Describe your question**

I am the creator of [hdfs-native](https://github.com/Kimahriman/hdfs-native) and the object store implementation that sits on top of it [hdfs-native-object-store](https://github.com/datafusion-contrib/hdfs-native-object-store). I'm wondering if I could just roll the object store implementation directly into this crate.

The object store implementation itself is a fairly small wrapper around the underlying `hdfs-native` library, so there shouldn't be a huge maintenance tail involved, and it would be great to get more eyes on how I have implemented the `object_store` semantics.

There are a few other crates I know of now relying on this object store, and whenever there's a new `object_store` minor release they are also dependent on me cutting a new release of my crate for binary compatibility, and usually there are very minor API changes I just need to make.

Delta is currently the only library that has a dependency on it, but there other non-library projects using it as well, though I don't remember all of them off the top of my head.

https://crates.io/crates/hdfs-native-object-store/reverse_dependencies
https://github.com/lakehq/sail

Additionally polars attempted to add HDFS support but chose not too partially because of the uncertainty of whether the crate would be supported: https://github.com/pola-rs/polars/pull/17506

There are a decent number of additional dependencies it would add, but it shouldn't be a huge issue if it's behind a feature I assume.

**Additional context**

If you are willing to accept it, I can work on a PR (and probably cleaning things up a little bit) as well as figuring out how/what other integration tests I should add. Any users with questions or errors would most likely just need to be forwarded to `hdfs-native` repo for me to investigate.

Kimahriman:

Closes #424 

# Rationale for this change
 

Integrate https://github.com/datafusion-contrib/hdfs-native-object-store, see issue for details.

New object store implementation for HDFS

Just new object store implementaiton.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -200,7 +200,7 @@ jobs:
       - name: Install wasm32-wasip1
         run: rustup target add wasm32-wasip1
       - name: Build wasm32-wasip1
-        run: cargo build fs,aws,azure,gcp,http,cloud,tls-webpki-roots --target wasm32-wasip1
+        run: cargo build --features fs,aws,azure,gcp,http,cloud,tls-webpki-roots --target wasm32-wasip1
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
       - uses: actions/setup-node@v4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,8 @@ jobs:
         run: cargo clippy --features gcp -- -D warnings
       - name: Run clippy with azure feature
         run: cargo clippy --features azure -- -D warnings
+      - name: Run clippy with hdfs feature
+        run: cargo clippy --features hdfs -- -D warnings
       - name: Run clippy with http feature
         run: cargo clippy --features http -- -D warnings
       - name: Run clippy with integration feature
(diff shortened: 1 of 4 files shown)


tustvold:
Thank you for this, however, my feelings on this matter have not materially changed since this was last discussed https://github.com/apache/arrow-rs-object-store/issues/450.

Ultimately we already struggle with maintenance capacity for this project, and HDFS is a significant additional dependency, with very little in common with our current object stores, and that I at least have no capacity to support. Specifically, if people were to raise HDFS issues, I cannot see myself helping triage or resolve them. Therefore I am not really sure what benefit inclusion within this project would really add, other than potentially misleading people about the support status of this integration?

Tagging @alamb @crepererum who may have different perspectives.

Kimahriman:
> @Kimahriman given we haven't really done any breaking releases for a while, have you had to do this recently? Or does this quote reflect your past experience?

Mostly just this most recent 0.12 release that you referred to.

During this process I found some of the helper methods were made public and found that the integration tests were public as well, so that at least helps make it easier to maintain external crates. I was hoping the implementation was simple enough since it's basically a local file system and all the complexity is in my other lib, but I understand the dependency tail is lot since it's not just based on HTTP requests

alamb:
Thank you for understanding @Kimahriman

tustvold:
I guess my concern is it would still represent a non-trivial maintenance burden, webhdfs can only do so much to paper over HDFS, whilst it not being clear that many people would actually use it

PR #445, opened 2025-07-22
Fix not retrying connection errors

### LINKED ISSUE #368: Retry does not cover connection errors

johnnyg:

Closes #368.

# Rationale for this change

As described in #368 connection reset errors aren't being retried due to them being identified as `HttpErrorKind::Unknown` rather than `HttpErrorKind::Interrupted`. This is because the `reqwest::Error` has kind unknown and its immediate source error is a generic `Request` error however if you continue following the error source we eventually get to a `std::io::ErrorKind::ConnectionReset` error.

Currently we stop refining the HTTP error kind once we've been able to downcast the error to either a `hyper::Error` or `std::io::Error` regardless of whether or not we were able to provide a more useful error kind.
This PR changes that so that we keep on refining until we are able to provide a more useful error kind (i.e. not unknown) or we run out of nested errors to check. Because we are only refining while the error kind is unknown, we are not going to accidentally override a more specific error kind with a more generic one.

A test is also included which demonstrates the fix and prevents future regressions.

No.

### DIFF
--- a/src/client/http/connection.rs
+++ b/src/client/http/connection.rs
@@ -101,27 +101,29 @@ impl HttpError {
 
         // Reqwest error variants aren't great, attempt to refine them
         let mut source = e.source();
-        while let Some(e) = source {
-            if let Some(e) = e.downcast_ref::<hyper::Error>() {
-                if e.is_closed() || e.is_incomplete_message() || e.is_body_write_aborted() {
-                    kind = HttpErrorKind::Request;
-                } else if e.is_timeout() {
-                    kind = HttpErrorKind::Timeout;
+        while kind == HttpErrorKind::Unknown {
+            if let Some(e) = source {
+                if let Some(e) = e.downcast_ref::<hyper::Error>() {
+                    if e.is_closed() || e.is_incomplete_message() || e.is_body_write_aborted() {
+                        kind = HttpErrorKind::Request;
+                    } else if e.is_timeout() {
+                        kind = HttpErrorKind::Timeout;
+                    }
                 }
+                if let Some(e) = e.downcast_ref::<std::io::Error>() {
+                    match e.kind() {
+                        std::io::ErrorKind::TimedOut => kind = HttpErrorKind::Timeout,
+                        std::io::ErrorKind::ConnectionAborted
+                        | std::io::ErrorKind::ConnectionReset
+                        | std::io::ErrorKind::BrokenPipe
+                        | std::io::ErrorKind::UnexpectedEof => kind = HttpErrorKind::Interrupted,
+                        _ => {}
+                    }
+                }
+                source = e.source();
+            } else {
                 break;
             }
-            if let Some(e) = e.downcast_ref::<std::io::Error>() {
-                match e.kind() {
-                    std::io::ErrorKind::TimedOut => kind = HttpErrorKind::Timeout,
-                    std::io::ErrorKind::ConnectionAborted
-                    | std::io::ErrorKind::ConnectionReset
-                    | std::io::ErrorKind::BrokenPipe
-                    | std::io::ErrorKind::UnexpectedEof => kind = HttpErrorKind::Interrupted,
-                    _ => {}
-                }
-                break;
-            }
-            source = e.source();
         }
         Self {
             kind,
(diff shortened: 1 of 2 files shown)


mbrobbel:
> @mbrobbel the failing clippy check is due to deprecation warnings in `src/client/backoff.rs`. All the PRs opened this week (#441, #442 and #444) have the same failure.

This was just fixed by #449. You can update with main to fix the warning.

Barre:
I still seem to hit the same issue with this patch.

```task exited with ObjectStoreError(Generic { store: "S3", source: RetryError(RetryErrorImpl { method: PUT, uri: Some(.../01K18PVVKRAMNG8D1K85ASFQNR.sst), retries: 0, max_retries: 10, elapsed: 30.000417786s, retry_timeout: 180s, inner: Http(HttpError { kind: Timeout, source: reqwest::Error { kind: Request, source: TimedOut } }) }) })```

`retries: 0, max_retries: 10`

**Comment by johnnyg** [comment]
@Barre that looks like a different issue; I would recommend creating a separate issue for that.
The issue here was that the request was never being sent properly because the connection would get reset (on the TCP level it would never get back an ACK but instead a RST) however in your case it seems like the PUT request is being sent but is timing out before you get a reply. The error is correctly recognised as a timeout however by default object_store will only retry timeouts for GET, HEAD, OPTIONS and TRACE methods (not PUT).
I imagine changing that default would be more of an ideological question to the maintainers of object_store rather than a difficult or large code change (it should be a one liner to change).

**Comment by Barre** [comment]
> I imagine changing that default would be more of an ideological question to the maintainers of object_store rather than a difficult or large code change (it should be a one liner to change).

I don't think there's ideology there :-) 

In the end, this breaks user apps such as SlateDB because they assume requests will be retried. Perhaps they shouldn't and just implement their own retry policies?

File: src/client/http/connection.rs [hunk 1]
```diff
@@ -101,27 +101,29 @@ impl HttpError {
 
         // Reqwest error variants aren't great, attempt to refine them
         let mut source = e.source();
-        while let Some(e) = source {
-            if let Some(e) = e.downcast_ref::<hyper::Error>() {
-                if e.is_closed() || e.is_incomplete_message() || e.is_body_write_aborted() {
-                    kind = HttpErrorKind::Request;
-                } else if e.is_timeout() {
-                    kind = HttpErrorKind::Timeout;
+        while kind == HttpErrorKind::Unknown {
```
tustvold, line 104: :+1: I think this should mean this only reclassifies errors that were previously Unknown and doesn't potentially reclassify others.

src/client/retry.rs
```diff
@@ -818,4 +825,55 @@ mod tests {
         // Shutdown
         mock.shutdown().await
     }
+
+    #[tokio::test]
+    async fn test_connection_reset_is_retried() {
```
tustvold, line 830: I confirmed without the changes in this PR this test fails with

```
request should eventually succeed: RetryError(RetryErrorImpl { method: GET, uri: Some(http://[::1]:43285/), retries: 0, max_retries: 2, elapsed: 662.178µs, retry_timeout: 1s, inner: Http(HttpError { kind: Unknown, source: reqwest::Error { kind: Request, source: hyper_util::client::legacy::Error(Canceled, hyper::Error(Canceled, hyper::Error(Io, Os { code: 104, kind: ConnectionReset, message: "Connection reset by peer" }))) } }) })
```

PR #456, opened 2025-08-02
Add storage class for aws, gcp, and azure

### LINKED ISSUE #330: Support setting storage class when objects are written

Storage class in an important concept in many commercial object stores. Using storage classes can generate important savings.
 
A uniform way to set storage class on `put` that works on AWS, GCS, etc. It can be ignored if the object store doesn't support storage classes.  

`Attributes` is a fixed list of names that doesn't include storage class. I suppose the HTTP client could be configured to add the extra header, but that breaks the abstraction completely.

matthewmturner:

Closes #330 

# Rationale for this change
 
I would like to set the storage class on objects without a custom HTTP client

1. Add `StorageClass` variant to `Attribute`
2. Add implementations for `aws` and `gcp` clients

I started looking into [Azure blob tiers](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tier?tabs=microsoft-entra-id) but I have never used Azure before and it wasn't immediately obvious to me that it works the same way as AWS and GCP so I did not include anything for that.

A new `Attribute` variant is introduced

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -196,6 +196,10 @@ impl Client {
                         has_content_type = true;
                         builder.header(CONTENT_TYPE, v.as_ref())
                     }
+                    Attribute::StorageClass => {
+                        tracing::warn!("StorageClass attribute not supported on HTTP client as header key is unknown");
+                        builder
+                    }
                     // Ignore metadata attributes
                     Attribute::Metadata(_) => builder,
                 };


tustvold:
I think this may also tie into #116 as a mechanism to change the storage tier of an object

mbrobbel (approved):

alamb:
@tustvold are you happy with merging this?

tustvold (approved):
Looks fine to me, although I haven't tested it

src/attributes.rs
```diff
@@ -45,6 +45,14 @@ pub enum Attribute {
...
     /// See [Cache-Control](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control)
     CacheControl,
+    /// Specifies the storage class of the object.
+    ///
+    /// See [AWS](https://aws.amazon.com/s3/storage-classes/),
+    /// [GCP](https://cloud.google.com/storage/docs/storage-classes), and
+    /// [Azure](https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tier).  
+    /// `StorageClass` is used as the name for this attribute because 2 of the 3 storage providers
+    /// use that name
+    StorageClass,
```
alamb, line 55: I verified that adding a new enum variant is not a breaking API change as this enum is marked `#[non_exhaustive]` ✅

PR #460, opened 2025-08-06
Add support for Android

### LINKED ISSUE #459: Fails to create table on Android
**Describe the bug**
I'm using lancedb and trying to run it on Android. Hard link creation seems to be not allowed?

Here's a log I introduce:
```
failed create hard link path="/data/data/io.openinfer.chat/openinfer.io/vector-dbs/research/research_response.lance/_versions/1.manifest" staging_path="/data/data/io.openinfer.chat/openinfer.io/vector-dbs/research/research_response.lance/_versions/1.manifest#1" source=Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
```

**To Reproduce**
* Create an empty lancedb table

**Expected behavior**
Table creates

**Additional context**
I understand that lancedb isn't this project but the crash is coming from here in src/local.rs when using `PutMode::Create`

vlovich:
Android doesn't support hard_link so provide an alternate path for performing the requested operation.

Closes #459.

# Rationale for this change
 
On Android hard links are not allowed. However, there are alternate ways to express atomicity of the same operations - we use OpenOptions::create_new instead to guarantee the target exists or not before clobbering it.

LocalFilesystem put/copy/copy_if_not_exists alternatives for Android that don't use hard_link.

No.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,9 @@ serde_json = { version = "1.0", default-features = false, features = ["std"], op
 serde_urlencoded = { version = "0.7", optional = true }
 tokio = { version = "1.29.0", features = ["sync", "macros", "rt", "time", "io-util"] }
 
+[target.'cfg(target_os="android")'.dependencies]
+nix = { version = "0.30.0" }
+
 [target.'cfg(target_family="unix")'.dev-dependencies]
 nix = { version = "0.30.0", features = ["fs"] }
 
--- a/src/local.rs
+++ b/src/local.rs
@@ -133,6 +133,37 @@ impl From<Error> for super::Error {
     }
 }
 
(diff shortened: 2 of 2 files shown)


tustvold (commented):
This will briefly create an empty file which could be read by a racing reader, violating atomicity

vlovich:
Found a more elegant solution using renameat2 for put / copy_if_not_exists and for copy we just use std::fs::copy to the staged path to create the temporary (this is efficient because under the hood it's a copy_file_range, not an actual copy).

tustvold:
> Maybe we could use an android emulator, similar to how we test using an s3 emulator

It doesn't even need to be that complicated, the APIs exist on Linux, it just needs a config flag to switch to using them for the purposes of testing.

PR #462, opened 2025-08-11
DRAFT: Pluggable crypto

### LINKED ISSUE #413: Change crypto provider from ring to aws-rust-lc

ring is currently the library used in this project as crypto functions provider. 

I suggest that aws-lc-rs would be a better choice as it's FIPD compliant which could be a better selling point in some industries. The migration from ring to aws-ls-rs can easily be done.

aws-lc-rs is also the default library used by rustls.

Replace ring as explicit dependency of the crate with aws-lc-rs

**Additional context**

If aws-lc-rs is already used in some project, having to compile ring is actually redundant. Both libraries are similar in purpose but aws-lc-rs has a higher standard in terms of validation.

JakeDern:

Closes #413.

# Rationale for this change
 
Many companies, including my current employer, are picky about what cryptographic implementations may be used. This PR seeks to accomplish two things:

1. Make it possible to have no default crypto provider, such as ring, in the dependency tree
2. Make it possible to bring your own crypto provider

This is heavily WIP - I'm just opening this to make it clear how the approach with generic types would work vs an implementation that would use `dyn` instead at the expense of having to allocate and copy the outputs of various crypto implementations.

- Add a `CryptoProvider` trait
- Add an optional `RingProvider` implementation based on the `ring` feature

Yes, several including:

- Making builders for clouds that need crypto generic
- Making `default::Default` implementations dependent on the `ring` feature 
- Making `new()` require a crypto provider

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -150,6 +150,13 @@ jobs:
       - name: Run object_store tests
         run: cargo test --features=aws,azure,gcp,http
 
+      - name: Run crypto feature tests
+        run: |
+          # With ring
+          cargo test crypto --no-default-features --features=aws,azure,gcp,ring
+          # Without ring
+          cargo test crypto --no-default-features --features=aws,azure,gcp
+
       # Don't rerun doc tests (some of them rely on features other than aws)
       - name: Run object_store tests (AWS native conditional put)
         run: cargo test --lib --tests --features=aws


src/crypto.rs
```diff
@@ -0,0 +1,97 @@
+use std::fmt::{Debug, Display};
+
+type DynError = Box<dyn std::error::Error + Send + Sync>;
+
+#[derive(Debug)]
+pub struct Error {
```
tustvold: I think it'd be better to just use object_store::Error, whilst it contains variants that aren't relevant, I don't think this matters

src/crypto.rs
```diff
@@ -0,0 +1,97 @@
...
+    fn from(err: DynError) -> Self {
+        Error { inner: err }
+    }
+}
+
+/// TODO(jakedern): Docs
+pub trait CryptoProvider: Send + Sync + Debug + 'static {
+    fn digest_sha256(bytes: &[u8]) -> Result<impl AsRef<[u8]>, Error>;
+    fn hmac_sha256(secret: &[u8], bytes: &[u8]) -> Result<impl AsRef<[u8]>, Error>;
+    fn hex_digest(bytes: &[u8]) -> Result<String, Error> {
```
tustvold: Probably don't need this to be here unless you expect people to override it

src/crypto.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    impl Default for RingProvider {
+        fn default() -> Self {
+            RingProvider
+        }
+    }
+}
+
+pub mod openssl_crypto {
```
tustvold, line 143: I would personally prefer people to add additional providers out of tree if they need them. I'm not massively excited about adding a dependency on openssl, optional or not.

tustvold (commented):
Took a quick look, some comments

* I think Object safety is pretty important, given the computed types are fixed width it should be possible to use `[u8; 32]` or similar to avoid allocations (although given the operations involved an allocation is likely irrelevant)
* We should avoid changing the signature of `new` as this will break lots of code, instead we could add a new constructor, e.g. new_with_crypto or something

src/crypto.rs
JakeDern: Yep, will take this out - Was just trying out a couple of providers to see how they compare and if the trait was flexible enough

JakeDern:
@tustvold I have a pretty complete implementation here. Other than more docs and bits of cleanup along those lines, the last major bit I haven't accounted for is the [Rsa key stuff](https://github.com/apache/arrow-rs-object-store/blob/94c25d2dea15d2a7154bb166ae58cbf9452ebcd9/src/gcp/credential.rs#L125C1-L126C1) for GCP credentials. 

This is (today) specific to GCP and appears at first glance to be quite a bit more complicated to put into an object safe trait due to the requirement to create an intermediate `RsaKeyPair` equivalent type that you can then re-use.

I'm assuming we'd want to also encapsulate that into the trait, but I just want to raise it for discussion. Maybe there's a straightforward way to do this that I don't know, I only have a few months experience with rust. Usually I'd have something in the interface that returns `impl MyRsaKeyPairTrait` that I could hold the library specific state in, but of course that won't be object safe anymore. 

Let me know what your thoughts are and thanks again for working with me on this! I know it's a lot of mental overhead, so I appreciate it!

JakeDern:
@tustvold friendly ping on [the above](https://github.com/apache/arrow-rs-object-store/pull/462#issuecomment-3186411496) - Please let me know if you have thoughts about whether the GCP certificate credential stuff should be included in this trait and if you have any ideas on a reasonable object safe way to do so. Thanks again for your time!

Issue #464, opened 2025-08-12
Allow flagging `PUT` operations as idempotent.


calavera:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Put operations in object store are usually idempotent. However, that guarantee does not exist since this crate calls APIs that it does not control.

Some clients might want to take the risk of ensuring all `put` operations are idempotent, and retried when unexpected errors happen, like SlateDB.

I'm bringing the discussion of allow flagging operations as idempotent here, since I believe it'd be better solved in this repository.

**Describe the solution you'd like**

I'd like to add a flag to `PutOptions` that people can use to make `PUT` operations idempotent.

**Describe alternatives you've considered**

The only alternative is to force consumers of this library to implement their own retry logic. Even if this is doable, it adds repetition and additional complexity to people that want to integrate this crate in their applications. The SDKs and HTTP libraries that call the cloud providers APIs already implement retry logic that we can leverage.

**Additional context**

See https://github.com/slatedb/slatedb/issues/766 for context.

PR #465, opened 2025-08-12
Optional put idempotency.

calavera:

Closes https://github.com/apache/arrow-rs-object-store/issues/464.

# Rationale for this change

Put operations in object store are usually idempotent. However, that guarantee does not exist since this crate calls APIs that it does not control.

Some clients might want to take the risk of ensuring all `put` operations are idempotent, and retried when unexpected errors happen, like SlateDB.
 

This change introduces a new flag in `PutOptions` to let callers mark the `put` operation as idempotent, triggering the retry logic that the underlying HTTP client implements.

Yes, there is a new flag in `PutOptions` that users can enable. I've included rustdoc changes with the new flag.

### DIFF
--- a/src/aws/mod.rs
+++ b/src/aws/mod.rs
@@ -167,6 +167,7 @@ impl ObjectStore for AmazonS3 {
             tags,
             attributes,
             extensions,
+            idempotent,
         } = opts;
 
         let request = self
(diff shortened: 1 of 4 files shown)


tustvold:
I think if anything this should be a config option of the store, not a PutOption, as you're then making assertions about the semantics of a store at a level where the underlying store has potentially been type erased.

That being said I'm not really sure how this would actually help in practice, the types of error that idempotency gates retries for are where the outcome is not known - e.g. a timeout occured. If you blindly retry say a PutMode::Create you may now get an AlreadyExists error.

Whilst this is technically idempotent, as idempotency refers to the server side state, you will still need retry/error handling machinery at the higher level to account for this and handle this. The fact PutMode::Create is being used is implicitly stating that blindly retrying is not the desired outcome, otherwise you could just use PutMode::Overwrite (which will retry such errors)

tustvold:
Closing as I believe this has been solved another way

Issue #489, opened 2025-09-17
Release object store  `0.12.4` (non breaking API) Release September 2025


alamb:
**Which part is this question about**

- Part of https://github.com/apache/arrow-rs-object-store/issues/392
- Follow on to https://github.com/apache/arrow-rs-object-store/issues/428

Per [release schedule](https://github.com/apache/arrow-rs-object-store?tab=readme-ov-file#release-schedule) we aim for2 months release: 

**Additional context**

**Describe the solution you'd like**
- [x] Update version + make release notes: https://github.com/apache/arrow-rs-object-store/pull/491
- [x] Test with DataFusion: https://github.com/apache/datafusion/pull/17671
- [x] Test with other crates: https://github.com/developmentseed/obstore/pull/568
- [x] Prepare release candidate + begin vote thread: https://lists.apache.org/thread/bzdxosoms6zqj1ss0v7wq3s5l40w3fjj
- [ ] Release to crates.io
- [ ] File tickets for next planned releases:

alamb:
FYI @crepererum @mbrobbel and @tustvold  -- I will hope to make a release candidate later this week

alamb:
Here are the PRs that we are waiting on (everything should be ready to go shortly)
- https://github.com/apache/arrow-rs-object-store/pull/480 
- https://github.com/apache/arrow-rs-object-store/pull/492
- https://github.com/apache/arrow-rs-object-store/pull/491

alamb:
If anyone else has time to test their crates with this release that would be super helpful

Here is an example of how I tested with DataFusion:
- https://github.com/apache/datafusion/pull/17671

alamb:
Given the results from @kylebarron and myself in DataFusion I will proceed to making a RC and starting a vote. If anyone else tests and finds anything (either good or bad) over the voting period (3 days) please let us know

evanj:
In case it helps: I was able to build/pass unit tests for the thing I work on where I asked about this release. Confirming that this fixes the retry issue that we see occasionally when talking to real object storage will take a bit longer though, since we will need to deploy in then wait for the connection resets to happen.

Thanks for preparing this release!

alamb:
Thanks @evanj  🙏

Issue #373, opened 2025-05-17
Deprecate and Remove DynamoCommit


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

DynamoCommit was added to allow applications to be developed requiring conditional operations despite S3 being the only store not to natively support them. Since then AWS has added native support for conditional operations #54, we can therefore remove the DynamoCommit functionality.

**Describe the solution you'd like**

I would like to deprecate the DynamoCommit functionality and eventually remove it

**Describe alternatives you've considered**

**Additional context**

alamb:
I reverted  the original commit to make a 0.12.4 release:
- https://github.com/apache/arrow-rs-object-store/pull/493

PR #493, opened 2025-09-19
Revert "refactor: remove AWS dynamo integration (#407)"

alamb:

- part of https://github.com/apache/arrow-rs-object-store/issues/489

# Rationale for this change

In order to release a 0.12.4, we can't have breaking API changes, thus we need to revert https://github.com/apache/arrow-rs-object-store/pull/407 if we would like to make that releqase

- This reverts commit 034733f5aa760529a647041468681328d2f222c7 / https://github.com/apache/arrow-rs-object-store/pull/407

Old API is restored

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@
 # under the License.
 
 ---
+
 name: CI
 
 concurrency:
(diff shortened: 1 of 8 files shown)


alamb:
- If we merge this PR, we should also reopen https://github.com/apache/arrow-rs-object-store/issues/373

PR #500, opened 2025-10-06
Allow more settings without vendor prefix

jonashaag:

Closes #.

# Rationale for this change

Similar to https://github.com/apache/arrow-rs-object-store/pull/467, also allow `server_side_encryption` instead of `aws_server_side_encryption` and some other `aws_` prefixed settings.
 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -195,7 +195,7 @@ jobs:
         run: cargo build --all-features --target wasm32-wasip1
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
-      - uses: actions/setup-node@v5
+      - uses: actions/setup-node@v6
         with:
           node-version: 20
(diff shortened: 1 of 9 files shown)


kylebarron (approved):
I think this is good to align with Google and Azure config keys, where every key supports both the vendor prefix like `aws_bucket` as well as the bare version `bucket`

https://github.com/apache/arrow-rs-object-store/blob/4102e4d28ceef960aaacbbdf5b1ef37fdb697a25/src/gcp/builder.rs#L191-L201

https://github.com/apache/arrow-rs-object-store/blob/4102e4d28ceef960aaacbbdf5b1ef37fdb697a25/src/azure/builder.rs#L421-L470

alamb:
I restarted the failed MSRV check (some sort of network error)

kylebarron:
I don't think there are existing tests for the string config parsing, but maybe I missed them.

I'm not sure what kind of test would make sense here, because this feels like a situation where the test just copies 1:1 the implementation of the enum creation

PR #503, opened 2025-10-13
gcp: implement multipart uploads using resumable upload API

### LINKED ISSUE #504: implement GCS resumable uploads as a multipart upload method

When performing a streaming upload to GCS, the only option in this crate is [Multipart Uploads](https://cloud.google.com/storage/docs/multipart-uploads). This works the same way it does for S3, and has the same limitations: you cannot perform an upload with >10000 parts. This presents a challenge for large uploads with unknown size. You have to pick a part size large enough to allow the largest plausible upload size, but still small enough that your buffers don't grow too large, and your cost of retrying a part upload is not too great.

The [Resumable upload](https://cloud.google.com/storage/docs/resumable-uploads) API does not have this limitation, you can upload as many parts as you want at whatever size you want. 

The downside of resumable uploads is that part uploads must be done serially, you cannot upload parts in parallel or out-of-order.

I'd like the option to use resumable uploads when using `put_multipart`. I would configure this in the client builder.

I can't think of one.

**Additional context**

james-rms:

Closes #504 

- [x] Implement resumable uploads for GCP
- [ ] Add GCPClientBuilder option to use resumable uploads rather than multipart uploads

Yes, see changes to the client builder.

### DIFF
--- a/src/gcp/mod.rs
+++ b/src/gcp/mod.rs
@@ -34,6 +34,7 @@
 //! enabled by setting [crate::ClientConfigKey::Http1Only] to false.
 //!
 //! [lifecycle rule]: https://cloud.google.com/storage/docs/lifecycle#abort-mpu
+use std::sync::atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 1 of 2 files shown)


src/gcp/mod.rs
```diff
@@ -114,6 +115,127 @@ struct UploadState {
...
+    client: Arc<GoogleCloudStorageClient>,
+    session_uri: String,
+    in_flight: AtomicUsize,
+    next_offset: AtomicU64,
+    closed: AtomicBool,
+}
+
+#[derive(Debug)]
+struct GCSResumableUpload {
+    state: Arc<ResumableState>,
```
tustvold, line 129: Why do we need to use Arc and atomics here, AFAICT there isn't an obvious reason for this?

Edit: Oh I see, it is to try to prevent concurrent put part... NGL kind of feels like we're trying to shoehorn something into an interface that doesn't really fit...

james-rms: That's a fair judgement.

tustvold: I wonder if we could use https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html to serialise the uploads, I think this would be more in keeping with the design than erroring if polled concurrently

james-rms: We could, but i'm not sure if it's in keeping with the design enough to not break something. I was worried about applications that try to do parallel upload and being surprised that most of their tasks are stalled.

Issue #510, opened 2025-10-16
Signature mismatch (sigv4) when using attribute values with `double whitespace`


ion-elgreco:
**Describe the bug**
I am setting an attribute such as `attributes: [("foo", "hello  world")].into()` on my object which has a double whitespace. When I put the object with object-store I get an authorization mismatch because the signature that object-store creates doesn't match with LakeFS signature creation. If however set the object with the value `hello world` with a single whitespace the object is successfully put on the store.

cc lakefs dev @arielshaqed 

I guess the question is where is the issue, is it the client signature or server signature creation.

PR #512, opened 2025-10-17
fix: canonicalize multiple whitespaces in SigV4

ion-elgreco:
- closes https://github.com/apache/arrow-rs-object-store/issues/510

I've extended `src/integration.rs` to have a metadata attributeValue with multiple whitespaces, I am not sure if this is valid for other stores though :)

### DIFF
--- a/src/aws/credential.rs
+++ b/src/aws/credential.rs
@@ -414,7 +414,7 @@ fn canonicalize_query(url: &Url) -> String {
     encoded
 }
 
-fn append_normalized_whitespace_value<'a>(headers: &'a mut String, input: &str) {
+fn append_normalized_whitespace_value(headers: &'_ mut String, input: &str) {
     let mut iter = input.split_whitespace();
 
     if let Some(first) = iter.next() {


src/client/backoff.rs
```diff
@@ -110,7 +110,7 @@ impl Backoff {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use rand::rand_core::impls::fill_bytes_via_next;
+    use rand_core::impls::fill_bytes_via_next;
```
tustvold: Why this change?

Cargo.toml
```diff
@@ -54,6 +54,7 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.10.6", default-features = false, optional = true }
 quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
+rand_core = "0.9.3"  # ensure this matches rand's requirement
```
tustvold: This change seems unrelated (and brittle)?

src/aws/credential.rs
```diff
@@ -423,6 +423,10 @@ fn canonicalize_headers(header_map: &HeaderMap) -> (String, String) {
     let mut value_bytes = 0;
     let mut key_bytes = 0;
 
+    fn normalize_whitespace(s: &str) -> String {
+        s.split_whitespace().collect::<Vec<_>>().join(" ")
```
tustvold: The specification only says to convert consecutive spaces, not arbitrary whitespace (although in practice unicode characters or control characters in headers can cause interesting things to happen).

src/aws/credential.rs
```diff
@@ -452,7 +456,7 @@ fn canonicalize_headers(header_map: &HeaderMap) -> (String, String) {
             if value_idx != 0 {
                 canonical_headers.push(',');
             }
-            canonical_headers.push_str(value.trim());
+            canonical_headers.push_str(&normalize_whitespace(value));
```
tustvold: It seems rather wasteful to allocate both a Vec and a String only to append there contents to another string.

Perhaps we could just do something like (not tested)

```
let mut iter = value.trim().split(' ').filter(|x| !x.is_empty());
if let Some(first) = iter.next() {
    canonical_headers.push(first);
    for x in iter {
       canonical_headers.push(' ');
       canonical_headers.push(x);
    }
}
```

src/client/backoff.rs
ion-elgreco: It wouldn't compile for me since it said rand_core wasn't available in rand

tustvold: Maybe some lock file nonsense?

ion-elgreco: Likely yeah, I'll look at it tomorrow and your other comment!

ion-elgreco: Yup cargo updating was enough 😛

src/aws/credential.rs
ion-elgreco: Well split_whitespace does that as well, since if it's a single whitespace, joining it later it will have a single whitespace again.

Looking at the Minio implementation in Go, they replace the unicode characters so not just a whitespace:

https://github.com/minio/minio/blob/05e569960ac584c8927a9af76755708d20f16129/cmd/signature-v4-utils_test.go#L307-L324

src/aws/credential.rs
ion-elgreco: @tustvold I've updated the code to push the strs directly in `&mut headers` and not allocate the Vec, hope thats fine?

tustvold:
Msrv failure is unrelated to this PR

Issue #51, opened 2024-08-28
Better Document Backoff Algorithm


westonpace:
**Describe the bug**

In `object_store::BackoffConfig` we have:

```
Exponential backoff with jitter

See https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/
```

However, the actual implementation is "decorrelated jitter" (which is not exponential).

**To Reproduce**
N/A docs bug

**Expected behavior**
Docs describe implementation

**Additional context**
N/A

PR #513, opened 2025-10-18
fix: correct logic error in backoff behavior

rtyler:
This error in the logic of the backoff configuration meant that in
practice when an object store (e.g. S3) would produce an error, the
steps of the backoff _may_ be incredibly small and not exponentially
backoff properly.

Before this change, a few runs of the test case I added had backoff
steps far exceeding 20 steps for the default configuration, e.g.:

```
  left: 20
 right: 70

  left: 20
 right: 194

  left: 20
 right: 31

  left: 20
 right: 407

  left: 20
 right: 407
```

Due to these steps, a rate-limiting or other error by an object store
would result in an _explosion_ in repeated requests to the object store.
In the case of a rate-limiting situation this further exacerbated the
failure mode.

A rather simple fix however. I did also remove some logic from the
existing test because I do not believe repeating the logic of the
`next()` function in the test is useful from a testing perspective.

### DIFF
--- a/src/client/backoff.rs
+++ b/src/client/backoff.rs
@@ -17,6 +17,7 @@
 
 use rand::{prelude::*, rng};
 use std::time::Duration;
+use tracing::log::trace;
 
 /// Exponential backoff with decorrelated jitter algorithm
 ///
@@ -95,7 +96,8 @@ impl Backoff {
 
     /// Returns the next backoff duration to wait for
     pub(crate) fn next(&mut self) -> Duration {
-        let range = self.init_backoff..(self.next_backoff_secs * self.base);
+        let range = self.next_backoff_secs..(self.next_backoff_secs * self.base);
+        trace!("Backoff range sliding to {range:?}");
 
         let rand_backoff = match self.rng.as_mut() {
             Some(rng) => rng.random_range(range),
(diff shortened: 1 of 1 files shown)


rtyler:
The graph [linked](https://github.com/apache/arrow-rs-object-store/issues/51#issuecomment-2741692035) by @westonpace reflects what we have been seeing in real-world scenarios.

I think the default settings for backoff are non-functional. With the AmazonS3 object store we regularly see operations fail with only about 1s of "total retries" since each backoff step is  roughly 100ms. In a lot of ways this is not fundamentally different than just a `loop {}` on the operation.

The potential retry loop executing over 400 times before the max retries is exhausted I don't think is reasonable for any real-world usage.

This behavior is particularly problematic when the retries are occurring due to a rate-limiting situation, `object_store` basically makes matters _dramatically_ worse with its current retry and backoff implementation

tustvold:
The default retry configuration should only retry 10 times? https://docs.rs/object_store/latest/src/object_store/client/retry.rs.html#248

Also with regards to rate limiting, full jitter is purportedly the optimal way to recover from such a scenario, as it better handles bursty load. That being said perhaps the base delay could be lengthened, but that'd worsen the recovery from 500 errors....

I think there's definitely a discussion about introducing new backoff behaviours, and allowing people to choose the one most applicable for their scenarios - I suspect there is no perfect solution

rtyler:
How about a forward compatible `exponential: Option<bool>` flag in `BackoffConfig`, then I can gate the original algorithm change in this pull request behind that

tustvold:
The bummer with that is BackoffConfig and RetryConfig are not marked as non_exhaustive, and so that would have to wait for the next major release... Really they should have been opaque structs but what's done is done...

One option might be to introduce a new `RetryStrategy` struct that is opaque and deprecate `RetryConfig` and `BackoffConfig` :thinking:

ion-elgreco:
> One option might be to introduce a new `RetryStrategy` struct that is opaque and deprecate `RetryConfig` and `BackoffConfig` 🤔

Can't we add non_exhaustive to the existing structs?

PR #514, opened 2025-10-20
Introduce a platform agnostic sleep future

### LINKED ISSUE #1: [object_store] Runtime Agnostic Retry Machinery

Whilst the HttpClient abstraction allows spawning work to a separate tokio runtime, see apache/arrow-rs#7253, the retry machinery itself also makes use of tokio - see [here](https://github.com/apache/arrow-rs/blob/main/object_store/src/client/retry.rs#L361).

This forces the calling context to be a tokio content.

I think we should make it possible to use object_store in a non-tokio runtime.

We could simply disable backoff and rety if not called from a tokio runtime.

**Additional context**

AdamGS:

Closes #1.

# Rationale for this change
 
Step closer to removing tokio as a runtime dependency.

New `Sleep` future based on `futures-timer` crate. It seems like a reputable crate that's used in many popular projects (like `rstest` and `governer`).

I've also bumped the `rand` dependency from `0.9` to `0.9.1` as there are tests that depend on a re-export of `rand_core` that was introduced on `0.9.1`.

No. Might some async scheduling behavior subtly when running with tokio.

### DIFF
--- a/src/client/sleep.rs
+++ b/src/client/sleep.rs
@@ -1,3 +1,20 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
+//   http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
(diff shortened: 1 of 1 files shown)


crepererum:
So instead of relying on the respective runtime e.g. tokio (which as far as I can tell is [not exactly a simple mechanism](https://github.com/tokio-rs/tokio/tree/master/tokio/src/runtime/time).), we are now using [a thread](https://github.com/async-rs/futures-timer/blob/07dbe53da1b5ddb703cdfc74ad59e15cb4134ce0/src/native/global.rs#L26-L28)? Mmh, it's hard to say if that works under high load as intended.

Given how deeply `tokio` is embedded into the HTTP stack, I wonder if this whole endeavor is actually gonna lead anywhere or if that's mostly a dead end? (serious question, if you think it's worthwhile and tractable, say so)

AdamGS:
Personally, I wish there was some standardized way to abstract over the common parts of the runtime. Tokio is great and there are many good reasons to use it but it can be quite limiting when dependencies effectively force everything around them to use it.
I haven't read through every use of Tokio in `object_store`, but I can imagine it being controlled by some feature flag, allowing to opt into a runtime-agnostic version

PR #515, opened 2025-10-21
Bump rust edition to 2024, rust version to 1.85

AdamGS:

Unlocks https://github.com/apache/arrow-rs-object-store/pull/514.

# Rationale for this change
 
90% of downloads use 1.85, and 2024 unlocks future msrv bumps.

Bumping rust edition and msrv.

Increasing the msrv might have an impact on some dependencies and users, but I assume the vast majority of them will be fine with upgrading (For example - DataFusion is at 1.86 and arrow-rs is at 1.85).

### DIFF
--- a/src/aws/credential.rs
+++ b/src/aws/credential.rs
@@ -1332,7 +1332,10 @@ mod tests {
             ("   a   b  c   ", "a b c"),
             ("a \t b  c   ", "a b c"),
             ("\"a \t b  c   ", "\"a b c"),
-            (" \t\n\u{000b}\r\u{000c}a \t\n\u{000b}\r\u{000c} b \t\n\u{000b}\r\u{000c} c \t\n\u{000b}\r\u{000c}", "a b c"),
+            (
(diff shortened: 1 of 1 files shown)


Xuanwo:
Arrow-rs is bumped to edition 2024, and 1.85 is now used by 90% of downloads based on stats from https://lib.rs/stats. I believe it's fair to bump to 2024 now. Another advantage of edition 2024 is that raising the MSRV is no longer a breaking change, making this a one shot solution.

AdamGS:
good points, I'll have that later today/tomorrow

alamb (commented):
Since this increases the MSRV woudl that make it a breaking change (aka we can only release this as part of a major release)?

alamb:
> IIRC MSRV updates are considered to be "non-breaking", though I can't recall a canonical source for that.

Here is one source:
- https://github.com/rust-lang/api-guidelines/discussions/231

Though it has a lot of discussion which I didn't read

mbrobbel (approved):

alamb:
Sounds great -- I will plan on making a new release in Nov
- https://github.com/apache/arrow-rs-object-store/issues/367
Let's get this code moving

PR #517, opened 2025-10-23
feat: refactor GetOptions with builder, add binary examples

### LINKED ISSUE #516: Refactor GetOptions with a builder pattern

In service of https://github.com/apache/datafusion/issues/18211

GetOptions should be refactored to support a builder API pattern.

Leaving it as is. The current DX requires building the strict manually with defaults or mutating an object yourself.

**Additional context**

peasee:

Closes #516.

# Rationale for this change
 

* Adding `get_range_opts` and `get_ranges_opts` makes DX better when retrieving ranges from pre-existing options (like against specific `e_tag` or `version`)

* Adds some additional examples and documentation for using `get_opts` to retrieve ranges of bytes against versioned files to make it easier for developers to use this functionality

* Refactors `GetOptions` to support a builder pattern for creating the options
* Adds some examples and more documentation about using `get_opts` with range and etag/version

* There are user-facing changes for the introduction of the new public builder functions on `GetOptions`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
         with:
           node-version: 20
       - name: Run wasm32-unknown-unknown tests (via Node)
-        run: wasm-pack test --node --features http --no-default-features --lib --bins --tests
+        run: wasm-pack test --node --features http --no-default-features
 
   windows:
     name: cargo test LocalFileSystem (win64)


src/lib.rs
```diff
@@ -1035,6 +1052,84 @@ impl GetOptions {
         }
         Ok(())
     }
+
+    /// Create a new [`GetOptions`]
```
alamb, line 1233: these are great improvements

src/lib.rs
```diff
@@ -646,10 +646,19 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
+    }
+
+    /// Return the bytes that are stored at the specified location
+    /// in the given byte range with options.
+    /// 
+    /// The `options` provided will be augmented with the specified `range`.
+    /// Any existing `range` in `options` will be overridden.
+    ///
+    /// See [`GetRange::Bounded`] for more details on how `range` gets interpreted
+    async fn get_range_opts(&self, location: &Path, range: Range<u64>, options: GetOptions) -> Result<Bytes> {
```
alamb: Why we would add a new method when you can already pass in a `range` on the `GetOptions`? I think this will actually make the developer experience worse (as now a developer has to figure out which one of the methods to call)

If we want to make DX better I suggest:
1. Adding more examples to `get_opts` and `get_range` that shows how to call them with options / a range
2. Eventually deprecate / remove `get_range` entirely (in favor of just `get_opts`

peasee: I think that is fair. My main point of improvement was `get_ranges`, because it hides away the call over the list of ranges you provide. If you want to get a variety of ranges against a single set of options, that would simplify it for you.

I read the linked issue for cleaning up `ObjectStore` though, and I agree these extra `_opts` functions can be removed.

examples/get.rs
```diff
@@ -0,0 +1,34 @@
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use object_store::{local::LocalFileSystem, path::Path, ObjectStore};
+#[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
+use tempfile::tempdir;
+
+fn main() {
```
alamb: Thank you @peasee  -- this is great. 

I am not sure how discoverable these examples will be (people have to know to go into the code and check them out) How about putting them in the doc comments

For example, we could put them directly on https://docs.rs/object_store/latest/object_store/trait.ObjectStore.html#method.get

src/buffered.rs
```diff
@@ -590,13 +590,7 @@ mod tests {
...
         writer.shutdown().await.unwrap();
         let response = store
-            .get_opts(
-                &path,
-                GetOptions {
-                    head: true,
-                    ..Default::default()
-                },
-            )
+            .get_opts(&path, GetOptions::new().with_head(true))
```
alamb, line 593: ❤️

src/lib.rs
```diff
@@ -324,6 +324,32 @@
 //! # }
 //! ```
 //!
+//! To retrieve ranges from a versioned object, use [`ObjectStore::get_opts`] by specifying the range in the [`GetOptions`].
+//!
+//! ```ignore-wasm32
```
alamb, line 329: I am not familiar with this `ignore-wasm32` annotation -- it seems like it may just not run these examples at all

If we are having trouble running the doc tests on wasm32, I suggest we just don't run the doc examples on wasm32 in CI

src/lib.rs
```diff
@@ -1035,6 +1054,84 @@ impl GetOptions {
...
+    /// Create a new [`GetOptions`]
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Sets the `if_match` condition.
+    ///
+    /// See [`GetOptions::if_match`]
+    #[must_use]
```
alamb, line 1241: Why the `must_use` annotation?

src/lib.rs
```diff
@@ -1035,6 +1054,84 @@ impl GetOptions {
...
+    pub fn with_if_match(mut self, etag: impl Into<String>) -> Self {
```
alamb: With this API there is no way to unset the `if_match` field (set it back to `None`)

I think it is more typical in builders to pass in the same type, something like

```suggestion
    pub fn with_if_match(mut self, etag: Option<impl Into<String>>) -> Self {
```

That would also reduce the number of API functions required by a factor of 2x

alamb (commented):
Thank you @peasee  -- this is looking great. I had a few comments but I think it is really close

src/lib.rs
peasee: It looks like only a few doc examples are excluded on wasm32. A number of them build (but don't necessarily run) like the [conditional fetch example](https://github.com/apache/arrow-rs-object-store/blob/main/src/lib.rs#L376). With the docs target disabled, only a single wasm test [`basic_wasm_get`](https://github.com/apache/arrow-rs-object-store/blob/main/tests/http.rs#L50) runs so I think we should retain the `ignore-wasm32` and keep building the other docs targets.

src/lib.rs
peasee: I tend to use `-D clippy::pedantic` in all my workspaces. When pedantic is enabled, it throws a lint for candidates that could have a must use. I think it's accurate, because it would seem weird you build options but never use them.

src/lib.rs
peasee: I'm not sure I agree. Most of the builders I have seen either don't provide a way to supply an option and `with_x` has the assumption you're setting a value, or they provide a secondary `set_x` to set the specific `Option<T>`. For example:

* [`opentelemetry_sdk::metrics::MeterProviderBuilder`](https://docs.rs/opentelemetry_sdk/0.31.0/src/opentelemetry_sdk/metrics/meter_provider.rs.html#250) provides a `with_resource(mut self, resource: Resource)` where the underlying property is a `Option<Resource>`. It provides no alternative to set the option back to `None`
* [`std::sys::unix::process::Command`](https://stdrs.dev/nightly/x86_64-unknown-linux-gnu/src/std/sys/unix/process/process_common.rs.html#248) stores a `cwd: Option<CString>` but only provides `cwd(&mut self, dir: &OsStr)` with no alternative to set the option back to `None`

I'm not sure setting back to `None` is a common scenario, you should just build a new option?

peasee: Actually, I thought about this some more and I think setting as an option does make more sense mainly because if you're setting an etag from a prior request (like the use case for https://github.com/apache/datafusion/issues/18211, or use cases that list files and then retrieve them) the data type in `ObjectMeta` is already an option. The chain becomes simpler then, because you don't have to unwrap the option and can just clone it directly into the builder.

alamb: Yes -- exactly -- 👍

src/lib.rs
```diff
@@ -634,22 +657,196 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     ) -> Result<Box<dyn MultipartUpload>>;
 
     /// Return the bytes that are stored at the specified location.
+    ///
+    /// ## Example
+    ///
+    /// This example uses a basic local filesystem object store to get an object.
+    ///
+    /// ```ignore-wasm32
```
alamb, line 665: I was worried that tests marked with `ignore-wasm32` aren't actually tested in rust tests
However, I made a deliberate mistake
```diff
diff --git a/src/lib.rs b/src/lib.rs
index e37fb69..0a2b815 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -701,6 +701,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// # use tempfile::tempdir;
     /// # use object_store::{path::Path, ObjectStore, GetOptions};
     /// async fn get_opts_example() {
+    /// XXX
     ///     let tmp = tempdir().unwrap();
     ///     let store = LocalFileSystem::new_with_prefix(tmp.path()).unwrap();
     ///     let location = Path::from("example.txt");
```

And then ran `cargo test`

And it failed as expected 👍 

```
---- src/lib.rs - ObjectStore::get_opts (line 699) stdout ----
error: expected `;`, found keyword `let`
  --> src/lib.rs:705:4
   |
 9 | XXX
   |    ^ help: add `;` here
10 |     let tmp = tempdir().unwrap();
   |     --- unexpected token
```

PR #524, opened 2025-10-30
feat!: use `'static` lifetime for `delete_stream`

### LINKED ISSUE #518: Use `'static` lifetime for `delete_stream`

Right now the `ObjectStore::delete_stream` method requires the input stream to have a lifetime `'a`. This restriction makes it more challenging for implementers. For example, in my project, I have a `ObjectStore` [wrapper](https://github.com/lakehq/sail/blob/5b9f0e500053c036ae3b784bedb15e9ec5eb57d9/crates/sail-object-store/src/layers/runtime.rs#L182-L185) that needs to spawn tasks for the operation, but since the input is `'a`, I cannot move the input to the future due to Rust's lifetime checker.

It seems `delete_stream` is the only method that requires the `'a` lifetime. Other methods have been changed to use `'static` streams (<https://github.com/apache/arrow-rs/pull/6619>) already, so this may be a leftover due to historical reasons.

I'd imagine the `delete_stream` method to be changed to the following.

```rust
    fn delete_stream(
        &self,
        locations: BoxStream<'static, Result<Path>>,
    ) -> BoxStream<'static, Result<Path>> {
````

This is a breaking change on the interface, so all the implementations need to be adjusted accordingly. Fortunately, the method won't become harder to implement since this is a restriction on the input so it refines the problem space for the implementer.

For users, this breaking change won't be too disruptive either, since the typical use case is to get the stream from `list` or `list_with_offset` (where the returned stream is of `'static` lifetime already), and then pass the stream to `delete_stream`.

Since this would be a breaking change, I wonder if this could be considered for the 0.13 release (<https://github.com/apache/arrow-rs-object-store/issues/367>).

N/A

**Additional context**

There was a more general discussion on this topic: <https://github.com/apache/arrow-rs/issues/6587>

### PR DESCRIPTION
**Pull request by linhr**

Closes #518.

# Rationale for this change

The current `delete_stream` method looks like this:

```rust
    fn delete_stream<'a>(
        &'a self,
        locations: BoxStream<'a, Result<Path>>,
    ) -> BoxStream<'a, Result<Path>>;
```
 
This PR changes the `delete_stream` method to work with `'static` lifetime. The reason for this change was discussed in the corresponding issue #518.

As I work on this PR, the changes appeared to be more complex than thought. The challenge is that `delete_stream` has a default implementation that takes `&self`, so the returned stream cannot be `'static`. I've considered the following options for this challenge.

**Option 1**: Only change the input stream to be `'static` without changing the lifetime of the output stream. The method would look like this:

```rust
    fn delete_stream<'a>(
        &'a self,
        locations: BoxStream<'static, Result<Path>>,
    ) -> BoxStream<'a, Result<Path>>;
```

Or equivalently (with unnecessary lifetime elided):

```rust
    fn delete_stream(
        &self,
        locations: BoxStream<'static, Result<Path>>,
    ) -> BoxStream<'_, Result<Path>>;
```

This would result in minimum code change but the result is still not very elegant, since we're not fully migrating to `BoxStream<'static, _>`.

**Option 2**: Change both the input and output stream to be `'static` and **requires the implementation to define `delete_stream` explicitly** (i.e. no more default implementation). The method would look like this:

```rust
    fn delete_stream(
        &self,
        locations: BoxStream<'static, Result<Path>>,
    ) -> BoxStream<'static, Result<Path>>;
```

I ended up choosing **Option 2** for this PR. This is more work but gives a clean end result so that **all `ObjectStore` methods now handle input/output `BoxStream` of `'static` lifetime**.

And fortunately, it likely won't be challenging for downstream implementations even if `delete_stream` is now **required**. Most implementations already use `Arc` for internal states, which can be cheaply cloned and moved into the `'static` output stream. From the changes I made to various `ObjectStore` implementations in this PR, we can see that the logic is straightforward. And I've documented the pattern in `lib.rs` for downstream implementations to follow.

This is a breaking change that could be considered for 0.13 (#367).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -874,24 +874,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     ///
     /// When supported, this method will use bulk operations that delete more
     /// than one object per a request. Otherwise, the implementation may call
-    /// the single object delete method for each location. For example, the
-    /// following implementation deletes each object with up to 10 concurrent
-    /// requests:
-    ///
-    /// ```ignore
-    /// let client = Arc::clone(&self.client);
-    /// locations
-    ///     .map(move |location| {
-    ///         let client = Arc::clone(&client);
-    ///         async move {
-    ///             let location = location?;
-    ///             client.delete(&location).await?;
-    ///             Ok(location)
-    ///         }
-    ///     })
-    ///     .buffered(10)
-    ///     .boxed()
-    /// ```
+    /// the single object delete method for each location.
     ///
     /// The returned stream yields the results of the delete operations in the
     /// same order as the input locations. However, some errors will be from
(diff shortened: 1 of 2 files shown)


linhr:
The changes are covered by existing tests since `delete_stream` is used in `src/integration.rs`.

src/lib.rs
```diff
@@ -873,9 +873,25 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
+    /// than one object per a request. Otherwise, the implementation may call
+    /// the single object delete method for each location. For example, the
+    /// following implementation deletes each object with up to 10 concurrent
+    /// requests:
+    ///
+    /// ```ignore
+    /// let client = Arc::clone(&self.client);
+    /// locations
+    ///     .map(move |location| {
+    ///         let client = Arc::clone(&client);
```
alamb: 👍

src/lib.rs
```diff
@@ -873,9 +873,25 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
...
     /// When supported, this method will use bulk operations that delete more
-    /// than one object per a request. The default implementation will call
-    /// the single object delete method for each location, but with up to 10
-    /// concurrent requests.
+    /// than one object per a request. Otherwise, the implementation may call
+    /// the single object delete method for each location. For example, the
+    /// following implementation deletes each object with up to 10 concurrent
+    /// requests:
+    ///
+    /// ```ignore
```
alamb: Could you please make this an actual example that compiles (so we are sure it won't drift over time)?

Perhaps following the model f the examples below.

We can do this as a follow on PR

src/lib.rs
```diff
@@ -873,9 +873,25 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// Delete all the objects at the specified locations
     ///
     /// When supported, this method will use bulk operations that delete more
-    /// than one object per a request. The default implementation will call
-    /// the single object delete method for each location, but with up to 10
-    /// concurrent requests.
+    /// than one object per a request. Otherwise, the implementation may call
```
alamb, line 876: We should probably call this out as a behavior change too in the API docs (but I think it is better to avoid implementing concurrent requests by default)

Also I think the documentation to explain how to get the old behavior is good

src/throttle.rs
```diff
@@ -237,6 +237,19 @@ impl<T: ObjectStore> ObjectStore for ThrottledStore<T> {
         self.inner.delete(location).await
     }
 
+    fn delete_stream(
```
alamb, line 240: I wonder if we should add a test for this new implementation, following the other tests at this file

src/lib.rs
linhr: Done!

src/lib.rs
linhr: Done!

I called out the behavior change in the docs and mentioned that the example can be used to get the previous behavior.

src/throttle.rs
linhr: Done!

src/lib.rs
```diff
@@ -873,9 +873,8 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// Delete all the objects at the specified locations
     ///
     /// When supported, this method will use bulk operations that delete more
```
kylebarron, line 875: It would be nice as an end user to know which backends (at least of those provided by object-store) support bulk delete

alamb: - Filed https://github.com/apache/arrow-rs-object-store/issues/527

src/lib.rs
```diff
@@ -927,6 +910,119 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// # let rt = tokio::runtime::Builder::new_current_thread().build().unwrap();
     /// # rt.block_on(example()).unwrap();
     /// ```
+    ///
+    /// Note: Before version 0.13, `delete_stream` has a default implementation
+    /// that deletes each object with up to 10 concurrent requests. This default
+    /// behavior has been removed, and each implementation must now provide its
+    /// own `delete_stream` implementation explicitly. The following example
+    /// shows how to implement `delete_stream` to get the previous default
```
alamb, line 918: ❤️

alamb:
Thanks for the review @kylebarron

PR #526, opened 2025-10-31
Implement ObjectStore for Arc<T> and Box<T>

### LINKED ISSUE #525: Implement ObjectStore for Arc<T> and Box<T>
I am working on some extensions to the ObjectStore trait. 
For example

```rust
pub trait ExtendedObjectStore: ObjectStore {
}

impl<T: ExtendedObjectStore> ExtendedObjectStore for Arc<T>{}

```

Implement ObjectStore for Arc<T> and Box<T> where T implements ObjectStore

None

wyatt-herkamp:

Closes #525 

# Rationale for this change
 
Allows you to put `Arc<T>` into something like struct `ObjectStoreWrapper<T: ObjectStore>(T)`

Implementing ObjectStore for `Arc<T>` and `Box<T>`

No and this should not break anything.  Just expand the flexibility of the ObjectStore Trait.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1885,9 +1885,8 @@ impl From<Error> for std::io::Error {
 #[cfg(test)]
 mod tests {
     use super::*;
-    
+
     use chrono::TimeZone;
-    
 
     macro_rules! maybe_skip_integration {
         () => {
@@ -1897,7 +1896,7 @@ mod tests {
             }
         };
     }
-    
+    pub(crate) use maybe_skip_integration;
 
     /// Test that the returned stream does not borrow the lifetime of Path
     fn list_store<'a>(
@@ -1939,6 +1938,9 @@ mod tests {
     {
         use bytes::Buf;
         use serde::Deserialize;
+        use tokio::io::AsyncWriteExt;
+
+        use crate::buffered::BufWriter;
 
         #[derive(Deserialize)]
         struct Tagging {


wyatt-herkamp:
> The reasoning for Box<T> is less clear to me because ObjectStore isn't mutable

I did Box<T> because it is pretty common for traits to implement `Box<T>`, and `Box<dyn ObjectStore>` was implemented.

src/lib.rs
```diff
@@ -1098,6 +1098,106 @@ macro_rules! as_ref_impl {
 as_ref_impl!(Arc<dyn ObjectStore>);
 as_ref_impl!(Box<dyn ObjectStore>);
 
+macro_rules! as_ref_generic_impl {
```
alamb: this macro looks very similar to `as_ref_impl` -- if we implement for all `Arc<T>`/`Box<T>` could we remove the explicit implementations for `Arc<dyn ..>` and `Box<dyn ..>` 🤔

wyatt-herkamp: We could, but I think it would be a breaking change.  I can merge the macros together by adding a second brand on the as_ref_impl macro if you would like?

kylebarron: It would be a breaking change?

wyatt-herkamp: Well, if we remove the implementation for `Arc<dyn ...>` and `Box<dyn ...>`, anyone who uses those implementations might have a problem.

Merging the macros isn't breaking. But removing those existing implementations might cause issues

kylebarron: My understanding of 

> if we implement for all `Arc<T>`/`Box<T>`

was that `dyn ObjectStore` already implemented `T`, so if you implement `impl<T: ObjectStore> ObjectStore for Arc<T>`, then maybe that already covers `Arc<dyn ObjectStore>`?

wyatt-herkamp: Oh I just remembered something!

Rust automatically adds the `Sized` bound. We need to add `+ ?Sized`

wyatt-herkamp: Ok `T: ObjectStore` doesn't cover `Box<dyn ObjectStore>` but `T: ObjectStore + ?Sized` does. 
The same applies to `Arc<T>`

kylebarron:
Looks like there's a merge conflict now

kylebarron:
Could we maybe add tests to verify that there's no breaking change here? That the existing implementations are included by the new bound?

kylebarron (approved):
I think this looks right, but I just got off a 14 hour flight 😅 so it would be great if @alamb can double check me

PR #530, opened 2025-11-05
refactor: move `put_multipart` to `ObjectStoreExt`

crepererum:
See #385 and #405.

# Rationale for this change
Similar to `put`, `put_multipart` is just `put_multipart_opts` with default options.

**Breaking:** `put_multipart` moved from `ObjectStore` to `ObjectStoreExt`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,11 +273,12 @@
 //!
 //! # Multipart Upload
 //!
-//! Use the [`ObjectStore::put_multipart`] method to atomically write a large amount of data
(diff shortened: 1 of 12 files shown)


src/lib.rs
mbrobbel:
```suggestion
/// Extension trait for [`ObjectStore`] with convenience functions.
```

mbrobbel (commented):
Noticed a typo in the other PR

src/lib.rs
crepererum: Thanks. Applied manually, because GH was convinced that the suggestion was outdated :roll_eyes:

mbrobbel: Yes, suggestions outside the diff of the PR can't be applied.

PR #531, opened 2025-11-05
fix(gcp): ignore ADC errors when explicit credentials are provided

nazq:
## Summary

Fixes an issue where `GoogleCloudStorageBuilder` would fail to build when Application Default Credentials (ADC) exist in an unsupported format, even when explicit credentials are provided via `with_service_account_path()`, `with_service_account_key()`, or `with_credentials()`.

## Problem

Previously, `GoogleCloudStorageBuilder::build()` unconditionally attempted to read ADC files and would fail immediately if the ADC format was unsupported (e.g., `external_account_authorized_user` from Workload Identity Federation with external identity providers). This prevented users from using explicit credentials in environments where ADC was configured with newer credential types.

## Solution

This PR makes ADC reading conditional:
- **When explicit credentials are provided**: ADC reading errors are ignored, allowing the builder to use the explicit credentials
- **When no explicit credentials exist**: ADC errors are propagated normally, preserving error visibility for users relying on ADC

The credential precedence remains unchanged: explicit credentials > ADC > instance metadata

## Changes

- Modified `src/gcp/builder.rs:495-503` to conditionally handle ADC reading errors based on whether explicit credentials were provided
- Added 4 comprehensive tests covering all credential paths

## Testing

All tests pass (113 passed, 0 failed):
- ✅ Explicit service account path ignores invalid ADC
- ✅ Explicit service account key ignores invalid ADC
- ✅ Custom credentials provider ignores invalid ADC
- ✅ ADC errors still propagate when no explicit credentials provided

## Impact

This change enables users in enterprise environments with Workload Identity Federation to use explicit credentials without being blocked by unsupported ADC formats, while maintaining backward compatibility and error visibility for ADC-only users.

### DIFF
--- a/src/gcp/builder.rs
+++ b/src/gcp/builder.rs
@@ -491,16 +491,15 @@ impl GoogleCloudStorageBuilder {
             };
 
         // Then try to initialize from the application credentials file, or the environment.
-        // Only fail on ADC errors if no explicit credentials were provided
-        let application_default_credentials = if service_account_credentials.is_none()
-            && self.credentials.is_none() {
-            // No explicit credentials, so ADC errors should be propagated
-            ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())?
-        } else {
-            // We have explicit credentials, so ignore ADC errors
-            ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())
-                .unwrap_or(None)
-        };
+        // Only attempt to read ADC if no explicit credentials were provided
+        let application_default_credentials =
+            if service_account_credentials.is_none() && self.credentials.is_none() {
+                // No explicit credentials, so try ADC and propagate errors
+                ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())?
+            } else {
+                // Explicit credentials provided, skip ADC reading entirely
+                None
+            };
 
         let disable_oauth = service_account_credentials
             .as_ref()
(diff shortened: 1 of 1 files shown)


src/gcp/builder.rs
```diff
@@ -491,8 +491,16 @@ impl GoogleCloudStorageBuilder {
...
-            ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())?;
+        // Only fail on ADC errors if no explicit credentials were provided
+        let application_default_credentials = if service_account_credentials.is_none()
+            && self.credentials.is_none() {
+            // No explicit credentials, so ADC errors should be propagated
+            ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())?
+        } else {
+            // We have explicit credentials, so ignore ADC errors
+            ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())
+                .unwrap_or(None)
```
crepererum: If we have credentials, then why do we attempt to call `ApplicationDefaultCredentials::read` at all? Can we not just skip it in this case?

nazq: Good catch! You're absolutely right - there's no need to attempt reading ADC at all when explicit credentials are already provided.

I've updated the code to skip the ADC read entirely in that case, changing:

```rust
} else {
    ApplicationDefaultCredentials::read(self.application_credentials_path.as_deref())
        .unwrap_or(None)
}
```

to:

```rust
} else {
    // Explicit credentials provided, skip ADC reading entirely
    None
}
```

This is cleaner, more efficient (no unnecessary file I/O)

PR #532, opened 2025-11-06
refactor!: move `get` to `ObjectStoreExt`

crepererum:
See #385 and #405.

# Rationale for this change
`get` is just `get_opts` with default options.

**Breaking:** `get` moved from `ObjectStore` to `ObjectStoreExt`.


kylebarron (approved):
Just a note on the conventional commit title (if we're trying to follow conventional commits?): since this is a breaking change I'd make the prefix `refactor!: description`

PR #534, opened 2025-11-06
Fix Azure Paths Parsing being broken since 0.12.3 #443

### LINKED ISSUE #443: az:// <container> not work as expected
**Describe the bug**
`az:// <container> / <blob> ` like uris (e.g. `az://testcontainer/ test.parquet` ) do not work as of version 0.12.3 at [pg_parquet](https://github.com/CrunchyData/pg_parquet) CI tests with azurite.

[This line change](https://github.com/apache/arrow-rs-object-store/pull/399/files#diff-5583e398a8deec11b274d9965eff8b5ade5226c7a020ca214a27d7d07dcc8a29R113) seems breaking the old behavior. 

**Before**
```sql
pg_parquet=# COPY (SELECT a FROM test) TO 'az://testcontainer/test.parquet' WITH (format 'parquet');
COPY 2
```

**After**
```sql
pg_parquet=# COPY (SELECT a FROM test) TO 'az://testcontainer/test.parquet' WITH (format 'parquet');
ERROR:  failed to finish parquet writer: External: Generic MicrosoftAzure error: Error performing PUT http://localhost:10000/devstoreaccount1/testcontainer in 9.424375ms - Server returned non-2xx status code: 400 Bad Request:
```

**Expected behavior**
It works as before in a similar way `azure://` prefix.

niltecedu:

Closes #443

# Rationale for this change
 
Two of the downstream projects, delta-rs and datafusion used the parse function which was changed on #399 
Which has led to those packages being rendered unusable with azure storage accounts since datafusion uses the direct https endpoints.

Delta-rs directly calls parse here:

https://github.com/delta-io/delta-rs/blob/main/crates/azure/src/lib.rs

Modifying/Reverting the strip bucket command along with azurite tests to not make it happen again.
Container name should always be included as its not a bug.

These might break obstore changed based the changes on #399  like
https://github.com/developmentseed/obstore/issues/477

Would need maintainers and collaborators to comment.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -307,6 +307,14 @@ mod tests {
                 (ObjectStoreScheme::MicrosoftAzure, "path"),
             ),
             (
+                "abfss://account/container",
+                (ObjectStoreScheme::MicrosoftAzure, ""),
+            ),
+            (
+                "abfss://account/container/path",
+                (ObjectStoreScheme::MicrosoftAzure, "path"),
+            ),
+            (
                 "gs://bucket/path",
                 (ObjectStoreScheme::GoogleCloudStorage, "path"),
             ),


niltecedu:
Sorry yeah completely forgot to finish this off actually, thanks for fixing!

PR #535, opened 2025-11-06
feat(gcp): add support for external_account_authorized_user credentials

nazq:
## Summary

Adds support for `external_account_authorized_user` credentials used by Google Cloud Workforce Identity Federation. This credential type is created when users authenticate with `gcloud auth application-default login` using workforce identity pools with external identity providers.

## Problem

Previously, `object_store` only supported two Application Default Credential types:
- `service_account` 
- `authorized_user`

Users with Workforce Identity Federation encountered deserialization errors when their ADC file contained `external_account_authorized_user` credentials, preventing them from using the library even though they had valid Google Cloud authentication.

This affected enterprise environments where organizations use external identity providers (like Azure AD, Okta, etc.) to authenticate users to Google Cloud resources.

## Solution

This PR extends credential support to include the `external_account_authorized_user` type:

1. **New credential struct** with OAuth2 fields (client_id, client_secret, refresh_token, token_url)
2. **TokenProvider implementation** using the STS (Security Token Service) OAuth token endpoint specified in the credential file
3. **Conversion to AuthorizedUserCredentials** for signing operations (same pattern as standard authorized_user)
4. **Builder integration** to handle the new credential type in selection logic

The implementation follows the same OAuth2 refresh flow as standard `authorized_user` credentials, but uses the custom `token_url` endpoint (typically `https://sts.googleapis.com/v1/oauthtoken`) specified in the credential file.

## Changes

**Files Modified:**
- `src/gcp/credential.rs`: Added ExternalAccountAuthorizedUserCredentials struct, enum variant, and TokenProvider implementation
- `src/gcp/builder.rs`: Updated credential selection logic to handle new type
- `src/gcp/mod.rs`: Added integration test

**Lines Changed:** +314 additions across 3 files

### Unit Tests
- ✅ Deserialization with full JSON format
- ✅ Deserialization with minimal required fields  
- ✅ Conversion to AuthorizedUserCredentials for signing
- ✅ Builder construction with sample credentials

### Integration Tests
- ✅ End-to-end test with put/get/delete operations (when TEST_INTEGRATION=1)
- ✅ Manual test with real ADC credentials (#[ignore])

**Test Results:**
```
test result: ok. 114 passed; 0 failed; 3 ignored
```

All existing tests continue to pass with no regressions.

## Credential Format Example

```json
{
  "type": "external_account_authorized_user",
  "audience": "//iam.googleapis.com/locations/global/workforcePools/pool/providers/provider",
  "client_id": "xxxxx.apps.googleusercontent.com",
  "client_secret": "secret",
  "refresh_token": "token",
  "token_url": "https://sts.googleapis.com/v1/oauthtoken",
  "token_info_url": "https://sts.googleapis.com/v1/introspect",
  "quota_project_id": "project-id"
}
```

## Impact

This change enables users in enterprise environments with Workforce Identity Federation to use `object_store` with their standard ADC configuration, matching the behavior of official Google Cloud client libraries (Python, Java, Go, etc.).

## References

- [Google Cloud Workforce Identity Federation](https://cloud.google.com/iam/docs/workforce-identity-federation)
- [Terraform Issue #34234](https://github.com/hashicorp/terraform/issues/34234) - Similar issue in Terraform

### DIFF
--- a/src/gcp/builder.rs
+++ b/src/gcp/builder.rs
@@ -533,6 +533,15 @@ impl GoogleCloudStorageBuilder {
                         self.retry_config.clone(),
                     )) as _
                 }
+                ApplicationDefaultCredentials::ExternalAccountAuthorizedUser(token) => Arc::new(
+                    TokenCredentialProvider::new(
+                        token,
+                        http.connect(&self.client_options)?,
+                        self.retry_config.clone(),
+                    )
+                    .with_min_ttl(TOKEN_MIN_TTL),
+                )
+                    as _,
             }
         } else {
             Arc::new(
@@ -565,6 +574,15 @@ impl GoogleCloudStorageBuilder {
                 }
                 ApplicationDefaultCredentials::ServiceAccount(token) => {
                     token.signing_credentials()?
+                }
+                ApplicationDefaultCredentials::ExternalAccountAuthorizedUser(token) => {
+                    // External account authorized user credentials don't have private keys,
+                    // so we use the same approach as AuthorizedUser
+                    Arc::new(TokenCredentialProvider::new(
+                        AuthorizedUserSigningCredentials::from(token.into())?,
+                        http.connect(&self.client_options)?,
+                        self.retry_config.clone(),
+                    )) as _
                 }
             }
         } else {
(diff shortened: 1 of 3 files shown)


src/gcp/builder.rs
```diff
@@ -746,4 +764,84 @@ mod tests {
...
+        assert!(
+            result.is_ok(),
+            "Build should succeed with external_account_authorized_user credentials: {:?}",
+            result.err()
+        );
+    }
+
+    #[test]
+    #[ignore] // Only run manually when testing with real ADC
+    fn gcs_test_real_external_account_authorized_user_adc() {
```
alamb, line 804: this test seems more like an example -- I wonder if it would be better as an example (or a no-run doc test)

src/gcp/mod.rs
```diff
@@ -450,4 +450,31 @@ mod test {
             err
         )
     }
+
+    #[tokio::test]
+    async fn gcs_test_external_account_authorized_user_integration() {
+        maybe_skip_integration!();
+
+        // This test verifies that external_account_authorized_user credentials
```
alamb, line 458: I don't understand how this test is verifying `external_account_authorized_user` -- it doesn't seem to configure such credentials 🤔

alamb (commented):
Thanks @nazq

PR #540, opened 2025-11-08
docs(client): improve warnings and links in client doc comments

CommanderStorm:

None, it is a cosmetic docs change.

# Rationale for this change
 
Looking though the client docs I noticed that some parts could be more consistent and others could be more highlighted.
This PR improves the docs in this department.

Changes like this:

image

🤷🏻‍♂️ It is only in the docs. Depends if you consider this user-facing. Likely yes. 😉

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -1060,10 +1060,9 @@ impl AmazonS3Builder {
                 (Some(_), None, _) => return Err(Error::MissingSecretAccessKey.into()),
                 (None, None, _) => unreachable!(),
             }
-        } else if let (Ok(token_path), Ok(role_arn)) = (
-            std::env::var("AWS_WEB_IDENTITY_TOKEN_FILE"),
-            std::env::var("AWS_ROLE_ARN"),
-        ) {
+        } else if let (Some(token_path), Some(role_arn)) =
+            (self.web_identity_token_file, self.role_arn)
+        {
             debug!("Using WebIdentity credential provider");
 
             let session_name = self
--- a/src/buffered.rs
+++ b/src/buffered.rs
@@ -19,8 +19,8 @@
 
 use crate::path::Path;
 use crate::{
-    Attributes, Extensions, ObjectMeta, ObjectStore, PutMultipartOptions, PutOptions,
-    PutPayloadMut, TagSet, WriteMultipart,
+    Attributes, Extensions, ObjectMeta, ObjectStore, ObjectStoreExt, PutMultipartOptions,
+    PutOptions, PutPayloadMut, TagSet, WriteMultipart,
 };
 use bytes::Bytes;
 use futures::future::{BoxFuture, FutureExt};
@@ -37,7 +37,7 @@ pub const DEFAULT_BUFFER_SIZE: usize = 1024 * 1024;
 
 /// An async-buffered reader compatible with the tokio IO traits
 ///
-/// Internally this maintains a buffer of the requested size, and uses [`ObjectStore::get_range`]
+/// Internally this maintains a buffer of the requested size, and uses [`ObjectStoreExt::get_range`]
 /// to populate its internal buffer once depleted. This buffer is cleared on seek.
 ///
 /// Whilst simple, this interface will typically be outperformed by the native [`ObjectStore`]
@@ -46,7 +46,7 @@ pub const DEFAULT_BUFFER_SIZE: usize = 1024 * 1024;
 /// round-trips is critical to throughput.
 ///
 /// Systems looking to sequentially scan a file should instead consider using [`ObjectStoreExt::get`],
-/// or [`ObjectStore::get_opts`], or [`ObjectStore::get_range`] to read a particular range.
+/// or [`ObjectStore::get_opts`], or [`ObjectStoreExt::get_range`] to read a particular range.
 ///
 /// Systems looking to read multiple ranges of a file should instead consider using
 /// [`ObjectStore::get_ranges`], which will optimise the vectored IO.
--- a/src/chunked.rs
+++ b/src/chunked.rs
@@ -61,6 +61,7 @@ impl Display for ChunkedStore {
 }
 
 #[async_trait]
+#[deny(clippy::missing_trait_methods)]
 impl ObjectStore for ChunkedStore {
     async fn put_opts(
         &self,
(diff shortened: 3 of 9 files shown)


src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
...
     pub fn with_allow_invalid_certificates(mut self, allow_insecure: bool) -> Self {
         self.allow_insecure = allow_insecure.into();
         self
     }
 
     /// Only use http1 connections
     ///
-    /// This is on by default, since http2 is known to be significantly slower than http1.
+    /// # See Also
+    /// * [`Self::with_http2_only`] if you only want to use http2
```
kylebarron: I think it would probably be good to have consistent casing of `HTTP1` and `HTTP2`

CommanderStorm: Changing the instances of the lowercased httpX is consistent with how we currently do doc comments.

Fixing the Spelling/punctuation is fine though. I fixed it in my last PR to HTTP/X which is used in the spec.

src/client/mod.rs
```diff
@@ -639,15 +673,15 @@ impl ClientOptions {
     /// Sets a timeout for receiving an acknowledgement of the keep-alive ping.
     ///
     /// If the ping is not acknowledged within the timeout, the connection will be closed.
-    /// Does nothing if http2_keep_alive_interval is disabled.
+    /// Does nothing if HTTP/2_keep_alive_interval is disabled.
```
kylebarron: I think this might need to stay lower case so it correctly points to the option

src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
...
 
-    /// Use http2 if supported, otherwise use http1.
+    /// Use HTTP/2 if supported, otherwise use HTTP/1.
+    ///
+    /// # See Also
+    /// * [`Self::with_http1_only`] if you only want to use HTTP/1
+    /// * [`Self::with_http2_only`] if you only want to use HTTP/2
+    ///
+    /// <div class="warning">
+    /// This is off by default, since HTTP/2 is known to be significantly slower than HTTP/1.
```
kylebarron: Also I don't know if this is correct; I usually think of HTTP2 being faster than HTTP1 because of multiplexing. But I know there has been some discussion somewhere on some issue here on why we don't default to HTTP2

CHANGELOG-old.md
```diff
@@ -475,7 +475,7 @@
 
 - Remove deprecated try\_with\_option methods [\#5237](https://github.com/apache/arrow-rs/pull/5237) [[object-store](https://github.com/apache/arrow-rs/labels/object-store)] ([tustvold](https://github.com/tustvold))
 - object\_store: full HTTP range support [\#5222](https://github.com/apache/arrow-rs/pull/5222) [[object-store](https://github.com/apache/arrow-rs/labels/object-store)] ([clbarnes](https://github.com/clbarnes))
-- feat\(object\_store\): use http1 by default [\#5204](https://github.com/apache/arrow-rs/pull/5204) [[object-store](https://github.com/apache/arrow-rs/labels/object-store)] ([wjones127](https://github.com/wjones127))
+- feat\(object\_store\): use HTTP/1 by default [\#5204](https://github.com/apache/arrow-rs/pull/5204) [[object-store](https://github.com/apache/arrow-rs/labels/object-store)] ([wjones127](https://github.com/wjones127))
```
kylebarron: also here I think better to leave the comment as it was

src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
         self
     }
 
-    /// Sets what protocol is allowed. If `allow_http` is :
-    /// * false (default):  Only HTTPS are allowed
-    /// * true:  HTTP and HTTPS are allowed
+    /// Sets what protocol is allowed.
+    ///
+    /// If `allow_http` is :
+    /// * `false` (default):  Only HTTPS are allowed
```
alamb:
```suggestion
    /// * `false` (default):  Only HTTPS is allowed
```

src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
...
     /// as a last resort or for testing
+    ///
+    /// </div>
     pub fn with_allow_invalid_certificates(mut self, allow_insecure: bool) -> Self {
         self.allow_insecure = allow_insecure.into();
         self
     }
 
-    /// Only use http1 connections
+    /// Only use HTTP/1 connections
```
alamb:
```suggestion
    /// Only use HTTP/1 connections (default)
```

src/client/mod.rs
alamb: I think the reference is here: 
- https://github.com/apache/arrow-rs-object-store/issues/104

The code, is here: 

https://github.com/apache/arrow-rs-object-store/blob/ddb5d2a18941d2da8df53bfca78a6e7b59567148/src/client/mod.rs#L303-L302

I found the term "off by default" confusing as it might refer to using HTTP/1 or the value of the setting

How about something like this

```suggestion
    /// HTTP/2 is not used by default. See details [#104](https://github.com/apache/arrow-rs-object-store/issues/104)
```

src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
...
+    ///
+    /// # See Also
+    /// * [`Self::with_http2_only`] if you only want to use HTTP/2
+    /// * [`Self::with_allow_http2`] if you want to use HTTP/1 or HTTP/2
     ///
-    /// This is on by default, since http2 is known to be significantly slower than http1.
+    /// <div class="warning">
+    /// This is off by default, since HTTP/2 is known to be significantly slower than http1.
```
alamb: Similarly to below, I suggest something more explicit

```suggestion
    /// HTTP/2 is not used by default. See details [#104](https://github.com/apache/arrow-rs-object-store/issues/104)
```

src/client/mod.rs
alamb: agreed

CHANGELOG-old.md
CommanderStorm:
```suggestion
- feat\(object\_store\): use http1 by default [\#5204](https://github.com/apache/arrow-rs/pull/5204) [[object-store](https://github.com/apache/arrow-rs/labels/object-store)] ([wjones127](https://github.com/wjones127))
```

src/client/mod.rs
CommanderStorm:
```suggestion
    /// Does nothing if `http2_keep_alive_interval` is disabled.
```

src/client/mod.rs
```diff
@@ -488,46 +492,76 @@ impl ClientOptions {
...
 
-    /// Only use http2 connections
+    /// Only use HTTP/2 connections
+    ///
+    /// # See Also
+    /// * [`Self::with_http1_only`] if you only want to use HTTP/1
+    /// * [`Self::with_allow_http2`] if you want to use HTTP/1 or HTTP/2
+    ///
+    /// <div class="warning">
+    /// This is off by default, since HTTP/2 is known to be significantly slower than HTTP/1.
```
CommanderStorm:
```suggestion
    /// HTTP/2 is not used by default. See details [#104](https://github.com/apache/arrow-rs-object-store/issues/104)
```

alamb:
> Sorry for the long wait from my side

No problems! Thank you for the contribution

alamb (approved):
Thank you @CommanderStorm  -- this looks like a great improvement to me

I am also interested to see what @kylebarron thinks too

PR #542, opened 2025-11-09
Use arc swap for credentials cache

### LINKED ISSUE #541: Heavy contention on credentials cache
**Describe the bug**

If you try to run ~1,000+ concurrent requests, the mutex on the credentials cache becomes a bottleneck

**To Reproduce**

Run several thousand concurrent requests, and you'll see performance drop (for our tests, p50 GET latency was 800ms on S3). Tracing revealed that almost all the time is spent acquiring the credentials mutex.

**Expected behavior**

Since the read:write ratio is so high, we should be able to avoid taking a lock in this case. Tests show that if you do, performance returns to the expected (8-20ms) p50 for HEAD/GET requests on S3.

ankrgyl:

Closes #541.

Use an arc swap to hold the cached credential, and acquire a mutex only if the credential's ttl is expired.

No

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,7 +61,7 @@ rustls-pemfile = { version = "2.0", default-features = false, features = ["std"]
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["std"], optional = true }
 serde_urlencoded = { version = "0.7", optional = true }
-tokio = { version = "1.29.0", features = ["sync", "macros", "rt", "time", "io-util"] }
+tokio = { version = "1.29.0", features = ["sync", "macros", "rt", "rt-multi-thread", "time", "io-util"] }
 
 [target.'cfg(target_family="unix")'.dev-dependencies]
 nix = { version = "0.30.0", features = ["fs"] }
@@ -107,3 +107,8 @@ features = ["js"]
 name = "get_range_file"
 path = "tests/get_range_file.rs"
 required-features = ["fs"]
+
+[[bench]]
+name = "cache_benchmark"
+harness = false
+required-features = []
--- a/benches/cache_benchmark.rs
+++ b/benches/cache_benchmark.rs
@@ -1,0 +1,412 @@
+use arc_swap::ArcSwapOption;
+use rand::Rng;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tokio::sync::{Mutex, RwLock};
+
+#[derive(Debug, Clone)]
+#[allow(dead_code)]
+struct Credentials {
+    token: String,
+    expiry: Instant,
+}
+
+// Arc-swap based cache implementation
+#[derive(Debug)]
+struct ArcSwapCache {
+    cache: ArcSwapOption<Credentials>,
+}
+
+impl ArcSwapCache {
+    fn new() -> Self {
+        Self {
+            cache: ArcSwapOption::new(None),
+        }
+    }
+
+    fn get(&self) -> Option<Credentials> {
+        self.cache.load_full().as_ref().map(|c| (**c).clone())
+    }
+
+    fn update(&self, creds: Credentials) {
+        self.cache.store(Some(Arc::new(creds)));
+    }
+}
+
+// RwLock based cache implementation
+#[derive(Debug)]
+struct RwLockCache {
+    cache: RwLock<Option<Credentials>>,
+}
+
+impl RwLockCache {
+    fn new() -> Self {
+        Self {
+            cache: RwLock::new(None),
+        }
+    }
+
+    async fn get(&self) -> Option<Credentials> {
+        self.cache.read().await.clone()
+    }
+
+    async fn update(&self, creds: Credentials) {
+        *self.cache.write().await = Some(creds);
+    }
+}
+
+// Mutex based cache implementation (original problematic approach)
+#[derive(Debug)]
+struct MutexCache {
+    cache: Mutex<Option<Credentials>>,
+}
+
+impl MutexCache {
+    fn new() -> Self {
+        Self {
+            cache: Mutex::new(None),
+        }
+    }
+
+    async fn get(&self) -> Option<Credentials> {
+        self.cache.lock().await.clone()
+    }
+
+    async fn update(&self, creds: Credentials) {
+        *self.cache.lock().await = Some(creds);
+    }
+}
+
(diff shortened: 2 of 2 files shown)


src/client/token.rs
```diff
@@ -33,15 +35,17 @@ pub(crate) struct TemporaryToken<T> {
 /// [`TemporaryToken`] based on its expiry
 #[derive(Debug)]
 pub(crate) struct TokenCache<T> {
-    cache: Mutex<Option<(TemporaryToken<T>, Instant)>>,
+    cache: ArcSwapOption<CacheEntry<T>>,
```
alamb, line 38: So in general, unless we have benchmark results that show arc-swap is necessary, I am opposed to adding a new dependency

Did you try a `RWLock` before reaching for a new crate? I always worry about adding new crates like [arcswap](https://crates.io/crates/arc-swap) as I don't want to have to deal with a RUSTSEC report if/when it becomes abandoned.

 I do see there are many other users

RWLocks would allow multiple concurrent readers but if you had a lot of writers you might still have contention.  If the you find update contention is too much, you could change to use `RWLock<Arc<..>>` so that the lock only needs to be held to clone an `Arc`

I understand the docs for arc swap claims https://docs.rs/arc-swap/latest/arc_swap/

> Better option would be to have [RwLock<Arc<T>>](https://doc.rust-lang.org/std/sync/struct.RwLock.html). Then one would lock, clone the [Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html) and unlock. This suffers from CPU-level contention (on the lock and on the reference count of the [Arc](https://doc.rust-lang.org/nightly/alloc/sync/struct.Arc.html)) which makes it relatively slow. Depending on the implementation, an update may be blocked for arbitrary long time by a steady inflow of readers.


I would imagine the overhead of actually using the token (making an HTTP request) is pretty huge compared to getting a lock.

ankrgyl: Sorry I just saw this after writing my other comment.

> I would imagine the overhead of actually using the token (making an HTTP request) is pretty huge compared to getting a lock.

The problem with the previous design, which may not apply to an RwLock (and sure, I will benchmark it and report back), is that "waiting in line" for the mutex would become so expensive with a high number of concurrent requests (eg. HEAD requests with 8ms p50 latencies), that it actually overwhelmed tokio's worker threads and dominated the execution time (we saw p50 "HEAD" operation latency spike to 700ms, and realized the mutex was the root cause).

Let me run a benchmark with arc swap vs. RwLock and report back

alamb: > Let me run a benchmark with arc swap vs. RwLock and report back

👍

alamb:
> Here are the benchmark numbers run on my macbook pro. What I saw in production was dramatically more pronounced (mutex vs. arc swap, not arc swap vs. rwlock), but I don't have much more bandwidth to investigate that in depth.

What did you benchmark? Is this a micro benchmark for locking, or is it actually a workload running object_store requests?

given the total request time reported is in ms I am guessing the actual workload you have

crepererum:
Also looking at the code, you roughly have something like this:

```
if token_fresh() {
  return token;
}

let _guard = refresh.lock().await;

...

write_token();
```

I don't see how an `ArcSwap` is really gonna help here. In fact with that construct the `RwLock` doesn't even need to be a tokio async lock. You can just use a stdlib or parking_lot `RwLock` since getting the fresh token doesn't involve I/O and replacing it while being under the refresh guard at the very end is also just a sync write operation. My suspicion is that that would be even faster, since async locks also have some overhead that we technically don't need here.

ankrgyl:
> > And yes I stated above that this benchmark does not illustrate the difference as extremely as what I saw in production. Perhaps it's because in a real workload, we're doing a lot more on the runtime than just GET operations, and that accentuates the impact of additional polls.
> 
> It might make sense to look into using a separate threadpool for CPU and IO work.
> 
> For example, you can move all your object store work to a different threadpool (tokio runtime) using the [SpawnedReqwestConnector](https://docs.rs/object_store/latest/object_store/client/struct.SpawnedReqwestConnector.html). There is an end to end example in datafusion: https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/thread_pools.rs
> 
> Something we spent quite a long time at InfluxData was that io/network latencies increased substantially with highly concurrent workloads. We eventually tracked this down to using the same threadpool (tokio pool) for CPU and IO work -- doing so basically starves the IO of the CPU it needs to make progress in the TCP state machine , and it seems that then the tcp stack treats the system as being congested and slows down traffic.

I appreciate it! We do a lot of this and are constantly optimizing it. I deployed the change on our end using `arc_swap`, and with no other changes, saw a pretty substantial impact on reported latency (and for us, indexing throughput).

I don't mind at all if you are uninterested in this contribution. I was mostly submitting it to pay-it-back and say thank you for this library. I'm very happy to just run our fork. Feel free to let me know what you'd like from this point onwards.

crepererum:
FWIW: I don't question that `ArcSwap` is better than the plain `Mutex`, but I don't think it's actually necessary. They way the code is written, the same effect can likely be gained using `RwLock`.

> It might make sense to look into using a separate threadpool for CPU and IO work.

I don't think there's a CPU contention here and IMHO using a threadpool would be totally overkill. The code in `main` has a reader contention for many concurrent requests though.

alamb:
> I don't mind at all if you are uninterested in this contribution. I was mostly submitting it to pay-it-back and say thank you for this library. I'm very happy to just run our fork. Feel free to let me know what you'd like from this point onwards.

Thank you -- we very much appreciate it

I think we are trying to get to the bottom of what is going on / figure out the best solution. In my opinion switching from `Mutex` to `RWLock` is a clear win and I would be happy to accept such a PR (or maybe will find time to write one myself)

I feel like we are still trying to figure out how much benefit, if any, the new Arc swap library really brings so we can make a final judgement call about if the new dependency is worth it

crepererum:
> the evidence is _not_ glaringly obvious from these micro benchmarks

Do you see a difference between `RwLock` and `AtomicArc` in your production environment? Because if you do, we could work on better benchmarks to replicate that.

ankrgyl:
> > the evidence is _not_ glaringly obvious from these micro benchmarks
> 
> Do you see a difference between `RwLock` and `AtomicArc` in your production environment? Because if you do, we could work on better benchmarks to replicate that.

I have not tested it yet. I can give it a go and report back

tustvold:
I've created https://github.com/apache/arrow-rs-object-store/pull/648 which switches to using RWLock as there appears to be consensus that this is a good idea

PR #543, opened 2025-11-10
refactor!: move `head` to `ObjectStoreExt`

crepererum:
- part of  #385 
- part of #405.

# Rationale for this change
`head` is just `get_opts` + `with_head(true)`.

**Breaking:** `head` moved from `ObjectStore` to `ObjectStoreExt`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -834,12 +834,6 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
         .await
     }
 
-    /// Return the metadata for the specified location
-    async fn head(&self, location: &Path) -> Result<ObjectMeta> {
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -1274,6 +1267,11 @@ where
         let options = GetOptions::new().with_range(Some(range));
         self.get_opts(location, options).await?.bytes().await
     }
+
+    async fn head(&self, location: &Path) -> Result<ObjectMeta> {
+        let options = GetOptions::new().with_head(true);
```
alamb, line 1272: TIL that GetOpts already has `with_head` 🤯 

The more I see of these PRs to extract code into ObjectStoreExt the better I like it

alamb (approved):
Thank you @crepererum

Issue #545, opened 2025-11-11
Improve `Path` ergonomics


Kinrany:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

`Path` is central to `object_store`, but manipulating values of this type can be hard because its API is mostly structured around one-way parsing.

**Describe the solution you'd like**
Expand `Path`s API. In no particular order:

- `fn empty()` constructor, equivalent to `Path::from("/")`
- `fn is_empty(&self) -> bool`, true if `self.raw` is empty
- `fn len(&self) -> usize`, equivalent to `.parts().count()`
- `fn prefix(&self) -> Self`, dual to the existing `Parts::filename`
  - Bikeshedding: `fn folder`?
- Change `parts` to return a custom iterator that is fused and double-ended.
- Implement `Expand` in addition to `FromIterator`

**Describe alternatives you've considered**
One could argue that `Path` is not designed to be used as a standalone type. But then I would expect it to be internal to `object_store`, with all APIs using `String` instead.

**Additional context**
TODO: PR link

kylebarron:
I think `Path` exists because there are multiple distinct ways to construct a path: `Path::from` and `Path::parse` do different things.

PR #546, opened 2025-11-11
`Path` improvements

Kinrany:

Closes https://github.com/apache/arrow-rs-object-store/issues/545

# Rationale for this change
 
`fn prefix(&self)` was the first thing I really wanted to have on `Path`, as opposed to copying from project to project.

It also benefits from direct access to `raw`. In userland we have to use `paths` and collect, and omitting the last item is not trivial.

Implementations of everything listed with minimal docs and tests.

Most of these changes are user-facing.

Internal changes:

 - `impl FromIterator` delegates most of the logic to `impl Expand`
 - `Path::parts` now returns a custom iterator type
 - Added `DELIMITER_CHAR`

There should be no breaking changes.

### DIFF
--- a/src/path/mod.rs
+++ b/src/path/mod.rs
@@ -307,7 +307,13 @@ impl Path {
     ///
     /// Returns `None` if this path has zero segments.
     pub fn parent(&self) -> Option<Self> {
-        let (prefix, _filename) = self.raw.rsplit_once(DELIMITER)?;
+        if self.raw.is_empty() {
+            return None;
+        }
+
+        let Some((prefix, _filename)) = self.raw.rsplit_once(DELIMITER) else {
+            return Some(Self::ROOT);
+        };
 
         Some(Self {
             raw: prefix.to_string(),
@@ -762,6 +768,14 @@ mod tests {
 
         p.extend(&path("baz"));
         assert_eq!(p, path("foo/bar/baz"));
+    }
+
+    #[test]
+    fn parent() {
+        assert_eq!(Path::ROOT.parent(), None);
+        assert_eq!(path("foo").parent(), Some(Path::ROOT));
+        assert_eq!(path("foo/bar").parent(), Some(path("foo")));
+        assert_eq!(path("foo/bar/baz").parent(), Some(path("foo/bar")));
     }
 
     /// Construct a [`Path`] from a raw `&str`, or panic trying.


Kinrany:
One alternative I'm considering is a `fn prefix` that takes ownership and avoids cloning. The caller can always clone the original path first if necessary.

src/path/mod.rs
```diff
@@ -255,14 +262,39 @@ impl Path {
...
-    pub fn parts(&self) -> impl Iterator<Item = PathPart<'_>> {
-        self.raw
-            .split_terminator(DELIMITER)
-            .map(|s| PathPart { raw: s.into() })
+    /// Returns the number of [`PathPart`]s in this [`Path`]
+    ///
+    /// # Performance
+    ///
+    /// This operation is `O(n)`, similar to calling `.parts().count()` manually.
+    pub fn len(&self) -> usize {
```
tustvold: This collides with the existing len provided by the deref to str

Edit: actually we don't implement deref but still callings this parts_count or something might be less ambiguous

Kinrany: Would renaming `empty` and `is_empty` into `root` and `is_root` make sense?
Edit: done now.

src/path/mod.rs
```diff
@@ -255,14 +269,53 @@ impl Path {
...
+    }
+
+    /// Returns the [`PathPart`]s of this [`Path`]
+    pub fn parts(&self) -> PathParts<'_> {
+        PathParts::new(&self.raw)
+    }
+
+    /// Returns a copy of this [`Path`] with the last path segment removed
+    ///
+    /// Returns `None` if this path has zero segments.
```
alamb, line 308: this seems pretty similar to https://doc.rust-lang.org/std/path/struct.Path.html#method.parent

Should we call this method `parent` as well to follow existing convention?

src/path/mod.rs
```diff
@@ -469,6 +533,23 @@ mod tests {
         assert_eq!(Path::default().parts().count(), 0);
     }
 
+    #[test]
+    fn parts_count() {
+        assert_eq!(Path::ROOT.parts().count(), Path::ROOT.parts_count());
+
+        let path = path("foo/bar/baz");
+        assert_eq!(path.parts().count(), path.parts_count());
```
alamb: could you also please add an assertion that the count is `3`?

src/path/mod.rs
```diff
@@ -348,13 +398,22 @@ where
...
-            .join(DELIMITER);
+        let mut this = Self::ROOT;
+        this.extend(iter);
+        this
+    }
+}
 
-        Self { raw }
+impl<'a, I: Into<PathPart<'a>>> Extend<I> for Path {
+    fn extend<T: IntoIterator<Item = I>>(&mut self, iter: T) {
```
alamb, line 444: could you please add tests (perhaps doctests) using this feature explicitly to help others find it?

src/path/mod.rs
```diff
@@ -348,13 +398,22 @@ where
...
+        for s in iter.into_iter() {
+            let s = s.into();
+            if s.raw.is_empty() {
+                continue;
+            }
+            self.raw.push(DELIMITER_CHAR);
```
alamb: Can you add some tests showing how this works for:
1. Empty iterators
2. Iterator with a single element 
3. Iterator with multiple elements
4. Extending an existing (non empty) Path? (it looks like maybe this will add an extra `/` 🤔 )

alamb (approved):
Thanks @Kinrany  -- these changes look like a nice improvement to me

I think the code needs a few more tests -- doctests are fine (maybe even best) but otherwise this is ready to go in my mind

alamb:
please mark it ready for review when it is ready for another look

Kinrany:
I was going to say I'll probably do it this weekend. But, oh look, I only meant to see how much I'll need to change and now it's done. Haha.

I added `impl IntoIterator for &Path`. I think that isn't controversial since it would be symmetrical to existing `FromIterator`?

alamb:
Sounds reasonable (but I haven't thought about it deeply yet)

alamb:
> Is there a way to run the same checks that run in CI locally? Should I do anything other than `cargo test && cargo clippy`?
> 
> When tests run, a few items end up marked as unused. I assume that's fine. Or should I fix that in main in a separate PR first? Could be a toolchain update thing.

You can look at the output of the CI runs and see what commands it is running (they are all listed in various yml files in `.github` as well(

src/path/mod.rs
Kinrany: Hmm, note that I haven't added any docs on `Path` directly. Since this is a standard trait, I expect people to find it by looking at the list of implemented traits 🤔

Kinrany:
That should do it. All the new stuff has tests directly covering it now.

Kinrany:
Thank you!

PR #547, opened 2025-11-12
Use builder fields instead of env vars for determining credential provider

### LINKED ISSUE #538: AmazonS3ConfigKey::WebIdentityTokenFile is ignored
**Describe the bug**

AmazonS3ConfigKey::WebIdentityTokenFile and AmazonS3ConfigKey::RoleArn config is ignored when using AWS WebIdentity credential provider. It only works if you set the AWS_WEB_IDENTITY_TOKEN_FILE and AWS_ROLE_ARN environment variables.

In src/aws/builder.rs it only checks the environment variables, not the config:
```
        } else if let (Ok(token_path), Ok(role_arn)) = (
            std::env::var("AWS_WEB_IDENTITY_TOKEN_FILE"),
            std::env::var("AWS_ROLE_ARN"),
        ) {
            debug!("Using WebIdentity credential provider");
```

**To Reproduce**

Do not set the environment variables and use a builder like the following:
```
        let builder = AmazonS3Builder::new()
            .with_bucket_name("some-bucket")
            .with_config(
                AmazonS3ConfigKey::WebIdentityTokenFile,
                "/tmp/fake-token-file",
            )
            .with_config(
                AmazonS3ConfigKey::RoleArn,
                "arn:aws:iam::123456789012:role/test-role",
            )
            .with_config(AmazonS3ConfigKey::RoleSessionName, "TestSession")
            .with_config(
                AmazonS3ConfigKey::StsEndpoint,
                "https://sts.us-west-2.amazonaws.com",
            );
```

**Expected behavior**

It should try and get a key from the STS endpoint (but doesn't)

Friede80:

Closes #538.

# Rationale for this change
 
Setting `web_identity_token_file` and `role_arn` via the config would be ignored as the builder was still only checking the env vars.

Setting the AWS web identity fields via the config builder will no longer be ignored

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -1060,10 +1060,9 @@ impl AmazonS3Builder {
                 (Some(_), None, _) => return Err(Error::MissingSecretAccessKey.into()),
                 (None, None, _) => unreachable!(),
             }
-        } else if let (Ok(token_path), Ok(role_arn)) = (
-            std::env::var("AWS_WEB_IDENTITY_TOKEN_FILE"),
-            std::env::var("AWS_ROLE_ARN"),
-        ) {
+        } else if let (Some(token_path), Some(role_arn)) =
+            (self.web_identity_token_file, self.role_arn)
+        {
             debug!("Using WebIdentity credential provider");
 
             let session_name = self


src/aws/builder.rs
```diff
@@ -1060,10 +1060,9 @@ impl AmazonS3Builder {
                 (Some(_), None, _) => return Err(Error::MissingSecretAccessKey.into()),
                 (None, None, _) => unreachable!(),
             }
-        } else if let (Ok(token_path), Ok(role_arn)) = (
-            std::env::var("AWS_WEB_IDENTITY_TOKEN_FILE"),
-            std::env::var("AWS_ROLE_ARN"),
-        ) {
+        } else if let (Some(token_path), Some(role_arn)) =
```
alamb, line 1063: This change makes sense to me -- thank you @Friede80 and @kylebarron 

I double checked that the environment variables are still honored and I think they will be -- specifically, they will be parsed when creating from the environment here:

https://github.com/apache/arrow-rs-object-store/blob/1b8ecc7f41eab2a3c2c373f1c2fe21bebe391727/src/aws/builder.rs#L566-L565

https://github.com/apache/arrow-rs-object-store/blob/1b8ecc7f41eab2a3c2c373f1c2fe21bebe391727/src/aws/builder.rs#L491-L494

Issue #297, opened 2024-01-05
Cross Bucket Copy Support


tustvold:
_Draft as needs further design, and a motivating use-case_

**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Currently ObjectStore only supports copying files within the current bucket, this is despite most stores having APIs for performing efficient copies across buckets within the same cloud provider.

**Describe the solution you'd like**

I would like some mechanism to support copying between ObjectStore, that is able to make use of the more efficient APIs where they are possible. I don't have a good idea of how to achieve this, nor a motivating use-case for this functionality. 

**Describe alternatives you've considered**

**Additional context**

Joe-Blount:
I also would find it useful to have some form cross bucket copy support. We're using Arrow and the `DynObjectStore` type. For backup and D/R purposes, we need to selectively copy some objects to another bucket. Doing it server side without buffering the object is necessary to allow high concurrency without high memory usage. I'm imagining a copy API that takes another `DynObjectStore` that's either the source or destination bucket (with `Self` being the other). But anything that lets me do server-side copies between buckets would be greatly appreciated.

Joe-Blount:
Thanks for the idea, yes we've looked at that.  In our case, the primary bucket is very high churn, and we need the second bucket to be low churn, only containing a few selected objects. Tagging the selected objects (and using that tag to drive replication) could work, but then we'd give up control on the ordering they're copied in.  So we might as well just copy them ourselves and retain control of the copy order.

tustvold:
Fair, just thought I'd suggest it as it wouldn't end up waiting on changes to object_store.

> I'm imagining a copy API that takes another DynObjectStore that's either the source or destination bucket (with Self being the other)

It's unclear to me how such an API would work, ObjectStore does not expose `Any` nor would it be especially useful if it did given people tend to wrap ObjectStore in multiple layers.

I suspect #116 and adding the ability to provide a source bucket is the way to go here, but I'm afraid I don't really have the capacity to help design or implement such an API. #121 is also potentially related.

crepererum:
> ObjectStore does not expose `Any` nor would it be especially useful if it did given people tend to wrap ObjectStore in multiple layers.

That could be changed though:

- support `Any`: either via `as_any` or -- for new rust versions -- via dyn up/downcasting
- layers: add a new method to the base trait `fn inner(&self) -> Option<&dyn ObjectStore>`.

[...]

PR #548, opened 2025-11-20
refactor!: `copy` & `copy_if_not_exists` => `copy_opts`

### LINKED ISSUE #116: Add CopyOptions
_Largely a placeholder as I lack a motivating use-case for this functionality_

Following apache/arrow-rs#4984 we now have put_opts and get_opts. It makes sense to expose the conditional copy semantics of many stores, in particular:

* Source preconditions, such as `If-Match`
* Destination preconditions, e.g. `PutMode`

I would like a copy_opts method, which is called by default implementations of copy and copy_if_not_exists, and supports the common preconditions supported by object stores.

**Additional context**

crepererum:
Closes #116.

# Rationale for this change
Change the `ObjectStore` core trait to have a single, extensible copy operation. This helps #385 and #297.

Also adds extensions similar to 
https://github.com/apache/arrow-rs/pull/7170
and
https://github.com/apache/arrow-rs/pull/7213 .

Interface change, see section below.

- new core method `copy_opts` alongside `CopyOptions` & `CopyMode`
- `copy` & `copy_if_not_exists` are moved to `ObjectStoreExt`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -890,7 +890,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// # use futures::stream::{BoxStream, StreamExt};
     /// # use object_store::path::Path;
     /// # use object_store::{
-    /// #     GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
+    /// #     CopyOptions, GetOptions, GetResult, ListResult, MultipartUpload, ObjectMeta, ObjectStore,
     /// #     PutMultipartOptions, PutOptions, PutPayload, PutResult, Result,
     /// # };
     /// # use std::fmt;
@@ -969,11 +969,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// #         todo!()
     /// #     }
     /// #
-    /// #     async fn copy(&self, _: &Path, _: &Path) -> Result<()> {
-    /// #         todo!()
-    /// #     }
-    /// #
-    /// #     async fn copy_if_not_exists(&self, _: &Path, _: &Path) -> Result<()> {
+    /// #     async fn copy_opts(&self, _: &Path, _: &Path, _: CopyOptions) -> Result<()> {
     /// #         todo!()
     /// #     }
     /// # }
@@ -1033,9 +1029,7 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     async fn list_with_delimiter(&self, prefix: Option<&Path>) -> Result<ListResult>;
 
     /// Copy an object from one path to another in the same object store.
-    ///
-    /// If there exists an object at the destination, it will be overwritten.
-    async fn copy(&self, from: &Path, to: &Path) -> Result<()>;
+    async fn copy_opts(&self, from: &Path, to: &Path, options: CopyOptions) -> Result<()>;
 
     /// Move an object from one path to another in the same object store.
     ///
@@ -1047,15 +1041,6 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
         self.copy(from, to).await?;
         self.delete(from).await
     }
-
-    /// Copy an object from one path to another, only if destination is empty.
-    ///
-    /// Will return an error if the destination already has an object.
-    ///
-    /// Performs an atomic operation if the underlying object storage supports it.
-    /// If atomic operations are not supported by the underlying object storage (like S3)
-    /// it will return an error.
-    async fn copy_if_not_exists(&self, from: &Path, to: &Path) -> Result<()>;
 
     /// Move an object from one path to another in the same object store.
     ///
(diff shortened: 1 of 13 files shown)


src/lib.rs
```diff
@@ -1633,7 +1638,7 @@ pub struct PutOptions {
     ///
     /// These extensions are ignored entirely by backends offered through this crate.
     ///
-    /// They are also eclused from [`PartialEq`] and [`Eq`].
+    /// They are also excluded from [`PartialEq`] and [`Eq`].
```
alamb, line 1641: 👍

src/lib.rs
```diff
@@ -1756,6 +1761,73 @@ pub struct PutResult {
...
+    /// Perform an atomic write operation, returning [`Error::AlreadyExists`] if an
+    /// object already exists at the provided path
+    Create,
+}
+
+/// Options for a copy request
+#[derive(Debug, Clone, Default)]
+pub struct CopyOptions {
+    /// Configure the [`CopyMode`] for this operation
+    pub mode: CopyMode,
```
alamb, line 1779: If we make all these fields `pub` I don't think we will be able to add new fields (e.g a different target object store) without a breaking API change

So I would suggest either marking these as non-pub or marking the struct as [`non_exhaustitve`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute)

src/lib.rs
```diff
@@ -1756,6 +1761,73 @@ pub struct PutResult {
...
+
+    /// Sets the `extensions`.
+    ///
+    /// See [`CopyOptions::extensions`].
+    #[must_use]
+    pub fn with_extensions(mut self, extensions: Extensions) -> Self {
+        self.extensions = extensions;
+        self
+    }
+}
```
alamb, line 1812: I recommend also adding accessors and remove the `pub`  from the fields for the reasons listed above

alamb (commented):
Thank you @crepererum 

I found reviewing with whitespace blind diff easier to understand: https://github.com/apache/arrow-rs-object-store/pull/548/files?w=1

I have a few suggestions related to API but I don't think they are required

cc @kylebarron

src/lib.rs
crepererum: Adding new fields IS a breaking change by design, implementations shouldn't just ignore them. This is also in line with all the other options structs. This is also enforced by fully destructing the options in the respective methods (so you cannot forget to handle new fields).

src/lib.rs
crepererum: I disagree as per https://github.com/apache/arrow-rs-object-store/pull/548#discussion_r2546894411

src/lib.rs
alamb: My point was that one of the stated  rationales for extracting options into `CopyOptions` is to help enable
- https://github.com/apache/arrow-rs-object-store/issues/297

However, it seems like if we go with this API we will not be able to add fields to CopyOptions to enable cross bucket copying without another breaking API change

Perhaps I am missing something

kylebarron: It would be nice to be able to support cross-bucket copies in the future without a breaking change if possible

crepererum: I cannot square the circle here: having a simple API (esp. not having hundreds of specialized methods or traits) + no breaking change is IMHO not gonna work. TBH I don't see what's the reason for avoiding breaking changes, it's a 0.x crate after all. And it's the question of "how much breaking" it is. We're not re-designing the entire interface for that again.

Different example: if we would add "attributes" to the copy operation (which might be actually happening at one point), then a store implementation HAS TO consider them. You cannot just YOLO it and drop that information to the floor. Hence, this would be a breaking change, not only because of the interface change, but purely based on the fact that there's a semantic change/addition to the operation that implementation HAVE TO consider.

crepererum: > My point was that one of the stated rationales for extracting options into CopyOptions is to help enable

The rational is to add such feature easier, not to do it without a breaking change at all.

alamb: > TBH I don't see what's the reason for avoiding breaking changes, it's a 0.x crate after all. And it's the question of "how much breaking" it is. We're not re-designing the entire interface for that again.

One rationale is to speed up the adoption downstream -- since object_store is "low" in the dependency stack, if we make a breaking change to the API, we then need to wait for that breaking change releases in all other downstream dependencies

So for example, it can take up to 4 months to get into datafusion (wait 3 months for next major arrow release and then a month for the next major datafusion release)

That timeframe is not a deal breaker nor does it mean we shouldn't make breaking changes ever, but I do think it is worth considering as we design APIs

tustvold: I think the only way to achieve this would be to make the options enumerable, i.e. a list of some non_exhaustive enum, that way implementations can error if they encounter an option they don't recognise.

Effectively this would take the pattern of Attributes and apply it more broadly.

We could do this, but I think that's probably a separate ticket, and we'd want to do it globally.

alamb: One possibility would be to setup that style as part of this PR (first for CopyOptions) and then apply it to other options / attributes in a series of follow on PRs

crepererum: > We could do this, but I think that's probably a separate ticket, and we'd want to do it globally.

So we're now waiting for the ideal solution? Should we do that for 0.13.x?

tustvold: I don't have strong opinions on the matter, I lack the time to drive such an initiative myself, but if someone else is willing to do so, I think it would be valuable

alamb: Given we have a PR here I vote we proceed with this PR and file a ticket to track a more easily migratable API going forward. I will file such a ticket

alamb: - Filed https://github.com/apache/arrow-rs-object-store/issues/553

PR #549, opened 2025-11-20
refactor!: move `delete` to `ObjectStoreExt`

crepererum:
- part of #385 
- part of #405.

# Rationale for this change
`delete` is just `delete_stream` with a single element.

**Breaking:** `delete` moved from `ObjectStore` to `ObjectStoreExt`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1304,12 +1304,12 @@ where
             self.delete_stream(futures::stream::once(async move { Ok(location) }).boxed());
         let _path = stream.try_next().await?.ok_or_else(|| Error::Generic {
             store: "ext",
-            source: "`delete_stream` was supposed to yield once but didn't".into(),
+            source: "`delete_stream` with one location should yield once but didn't".into(),
         })?;
         if stream.next().await.is_some() {
             Err(Error::Generic {
                 store: "ext",
-                source: "`delete_stream` yielded more than once".into(),
+                source: "`delete_stream` with one location expected to yield exactly once, but yielded more than once".into(),
             })
         } else {
             Ok(())


src/lib.rs
```diff
@@ -1306,6 +1298,24 @@ where
...
 
+    async fn delete(&self, location: &Path) -> Result<()> {
+        let location = location.clone();
+        let mut stream =
+            self.delete_stream(futures::stream::once(async move { Ok(location) }).boxed());
+        let _path = stream.try_next().await?.ok_or_else(|| Error::Generic {
+            store: "ext",
+            source: "`delete_stream` was supposed to yield once but didn't".into(),
+        })?;
+        if stream.next().await.is_some() {
```
alamb, line 1309: 👍

src/lib.rs
```diff
@@ -1306,6 +1298,24 @@ where
         Ok(self.get_opts(location, options).await?.meta)
     }
 
+    async fn delete(&self, location: &Path) -> Result<()> {
+        let location = location.clone();
+        let mut stream =
+            self.delete_stream(futures::stream::once(async move { Ok(location) }).boxed());
+        let _path = stream.try_next().await?.ok_or_else(|| Error::Generic {
+            store: "ext",
+            source: "`delete_stream` was supposed to yield once but didn't".into(),
```
alamb: Maybe it would help to explain why the stream was supposed to yield once

```suggestion
            source: "`delete_stream` with one location should yield once but didn't".into(),
```

src/lib.rs
```diff
@@ -1306,6 +1298,24 @@ where
...
+        })?;
+        if stream.next().await.is_some() {
+            Err(Error::Generic {
+                store: "ext",
+                source: "`delete_stream` yielded more than once".into(),
```
alamb:
```suggestion
                source: "`delete_stream` with one location expected to yield exactly once, but yielded more than once".into(),
```

PR #551, opened 2025-11-20
Improve Documentation on `ObjectStoreExt`

alamb:

- Part of https://github.com/apache/arrow-rs-object-store/issues/385

# Rationale for this change

One major API change in the next `object_store` release is the addition of the `ObjectStoreExt` trait
and the removal of several APIs from `ObjectStore` into `ObjectStoreExt`.

This will be a breaking change for downstream users of `object_store` so it is important
that we have good documentation about the rationale for hte change (and what they get out of the change)
and how to migrate to the new APIs.

Add some documentation to the `object_store` crate root and to the `ObjectStoreExt` trait itself

Yes more docs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -644,6 +644,56 @@ pub type MultipartId = String;
 /// described above. See [#385] for more details.
 ///
 /// [#385]: https://github.com/apache/arrow-rs-object-store/issues/385
+///
+/// # Wrappers
+/// If you wrap an [`ObjectStore`] -- e.g. to add observability -- you SHOULD
+/// implement all trait methods. This ensures that defaults implementations
+/// that are overwritten by the wrapped store are also used by the wrapper.
+/// For example:
+///
+/// ```ignore
+/// struct MyStore {
+///     ...
+/// }
+///
+/// #[async_trait]
+/// impl ObjectStore for MyStore {
+///     // implement custom ranges handling
+///     async fn get_ranges(
+///         &self,
+///         location: &Path,
+///         ranges: &[Range<u64>],
+///     ) -> Result<Vec<Bytes>> {
+///         ...
+///     }
+///
+///     ...
+/// }
+///
+/// struct Wrapper {
+///     inner: Arc<dyn ObjectStore>,
+/// }
+///
+/// #[async_trait]
+/// #[deny(clippy::missing_trait_methods)]
+/// impl ObjectStore for Wrapper {
+///     // If we would not implement this method,
+///     // we would get the trait default and not
+///     // use the actual implementation of `inner`.
+///     async fn get_ranges(
+///         &self,
+///         location: &Path,
+///         ranges: &[Range<u64>],
+///     ) -> Result<Vec<Bytes>> {
+///         ...
+///     }
+///
+///     ...
+/// }
+/// ```
+///
+/// To automatically detect this issue, use
+/// [`#[deny(clippy::missing_trait_methods)]`](https://rust-lang.github.io/rust-clippy/master/index.html#missing_trait_methods).
 #[async_trait]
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -615,66 +620,31 @@ pub type MultipartId = String;
...
-///     // we would get the trait default and not
-///     // use the actual implementation of `inner`.
-///     async fn get_ranges(
-///         &self,
-///         location: &Path,
-///         ranges: &[Range<u64>],
-///     ) -> Result<Vec<Bytes>> {
-///         ...
-///     }
+/// There are no default implementation for any of the methods in this trait by
```
alamb: This is not true (yet) but I believe that the goal for 0.13.0 is to remove all default implementations --  @crepererum  can you confirm this is the plan / goal?

crepererum: That's not gonna happen for 0.13, e.g. `list_prefixes` and `get_ranges` will still have default impls. I don't think we gonna remove them in 0.13, since they cannot easily be mapped to `_opts` methods either.

alamb: Ok, I will revise this wording

src/lib.rs
```diff
@@ -615,66 +620,31 @@ pub type MultipartId = String;
 
 /// Universal API to multiple object store services.
 ///
-/// For more convenience methods, check [`ObjectStoreExt`].
+/// See the [module-level documentation](crate) for a high leve overview and
```
kylebarron:
```suggestion
/// See the [module-level documentation](crate) for a high level overview and
```

src/lib.rs
```diff
@@ -615,66 +620,31 @@ pub type MultipartId = String;
...
+/// examples. See [`ObjectStoreExt`] for additional convenience methods.
 ///
 /// # Contract
-/// This trait is meant as a contract between object store implementations
-/// (e.g. providers, wrappers) and the `object_store` crate itself and is
+/// This trait is a contract between object store implementations
```
kylebarron:
```suggestion
/// This trait is a contract between object store _implementations_
```

src/lib.rs
```diff
@@ -615,66 +620,31 @@ pub type MultipartId = String;
...
-///         location: &Path,
-///         ranges: &[Range<u64>],
-///     ) -> Result<Vec<Bytes>> {
-///         ...
-///     }
-///
-///     ...
-/// }
+/// The [`ObjectStoreExt`] acts as an API/contract between `object_store` and
+/// the store users and provides additional methods that may be simpler to use
```
kylebarron:
```suggestion
/// the store _users_ and provides additional methods that may be simpler to use
```

kylebarron (approved):

src/lib.rs
alamb: Refined in 69eaabf

src/lib.rs
```diff
@@ -613,68 +618,32 @@ pub type DynObjectStore = dyn ObjectStore;
...
-/// (e.g. providers, wrappers) and the `object_store` crate itself and is
+/// This trait is a contract between object store _implementations_
+/// (e.g. providers, wrappers) and the `object_store` crate itself. It is
 /// intended to be the minimum API required for an object store.
 ///
-/// The [`ObjectStoreExt`] acts as an API/contract between `object_store`
-/// and the store users and provides additional methods that may be simpler to use but overlap
-/// in functionality with [`ObjectStore`].
-///
-/// # Wrappers
```
crepererum, line 629: I think the wrappers section is still relevant. In fact NOT using the clippy lint is still a foot gun (#537 proofs that even `object_store` itself suffered from this and we had a wide range of cases at Influx too).

alamb: done in https://github.com/apache/arrow-rs-object-store/pull/551/commits/c4c4f1c0b0dc8eb46bd1c4be5b6ceec0df3f164a

PR #555, opened 2025-11-26
 refactor!:  Consolidate `ObjectStore`: `rename` & `rename_if_not_exists` => `rename_opts`

crepererum:
This helps #385 and #297.

# Rationale for this change
Change the `ObjectStore` core trait to have a single, extensible copy operation.

Also adds extensions similar to 
https://github.com/apache/arrow-rs/pull/7170
and
https://github.com/apache/arrow-rs/pull/7213 .

Also see #548 for the same change that was made for `copy`.

Interface change, see section below.

- new core method `rename_opts` alongside `RenameOptions` & `RenameMode`
- `rename` & `rename_if_not_exists` are moved to `ObjectStoreExt`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1035,19 +1035,22 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     ///
     /// By default, this is implemented as a copy and then delete source. It may not
     /// check when deleting source that it was the same object that was originally copied.
-    ///
-    /// If there exists an object at the destination, it will be overwritten.
-    async fn rename(&self, from: &Path, to: &Path) -> Result<()> {
-        self.copy(from, to).await?;
-        self.delete(from).await
-    }
-
-    /// Move an object from one path to another in the same object store.
-    ///
-    /// Will return an error if the destination already has an object.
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1828,6 +1855,76 @@ impl PartialEq<Self> for CopyOptions {
 
 impl Eq for CopyOptions {}
 
+/// Configure preconditions for the target of rename operation.
```
alamb, line 1858: I did some research about the names used in cloud providers

S3 uses the term `If-None-Match` for the "don't overwrite" method: https://docs.aws.amazon.com/AmazonS3/latest/API/API_RenameObject.html

From what I can tell, GCP doesn't support an atoimic rename/overwrite operation. If you want to overwrite you have to delete the target first : https://docs.cloud.google.com/storage/docs/copying-renaming-moving-objects

Azure Blob Storage calls it "replace if exists", specifically https://learn.microsoft.com/en-us/rest/api/storageservices/rename-file `x-ms-file-rename-replace-if-exists`


So TLDR I think these options are well named and reflect the available capabilities 👍

Issue #121, opened 2023-10-20
Support Object Composition


tustvold:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Various stores have facilities for object composition:

* https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html
* https://cloud.google.com/storage/docs/xml-api/put-object-compose
* https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-from-url

**Describe the solution you'd like**

Not sure, but documenting that this is something supported by the stores but not the `ObjectStore` trait

**Describe alternatives you've considered**

**Additional context**

PR #561, opened 2025-12-03
aws: support multipart copy for objects larger than 5GB

james-rms:

- close https://github.com/apache/arrow-rs-object-store/issues/563

# Rationale for this change

Today, users that attempt to copy a >5GB object in S3 using `object_store` will see this error:
```
Server returned non-2xx status code: 400 Bad Request: 
<Error><Code>InvalidRequest</Code><Message>
The specified copy source is larger than the maximum allowable size for a copy source: 5368709120
</Message></Error>
```
The way to get around this problem per AWS's docs is to do the copy in several parts using multipart copies. This PR adds that functionality to the AWS client.

It adds two additional configuration parameters:

```rust
    /// The size threshold above which copy uses multipart copies under the hood. defaults to 5GB.
    multipart_copy_threshold: u64
    /// When using multipart copies, the part size used. Defaults to 5GB.
    multipart_copy_part_size: u64
```

The defaults are chosen to minimise surprise: if people are used to copies not requiring several requests, we don't switch to that method until it's absolutely necessary, and when necessary, we use as few parts as possible.

See above.

Yes - these configuration parameters should be covered by the docstring changes.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -46,6 +46,11 @@ static DEFAULT_METADATA_ENDPOINT: &str = "http://169.254.169.254";
 /// [CopyObject](https://docs.aws.amazon.com/AmazonS3/latest/userguide/copy-object.html) for more
 /// details.
 const MAX_SINGLE_REQUEST_COPY_SIZE: u64 = 5 * 1024 * 1024 * 1024;
+/// AWS S3 rejects multipart upload parts smaller than 5 MiB, unless they are the last part of the
+/// upload. See
+/// [CompleteMultipartUpload](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html)
+/// for more details.
+const MIN_PART_SIZE_BYTES: u64 = 5 * 1024 * 1024;
 
 /// A specialized `Error` for object store-related errors
 #[derive(Debug, thiserror::Error)]
@@ -1242,6 +1247,13 @@ impl AmazonS3Builder {
             .map(|val| val.get())
             .transpose()?
             .unwrap_or(MAX_SINGLE_REQUEST_COPY_SIZE);
+
+        if multipart_copy_part_size < MIN_PART_SIZE_BYTES {
+            return Err(crate::Error::Generic {
+                source: "Multipart copy part size must be >= 5MB".into(),
+                store: STORE,
+            });
+        }
 
         let config = S3Config {
             region,
--- a/src/aws/mod.rs
+++ b/src/aws/mod.rs
@@ -662,7 +662,7 @@ mod tests {
         let store = AmazonS3Builder::from_env()
             .with_bucket_name(bucket)
             .with_multipart_copy_threshold(5 * 1024 * 1024)
-            .with_multipart_copy_part_size(1337 * 1024)
+            .with_multipart_copy_part_size(6 * 1024 * 1024)
             .build()
             .unwrap();
 
@@ -687,7 +687,7 @@ mod tests {
     async fn small_file_copy_single_part() {
         maybe_skip_integration!();
 
-        let bucket = "test-bucket-for-multipart-copy-small";
+        let bucket = "test-bucket-for-multipart-copy-single";
         let store = AmazonS3Builder::from_env()
             .with_bucket_name(bucket)
             .with_multipart_copy_threshold(5 * 1024 * 1024) // trigger multipart copy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -132,7 +132,7 @@ jobs:
           aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket-for-checksum
           aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket-for-copy-if-not-exists
           aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket-for-multipart-copy-large
-          aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket-for-multipart-copy-small
+          aws --endpoint-url=http://localhost:4566 s3 mb s3://test-bucket-for-multipart-copy-single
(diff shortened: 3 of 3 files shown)


james-rms:
Great, created https://github.com/apache/arrow-rs-object-store/issues/563.

alamb (commented):
Thanks @james-rms  and @tustvold  -- the high level idea seems reasonable to me, but I think this code needs tests (maybe unit tests?) or something otherwise we may break the functionality inadvertently in some future refactor

src/aws/mod.rs
```diff
@@ -314,14 +368,31 @@ impl ObjectStore for AmazonS3 {
             mode,
             extensions: _,
         } = options;
+        // Determine source size to decide between single CopyObject and multipart copy
```
tustvold: Is there some way we can avoid this, e.g. we try CopyObject normally and on error fallback to multipart? Otherwise this adds an additional S3 roundtrip to every copy request.

src/aws/mod.rs
```diff
@@ -583,6 +626,63 @@ mod tests {
...
+        let bucket = "test-bucket-for-multipart-copy-large";
+        let store = AmazonS3Builder::from_env()
+            .with_bucket_name(bucket)
+            .with_multipart_copy_threshold(5 * 1024 * 1024)
+            .with_multipart_copy_part_size(5 * 1024 * 1024)
+            .build()
+            .unwrap();
+
+        let mut payload = BytesMut::zeroed(10 * 1024 * 1024);
+        rand::fill(&mut payload[..]);
```
tustvold, line 670: 👍

src/aws/mod.rs
```diff
@@ -583,6 +626,63 @@ mod tests {
...
 
+    #[tokio::test]
+    async fn large_file_copy_multipart() {
+        maybe_skip_integration!();
+
+        let bucket = "test-bucket-for-multipart-copy-large";
+        let store = AmazonS3Builder::from_env()
+            .with_bucket_name(bucket)
+            .with_multipart_copy_threshold(5 * 1024 * 1024)
+            .with_multipart_copy_part_size(5 * 1024 * 1024)
```
tustvold:
```suggestion
            .with_multipart_copy_part_size(1337 * 1024)
```

Perhaps to exercise some more interesting code paths?

tustvold (commented):
Looks good, and well tested, sorry for the delay in reviewing, been absolutely swamped.

I think we need to find a way to avoid regressing the common case of files smaller than 5GB, e.g. by first attempting CopyObject and then falling back if it errors (I am presuming S3 gives a sensible error here).

tustvold:
I guess another option would be to make this disabled by default and therefore opt-in... How do other clients handle this?

james-rms:
No worries, and thanks for putting up with my regular pings on this one!

src/aws/mod.rs
```diff
@@ -101,6 +101,74 @@ impl AmazonS3 {
...
+        mode: CompleteMultipartMode,
+    ) -> Result<()> {
+        // Perform multipart copy using UploadPartCopy
+        let upload_id = self
+            .client
+            .create_multipart(to, PutMultipartOptions::default())
+            .await?;
+
+        let mut parts = Vec::new();
+        for (idx, payload) in self
```
crepererum, line 144: Technically we could do perform these requests concurrently (at least up to a certain amount), but I think the serial implementation is "good enough" for now. It's certainly better than the status quo (i.e. not being able to copy the object at all).

src/aws/builder.rs
```diff
@@ -189,6 +199,11 @@ pub struct AmazonS3Builder {
     request_payer: ConfigValue<bool>,
     /// The [`HttpConnector`] to use
     http_connector: Option<Arc<dyn HttpConnector>>,
+    /// Threshold (bytes) above which copy uses multipart copy. If not set, all copies are performed
+    /// as single requests.
+    multipart_copy_threshold: Option<ConfigValue<u64>>,
+    /// Preferred multipart copy part size (bytes). If not set, defaults to 5 GiB.
```
crepererum, line 205: That's a question I have but that I also think should be included in the doc-string: if the object was created using a multi-part upload, are there any requirements on the alignment of the copy-parts or is this irrelevant?

src/aws/mod.rs
```diff
@@ -315,13 +374,37 @@ impl ObjectStore for AmazonS3 {
...
+                    CopyMethod::SingleRequest => {
+                        self.client
+                            .copy_request(from, to)
+                            .idempotent(true)
+                            .send()
+                            .await?;
+                    }
+                    CopyMethod::Multipart(size) => {
+                        self.copy_multipart(from, to, size, CompleteMultipartMode::Overwrite)
+                            .await?;
```
crepererum, line 405:
```suggestion
                        tracing::debug!(%from, %to, "multipart copy");
                        self.copy_multipart(from, to, size, CompleteMultipartMode::Overwrite)
                            .await?;
```

(you may need to check if `Path` implements `Display` or not)

Reasoning: we are changing the client behavior dynamically based on data size and if a server misbehaves due to that, it would be nice if we would at least leave some breadcrumbs.

james-rms: Seems reasonable, i'll take a look.

src/aws/builder.rs
james-rms: It's not explicitly documented ([though there is a note around range requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-guidelines.html#optimizing-performance-guidelines-get-range)). There's every chance it's faster to line up the part boundaries exactly, but I haven't tested this directly.

I hadn't thought about it and this bothers me. It doesn't feel good to be opaquely putting part boundaries into a copied object that the user has no visibility into. This also points to the right API being one where the user creates parts directly and keeps track of the boundaries.

tustvold: FWIW https://github.com/apache/arrow-rs-object-store/issues/121

PR #562, opened 2025-12-04
(0.12.4 cherry-pick): aws: support copies >5GB

james-rms:

Closes #.

# Rationale for this change
 

### DIFF
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -721,24 +721,24 @@ impl S3Client {
             // If SSE-C is used, we must include the encryption headers in every upload request.
             request = request.with_encryption_headers();
         }
-        let (parts, body) = request.send().await?.into_parts();
-        let checksum_sha256 = parts
-            .headers
-            .get(SHA256_CHECKSUM)
-            .and_then(|v| v.to_str().ok())
-            .map(|v| v.to_string());
 
-        let e_tag = match is_copy {
-            false => get_etag(&parts.headers).map_err(|source| Error::Metadata { source })?,
-            true => {
-                let response = body
-                    .bytes()
-                    .await
-                    .map_err(|source| Error::CreateMultipartResponseBody { source })?;
-                let response: CopyPartResult = quick_xml::de::from_reader(response.reader())
-                    .map_err(|source| Error::InvalidMultipartResponse { source })?;
-                response.e_tag
-            }
+        let (parts, body) = request.send().await?.into_parts();
+        let (e_tag, checksum_sha256) = if is_copy {
+            let response = body
+                .bytes()
+                .await
+                .map_err(|source| Error::CreateMultipartResponseBody { source })?;
+            let response: CopyPartResult = quick_xml::de::from_reader(response.reader())
+                .map_err(|source| Error::InvalidMultipartResponse { source })?;
+            (response.e_tag, response.checksum_sha256)
+        } else {
+            let e_tag = get_etag(&parts.headers).map_err(|source| Error::Metadata { source })?;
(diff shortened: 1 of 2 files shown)


src/aws/mod.rs
```diff
@@ -101,6 +103,56 @@ impl AmazonS3 {
...
+        }
+        .await;
+
+        // If the multipart upload failed, make a best effort attempt to
+        // clean it up. It's the caller's responsibility to add a
+        // lifecycle rule if guaranteed cleanup is required, as we
+        // cannot protect against an ill-timed process crash.
+        if res.is_err() {
+            let _ = self.client.abort_multipart(to, &upload_id).await;
+        }
```
Muon, line 160: Goes in docstring?

src/aws/builder.rs
```diff
@@ -1185,8 +1193,20 @@ impl AmazonS3Builder {
...
+        let multipart_copy_threshold = self
+            .multipart_copy_threshold
+            .map(|val| val.get())
+            .transpose()?
+            .unwrap_or(MAX_SINGLE_REQUEST_COPY_SIZE);
+        let multipart_copy_part_size = self
+            .multipart_copy_part_size
+            .map(|val| val.get())
+            .transpose()?
+            .unwrap_or(MAX_SINGLE_REQUEST_COPY_SIZE);
```
Muon, line 1205: Clamp to 5GiB because that's the documented maximum?

src/aws/mod.rs
```diff
@@ -305,89 +365,80 @@ impl ObjectStore for AmazonS3 {
...
+            .client
+            .get_opts(
+                from,
+                GetOptions {
+                    head: true,
+                    ..Default::default()
+                },
+            )
+            .await?
+            .meta;
```
Muon, line 387: Should this first try to use CopyObject and then fall back if it fails due to size?

src/aws/mod.rs
```diff
@@ -101,6 +103,56 @@ impl AmazonS3 {
...
+
+        // S3 requires minimum 5 MiB per part (except final) and max 10,000 parts
+        let part_size = self.client.config.multipart_copy_part_size;
+
+        let mut parts = Vec::new();
+        let mut offset: u64 = 0;
+        let mut idx: usize = 0;
+        let res = async {
+            while offset < size {
+                let end = std::cmp::min(offset + part_size, size);
```
Muon:
```suggestion
                let end = if size - offset <= part_size { size } else { offset + part_size };
```
Avoids overflow and I think is a little clearer.

src/aws/builder.rs
james-rms: I think if someone wants to push it over 5GB, they should be able to. There are many "s3-compatible" object stores that might not share the same limitations.

src/aws/mod.rs
james-rms: Good point - let me see if that's straightforward.

james-rms: The issue with that approach is that on error, AWS does not respond with anything more specific than "InvalidRequest": 
```
<Error><Code>InvalidRequest</Code><Message>The specified copy source is larger than the maximum allowable size for a copy source: 5368709120</Message><RequestId>8550KAYYHRYF33SM</RequestId><HostId>R7zaiPWt96z/yQm2PtDT+pyFmYF76YCBcW0AeukdrXpS4qlSuO1nmXTFI4Ak2YcHMsBoymw33j4=</HostId></Error>
```

So there's not really a stable API for determining that the request is invalid because of the size of the source.

PR #567, opened 2025-12-08
Allow explicitly specifying GCS base URL

### LINKED ISSUE #566: Allow explicitly specifying the GCS base URL
> [!NOTE]
> I've opened a PR addressing this issue [here](https://github.com/apache/arrow-rs-object-store/pull/567)

I need to build GCS `object_store` that communicate with a non-standard GCS endpoint (i.e. not `https://storage.googleapis.com`).

This is already somewhat supported, by reading the `gcs_base_url` key of a provided service account credentials JSON file ([here](https://github.com/apache/arrow-rs-object-store/blob/main/src/gcp/builder.rs#L511)). I believe this exists for the needs of integration tests mostly.

Unfortunately, in my production use-case, for infra reasons, I cannot provide such a credentials file.

I would like to be able to explicitly specify the base URL to be used, as simply as this:
```rust
GoogleCloudStorageBuilder::from_env()
  .with_base_url("https://some-other-gcs-endpoint")
  .build();
```

This is already supported on the AWS S3 object store, see [here](https://github.com/apache/arrow-rs-object-store/blob/main/src/aws/builder.rs#L822-L838).

Passing a service account credentials file, but doesn't fit my requirements.

**Additional context**

None.

rgehan:

Closes #566

# Rationale for this change
 
Currently, the only way to override the GCS base URL is to provide a service account credentials file that contains a `gcs_base_url`.

If for some reason the user cannot provide such a file (because they're authenticated a different way), being able to explicitly override the base URL is useful.

This can also be useful while testing locally, against a local implementation of GCS (I'm not personally doing this, but I suppose this is a valid use case)

This PR adds a `.with_base_url()` method on the GCP builder, accompanied with unit tests.

This override takes precedence over the URL from the service account credentials file, but is optional, so this should be strictly compatible with the previous implementation. 

New `.with_base_url()` on the GCS builder.

No breaking changes.

### DIFF
--- a/src/gcp/builder.rs
+++ b/src/gcp/builder.rs
@@ -160,7 +160,7 @@ pub enum GoogleConfigKey {
     /// - `bucket`
     /// - `bucket_name`
     Bucket,
-    
+
     /// Base URL
     ///
     /// See [`GoogleCloudStorageBuilder::with_base_url`] for details.
@@ -384,7 +384,7 @@ impl GoogleCloudStorageBuilder {
     }
 
     /// Sets the base URL for communicating with GCS.
-    /// 
+    ///
     /// If not explicitly set, it will be:
     /// 1. Derived from the service account credentials, if provided
     /// 2. Otherwise, uses the default GCS endpoint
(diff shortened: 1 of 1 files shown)


tustvold (commented):
Thank you for this, I think we should probably add a config key to match this, otherwise looks good

rgehan:
@tustvold I've added a configuration key, and fixed lint/doc issues. Let me know if there's anything else I can do!

kylebarron (approved):

PR #570, opened 2025-12-10
Implement typos-cli

jayvdb:

# Rationale for this change
 
Keep docs, comments, variable names, etc in good health.

Uses https://github.com/taiki-e/install-action for fast and checksum verified installation.

No


_typos.toml
crepererum, line 1: This file needs a license header similar to this:

https://github.com/apache/arrow-rs-object-store/blob/c9475c5c4eb73d5b23e242497ccecc6861c52858/Cargo.toml#L1-L16

jayvdb:
@crepererum rebased and fixed a new typo

Issue #527, opened 2025-10-31
Add documentation for what backends support bulk delete (`ObjectStore::delete_stream`)


alamb:
It would be nice as an end user to know which backends (at least of those provided by object-store) support bulk delete

_Originally posted by @kylebarron in https://github.com/apache/arrow-rs-object-store/pull/524#discussion_r2481745713_

PR #571, opened 2025-12-10
Documentation for backend support of bulk delete

jayvdb:

Closes https://github.com/apache/arrow-rs-object-store/issues/527

# Rationale for this change
 

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -862,6 +862,23 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
     /// than one object per a request. Otherwise, the implementation may call
     /// the single object delete method for each location.
     ///
+    /// # Bulk Delete Support
+    ///
+    /// The following backends support native bulk delete operations:
+    ///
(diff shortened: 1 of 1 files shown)


src/parse.rs
```diff
@@ -50,16 +50,16 @@ impl From<Error> for super::Error {
 /// # Supported formats:
 /// - `file:///path/to/my/file` -> [`LocalFileSystem`]
 /// - `memory:///` -> [`InMemory`]
-/// - `s3://bucket/path` -> [`AmazonS3`](crate::aws::AmazonS3) (also supports `s3a`)
```
kylebarron, line 53: Why remove all these hyperlinks?

jayvdb: They are errors when running `cargo doc`

kylebarron: You need to run with `--all-features`

jayvdb: ahh, of course! ok , I'll revert.  Thanks!

kylebarron (approved):

alamb:
Thanks @kylebarron and @jayvdb

Issue #573, opened 2025-12-11
local path filtering has different semantics


jayvdb:
**Describe the bug**

I ran into a problem where the local object store wasnt behaving the same as the other object stored.  I havent determined if this was intentional or not, but it doesnt feel like it.

When calling `list()` or `list_with_delimiter()` with a prefix parameter, the local filesystem backend treated the prefix as a directory path and walked only that directory, instead of properly filtering by prefix using `Path::prefix_matches` semantics.

**To Reproduce**

**Expected behavior**

The local backend behaves similar to the other backends, as much as possible.

**Additional context**

PR #574, opened 2025-12-11
fix(local): prefix filtering with Path::prefix_matches

jayvdb:
The LocalFileSystem backend incorrectly treated the prefix parameter as a directory path, walking only that specific directory instead of filtering results with prefix_matches.

Changes:
- Walk from parent directory instead of treating prefix as directory
- Filter results using Path::prefix_matches
- Handle special case: listing exact file path returns empty

Fixes https://github.com/apache/arrow-rs-object-store/issues/573

# Rationale for this change
 
So the backends work in a similar way.

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -699,14 +699,57 @@ impl LocalFileSystem {
         prefix: Option<&Path>,
         maybe_offset: Option<&Path>,
     ) -> BoxStream<'static, Result<ObjectMeta>> {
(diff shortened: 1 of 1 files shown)


tustvold:
All of these tests pass on current main, nor do I really understand what this PR is trying to fix?

tustvold:
Chalking this up as being AI slop

PR #576, opened 2025-12-13
Add support for aws-lc-rs

faysou:
## Cryptographic backend abstraction

This PR introduces support for selecting between `ring` and `aws-lc-rs` as cryptographic backends at compile time, providing users with flexibility in their cryptographic library choice while maintaining API compatibility.

#### Feature flags and dependencies

- Added `aws-lc-rs` as an optional dependency with feature flag `aws-lc`.

- Updated `Cargo.toml` to support both `ring` and `aws-lc-rs` backends:
  - `ring` remains the default backend (included in `default` features).
  - New feature flags: `azure-aws-lc`, `gcp-aws-lc`, `aws-aws-lc`, `http-aws-lc` for using `aws-lc-rs` with specific cloud providers.
  - Features `ring` and `aws-lc` are mutually exclusive.

- Updated `reqwest` TLS configuration from `rustls-tls-native-roots` to `rustls-tls-webpki-roots-no-provider` to support both backends.

#### Cryptographic abstraction layer

- Created new `crypto_backend.rs` module that provides a unified interface over `ring` and `aws-lc-rs`.

- The module re-exports the chosen backend as `crypto`, allowing existing code to work with either backend transparently.

- Added `rsa_key_modulus_len()` helper function to abstract API differences between `ring` and `aws-lc-rs` for RSA key operations.

#### Code updates

- Updated AWS client (`src/aws/client.rs`) to use the abstracted crypto backend instead of direct `ring` imports.

- Updated GCP credential handling (`src/gcp/credential.rs`) to use the abstracted crypto backend and the new `rsa_key_modulus_len()` helper.

- Updated utility functions (`src/util.rs`) for HMAC-SHA256 and SHA256 digest operations to use the abstracted backend.

- Updated all feature gate conditions throughout the codebase to include the new `*-aws-lc` feature variants:
  - `builder.rs`: Updated feature gates for HTTP request building methods.
  - `header.rs`: Updated feature gates for header parsing functions.
  - `body.rs`: Updated feature gates for JSON deserialization.
  - `mod.rs`: Updated feature gates for cloud provider modules and client options.
  - `retry.rs`: Updated feature gates for retry configuration.
  - `s3.rs`: Updated feature gates for S3-specific types.
  - `token.rs`: Updated feature gates for token caching.
  - `lib.rs`: Updated feature gates for module exports.

### Design decisions

- **API compatibility**: Since `aws-lc-rs` is API-compatible with `ring` for most operations, a simple re-export pattern was used rather than a trait-based abstraction, minimizing code complexity.

- **Compile-time selection**: The backend choice is made at compile time via feature flags, avoiding runtime overhead and ensuring type safety.

- **Mutual exclusivity**: The `ring` and `aws-lc` features are mutually exclusive to prevent conflicts and ensure clear backend selection.

- **Backward compatibility**: The default configuration continues to use `ring`, ensuring existing users are not affected unless they explicitly opt into `aws-lc-rs`.

### Testing considerations

- All existing tests should continue to pass with the default `ring` backend.

- New feature flag combinations (`*-aws-lc`) should be tested to ensure proper compilation and functionality with the `aws-lc-rs` backend.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ url = "2.2"
 walkdir = { version = "2", optional = true }
 
 # Cloud storage support
+aws-lc-rs = { version = "1.15", default-features = false, features = ["non-fips", "ring-io"], optional = true }
 base64 = { version = "0.22", default-features = false, features = ["std"], optional = true }
 form_urlencoded = { version = "1.2", optional = true }
 http-body-util = { version = "0.1.2", optional = true }
@@ -54,7 +55,7 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.10.6", default-features = false, optional = true }
 quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
-reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "http2"], optional = true }
+reqwest = { version = "0.12", default-features = false, features = ["http2", "rustls-tls-webpki-roots-no-provider"], optional = true }
 ring = { version = "0.17", default-features = false, features = ["std"], optional = true }
 rustls-pki-types = { version = "1.9", default-features = false, features = ["std"], optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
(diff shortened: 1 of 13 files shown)


Cargo.toml
```diff
@@ -54,7 +55,7 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.10.6", default-features = false, optional = true }
 quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
-reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "http2"], optional = true }
+reqwest = { version = "0.12", default-features = false, features = ["http2", "rustls-tls-webpki-roots-no-provider"], optional = true }
```
tustvold, line 58: This is a breaking change that will likely break most clients

Cargo.toml
```diff
@@ -71,14 +72,20 @@ wasm-bindgen-futures = "0.4.18"
...
-http = ["cloud"]
+cloud = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand", "http-body-util", "form_urlencoded", "serde_urlencoded"]
+azure = ["cloud", "httparse", "ring"]
+gcp = ["cloud", "rustls-pki-types", "ring"]
+aws = ["cloud", "md-5", "ring"]
+http = ["cloud", "ring"]
+azure-aws-lc = ["cloud", "httparse", "aws-lc"]
+gcp-aws-lc = ["cloud", "rustls-pki-types", "aws-lc"]
+aws-aws-lc = ["cloud", "md-5", "aws-lc"]
+http-aws-lc = ["cloud", "aws-lc"]
```
tustvold, line 84: I'm not a fan of this explosion of feature flags, it will make testing really complicated

src/crypto_backend.rs
```diff
@@ -0,0 +1,58 @@
...
+///
+/// Since aws-lc-rs is API-compatible with ring for most operations, we can use
+/// a simple alias pattern to switch between them at compile time.
+#[cfg(feature = "ring")]
+#[allow(unused_imports)]
+pub(crate) use ring as crypto;
+
+#[cfg(feature = "aws-lc")]
+#[allow(unused_imports)]
+pub(crate) use aws_lc_rs as crypto;
```
tustvold, line 36: I think this formulation ends up making the feature flags mutually exclusive which goes against best practice

Cargo.toml
faysou: It doesn't, it allows users to choose their own implementation of rustls instead of forcing them to use native tls which is harder to compile as it requires openssl.

Cargo.toml
faysou: The aws-lc flags are not necessary, it just avoids to redefine dependencies for users who want to use aws-lc. Another possibility is to use ring in the default feature.

src/crypto_backend.rs
faysou: Then what should be done ?

Cargo.toml
tustvold: native-roots doesn't depend on openssl, it concerns where it sources the trust store.

src/crypto_backend.rs
tustvold: https://github.com/apache/arrow-rs-object-store/issues/413#issuecomment-3160703614

Cargo.toml
faysou: From my past experience using native tls required openssl, then I'm not an expert on this. But using rustls no provider allows more control on which rustls crypto provider is used.

tustvold: https://crates.io/crates/rustls-native-certs/0.8.2/dependencies

PR #580, opened 2025-12-16
correctly expose HttpError through RetryError::source

### LINKED ISSUE #579: RequestError exposes underlying reqwest::Error rather than the nicer HttpError
**Describe the bug**
If an `HttpError` occurs in a retry context, the error is mapped into a `retry::RequestError`, which is not public to this crate. However, when you try to retrieve the underlying `HttpError` via recursively calling `source`, you instead receive the lower-level `reqwest::Error`. This makes handling HttpError types consistently difficult for consumers who may not want to know about `reqwest`.

**To Reproduce**
Receive an HttpError in a retry context. Attempt to retrieve the underlying HttpError using something like:

```rust
        let mut err: &dyn Error = &e;
        let mut found = false;
        while let Some(source) = err.source() {
            err = source;
            if let Some(err) = err.downcast_ref::<HttpError>() {
                found = true;
                assert_eq!(err.kind(), HttpErrorKind::Request);
            }
        }
        assert!(found, "HttpError not found in source chain");
```

**Expected behavior**
I get an HttpError

**Additional context**

carlsverre:

Closes #579

# Rationale for this change

Consistently expose HttpError in the source chain for all http errors.
 

The issue is caused by `#[error(transparent)]`, which forwards the source method to the underlying error's source impl. The intention is that `transparent` results in the underlying error appearing to be embedded transparently within the embedding error. This means a user can't retrieve the `HttpError` via reflection.

I also added a test to verify that you can retrieve the HttpError in the source chain.

`source` will start returning HttpError before `reqwest::Error` when executed recursively. Hopefully, users recursively search for errors. This will break users who expect `reqwest::Error` to be at a fixed number of steps below the root error type.

### DIFF
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -740,8 +740,9 @@ mod tests {
         while let Some(source) = err.source() {
             err = source;
             if let Some(http_err) = err.downcast_ref::<HttpError>() {
-                found = true;
                 assert_eq!(http_err.kind(), HttpErrorKind::Request);
+                found = true;
+                break;
             }
         }
         assert!(found, "HttpError not found in source chain");


src/client/retry.rs
```diff
@@ -122,7 +122,10 @@ pub enum RequestError {
     #[error("Server returned error response: {body}")]
     Response { status: StatusCode, body: String },
 
-    #[error(transparent)]
+    // We need to use `{0}` here rather than `error(transparent)` to ensure that
+    // `Error::source` returns `HttpError` rather than the underlying
+    // `reqwest::Error` which would happen with `error(transparent)`.
+    #[error("{0}")]
```
alamb, line 128: I found this change confusing given the docs on https://docs.rs/thiserror/latest/thiserror/#derives

> Errors may use error(transparent) to forward the source and [Display](https://doc.rust-lang.org/core/fmt/trait.Display.html) methods straight through to an underlying error without adding an additional message. This would be appropriate for enums that need an “anything else” variant.

Specifically the documentation mostly discusses the display implications.

However, I verified that the newly added test fails without this change:


```
thread 'client::retry::tests::test_retry' (1398936) panicked at src/client/retry.rs:745:9:
HttpError not found in source chain
```

So in summary nice 🕵️ work @carlsverre  (👋  btw -- it is great to see you on github)

carlsverre: Yea, perhaps the thiserror docs could be a bit better. I think it's somewhat of a footgun that `source` is also transparently forwarded, which effectively means the embedded `#[from] Foo` is no longer accessible via recursively searching source.

Nice to see you too!

PR #585, opened 2025-12-20
Pluggable Crypto / Update reqwest 0.13

tustvold:

Closes #462 #576 #413.

# Rationale for this change
 

Adds the ability to compile object_store without a dependency on ring, and instead use a user-provided crypto implementation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,10 @@ jobs:
       - uses: actions/checkout@v6
       - name: Setup Clippy
         run: rustup component add clippy
+      - name: Check no crypto crates
+        run:
+          cargo tree --features gcp-no-crypto,aws-no-crypto,azure-no-crypto,http-no-crypto \
+          | grep -qE '\b(ring|openssl)\b' && { echo "❌ disallowed crate found"; exit 1; } || echo "✅ no disallowed crates"
       # Run different tests for the library on its own as well as
       # all targets to ensure that it still works in the absence of
       # features that might be enabled by dev-dependencies of other
@@ -49,12 +53,20 @@ jobs:
         run: cargo clippy --no-default-features -- -D warnings
       - name: Run clippy with fs features
         run: cargo clippy --no-default-features --features fs -- -D warnings
+      - name: Run clippy with aws-no-crypto feature
+        run: cargo clippy --features aws-no-crypto -- -D warnings
       - name: Run clippy with aws feature
         run: cargo clippy --features aws -- -D warnings
+      - name: Run clippy with gcp-no-crypto feature
+        run: cargo clippy --features gcp-no-crypto -- -D warnings
       - name: Run clippy with gcp feature
         run: cargo clippy --features gcp -- -D warnings
+      - name: Run clippy with azure-no-crypto feature
+        run: cargo clippy --features azure-no-crypto -- -D warnings
       - name: Run clippy with azure feature
         run: cargo clippy --features azure -- -D warnings
+      - name: Run clippy with http-no-crypto feature
+        run: cargo clippy --features http-no-crypto -- -D warnings
       - name: Run clippy with http feature
         run: cargo clippy --features http -- -D warnings
       - name: Run clippy with integration feature
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,7 +212,7 @@ jobs:
         with:
           node-version: 20
       - name: Run wasm32-unknown-unknown tests (via Node)
-        run: wasm-pack test --node --features http --no-default-features
+        run: wasm-pack test --node --features http-no-crypto,ring --no-default-features
 
   windows:
     name: cargo test LocalFileSystem (win64)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
       - name: Install wasm32-wasip1
         run: rustup target add wasm32-wasip1
       - name: Build wasm32-wasip1
-        run: cargo build --all-features --target wasm32-wasip1
+        run: cargo build --features aws-no-crypto,gcp-no-crypto,azure-no-crypto,http-no-crypto,ring --target wasm32-wasip1
       - name: Install wasm-pack
         run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
       - uses: actions/setup-node@v6
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,7 @@ quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], 
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
 reqwest = { version = "0.13", default-features = false, features = ["http2", "rustls-no-provider"], optional = true }
 ring = { version = "0.17", default-features = false, features = ["std"], optional = true }
+aws-lc-rs = { version = "1.15", default-features = false, optional = true }
 rustls-pki-types = { version = "1.9", default-features = false, features = ["std"], optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["std"], optional = true }
@@ -72,8 +73,10 @@ wasm-bindgen-futures = "0.4.18"
 [features]
 default = ["fs"]
 cloud-no-crypto = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand","http-body-util", "form_urlencoded", "serde_urlencoded"]
-cloud = ["ring", "rustls-pki-types", "cloud-no-crypto"]
+cloud = ["aws-lc-rs", "cloud-no-crypto"]
 
+aws-lc-rs = ["dep:aws-lc-rs", "reqwest/rustls", "rustls-pki-types"]
+ring = ["dep:ring", "rustls-pki-types"]
 
 azure-no-crypto = ["cloud-no-crypto", "httparse"]
 azure = ["cloud", "azure-no-crypto"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,13 +54,14 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.10.6", default-features = false, optional = true }
 quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
-reqwest = { version = "0.12", default-features = false, features = ["http2", "rustls-tls-no-provider"], optional = true }
+reqwest = { version = "0.13", default-features = false, features = ["http2", "rustls-no-provider"], optional = true }
 ring = { version = "0.17", default-features = false, features = ["std"], optional = true }
 rustls-pki-types = { version = "1.9", default-features = false, features = ["std"], optional = true }
 serde = { version = "1.0", default-features = false, features = ["derive"], optional = true }
 serde_json = { version = "1.0", default-features = false, features = ["std"], optional = true }
 serde_urlencoded = { version = "0.7", optional = true }
 tokio = { version = "1.29.0", features = ["sync", "macros", "rt", "time", "io-util"] }
+webpki-root-certs = { version = "1", optional = true }
 
 [target.'cfg(target_family="unix")'.dev-dependencies]
 nix = { version = "0.30.0", features = ["fs"] }
@@ -72,7 +73,7 @@ wasm-bindgen-futures = "0.4.18"
 [features]
 default = ["fs"]
 cloud-no-crypto = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand","http-body-util", "form_urlencoded", "serde_urlencoded"]
-cloud = ["ring", "rustls-pki-types", "cloud-no-crypto", "reqwest?/rustls-tls-native-roots"]
+cloud = ["ring", "rustls-pki-types", "cloud-no-crypto"]
 
 
 azure-no-crypto = ["cloud-no-crypto", "httparse"]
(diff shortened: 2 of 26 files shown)


Cargo.toml
```diff
@@ -71,12 +71,24 @@ wasm-bindgen-futures = "0.4.18"
 
 [features]
 default = ["fs"]
-cloud = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand", "ring", "http-body-util", "form_urlencoded", "serde_urlencoded"]
-azure = ["cloud", "httparse"]
+cloud-no-crypto = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand","http-body-util", "form_urlencoded", "serde_urlencoded"]
```
tustvold, line 75: I'm not a massive fan of the feature explosion, but this was the only way to avoid needing to cut a breaking release.

src/client/crypto.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn digest(&self, algorithm: DigestAlgorithm) -> Result<Box<dyn DigestContext>>;
+
+    /// Compute an HMAC with the provided `secret`
+    fn hmac(&self, algorithm: DigestAlgorithm, secret: &[u8]) -> Result<Box<dyn HmacContext>>;
+
+    /// Sign a payload with the provided PEM-encoded secret
+    fn sign(&self, algorithm: SigningAlgorithm, pem: &[u8]) -> Result<Box<dyn Signer>>;
+}
+
+/// Incrementally compute a digest, see [`CryptoProvider::digest`]
```
tustvold, line 48: I opted to allow for incremental computation to somewhat future-proof this API

src/client/crypto.rs
```diff
@@ -0,0 +1,235 @@
...
+pub trait DigestContext: Send {
+    ///Updates the digest with all the data in data.
+    ///
+    /// It is implementation-defined behaviour to call this after calling [`Self::finish`]
+    fn update(&mut self, data: &[u8]);
```
tustvold, line 53: I debated making this method fallible, but decided any error can be returned by `finish`

src/client/crypto.rs
```diff
@@ -0,0 +1,235 @@
...
+    /// Sign the provided payload
+    fn sign(&self, string_to_sign: &[u8]) -> Result<Vec<u8>>;
+}
+
+/// Attempts to find a [`CryptoProvider`]
+///
+/// If `custom` is `Some(v)` returns `v` otherwise returns the compile-time default
+#[cfg(feature = "ring")]
+#[inline]
+pub(crate) fn crypto_provider(custom: Option<&dyn CryptoProvider>) -> Result<&dyn CryptoProvider> {
```
tustvold, line 83: This formulation is a bit funky, but was the best I could come up with that would:

* Avoid needing breaking changes to AwsAuthorizer and AzureAuthorizer
* Allow not specifying a crypto provider if skip_signature is enabled

src/gcp/credential.rs
```diff
@@ -123,44 +104,43 @@ pub struct GcpSigningCredential {
...
-        use rustls_pki_types::pem::PemObject;
-
-        match PrivateKeyDer::from_pem_slice(encoded) {
-            Ok(PrivateKeyDer::Pkcs8(key)) => Self::from_pkcs8(key.secret_pkcs8_der()),
-            Ok(PrivateKeyDer::Pkcs1(key)) => Self::from_der(key.secret_pkcs1_der()),
-            Ok(_) => Err(Error::MissingKey),
-            Err(source) => Err(Error::ReadPem { source }),
-        }
+    #[cfg(feature = "ring")]
+    pub fn from_pem(encoded: &[u8]) -> crate::Result<Self> {
```
tustvold, line 144: This is technically a breaking change, however, the previous error type was not public and therefore there wasn't any way to actually name it. I think we can therefore get away with this change

tustvold (commented):
One aspect that is potentially worth highlighting is that it if skip_signature is false, it should be possible to instantiate without a crypto provider.

The downside is that the absence is not necessarily detected until request time, I debated including a check in build, but in many cases it isn't always immediately obvious what crypto provision is necessary.

src/client/crypto.rs
```diff
@@ -0,0 +1,252 @@
...
+pub trait DigestContext: Send {
+    ///Updates the digest with all the data in data.
+    ///
+    /// It is implementation-defined behaviour to call this after calling [`Self::finish`]
+    fn update(&mut self, data: &[u8]);
+
+    /// Finalizes the digest calculation and returns the digest value.
+    ///
+    /// It is implementation-defined behaviour to call this after calling [`Self::finish`]
+    fn finish(&mut self) -> Result<&[u8]>;
```
tustvold, line 58: I included the fallibility because of openssl - but I'm not sure if any of those errors are actually worth propagating, or if they would imply incorrect usage of openssl...

JakeDern:
@tustvold Tried it out this morning, it worked great for me - Thank you for taking this up!

I tried it with both the built-in ring implementation and with a quick openssl implementation. Here it is in case you would like to include something similar as an example:

```rust
use object_store::client::{
    CryptoProvider, DigestAlgorithm, DigestContext, HmacContext, Signer, SigningAlgorithm,
};
use object_store::path::Path;
use object_store::{ObjectStore, ObjectStoreExt, Result};
use openssl::hash::MessageDigest;
use openssl::pkey::{PKey, Private};
use openssl::sign::Signer as OpenSslSigner;
use std::sync::Arc;

pub enum OpenSslCryptoError {
    #[error("OpenSSL error: {0}")]
    OpenSsl(#[from] openssl::error::ErrorStack),

    #[error("No RSA key found in PEM file")]
    MissingKey,
}

impl From<OpenSslCryptoError> for object_store::Error {
    fn from(value: OpenSslCryptoError) -> Self {
        Self::Generic {
            store: "OpenSslCryptoProvider",
            source: Box::new(value),
        }
    }
}

#[derive(Debug, Default)]
pub struct OpenSslCryptoProvider;

impl CryptoProvider for OpenSslCryptoProvider {
    fn digest(&self, algorithm: DigestAlgorithm) -> Result<Box<dyn DigestContext>> {
        let md = match algorithm {
            DigestAlgorithm::Sha256 => MessageDigest::sha256(),
            _ => unimplemented!("unsupported digest algorithm: {:?}", algorithm),
        };
        let hasher = openssl::hash::Hasher::new(md).map_err(OpenSslCryptoError::from)?;
        Ok(Box::new(OpenSslDigestContext {
            hasher,
            output: Vec::new(),
        }))
    }

    fn hmac(&self, algorithm: DigestAlgorithm, secret: &[u8]) -> Result<Box<dyn HmacContext>> {
        let md = match algorithm {
            DigestAlgorithm::Sha256 => MessageDigest::sha256(),
            _ => unimplemented!("unsupported digest algorithm: {:?}", algorithm),
        };
        let key = PKey::hmac(secret).map_err(OpenSslCryptoError::from)?;
        let signer = OpenSslSigner::new(md, &key).map_err(OpenSslCryptoError::from)?;
        Ok(Box::new(OpenSslHmacContext {
            signer,
            output: Vec::new(),
        }))
    }

    fn sign(&self, algorithm: SigningAlgorithm, pem: &[u8]) -> Result<Box<dyn Signer>> {
        match algorithm {
            SigningAlgorithm::RS256 => {
                let key = PKey::private_key_from_pem(pem).map_err(OpenSslCryptoError::from)?;
                Ok(Box::new(OpenSslRsaSigner { key }))
            }
            _ => unimplemented!("unsupported signing algorithm: {:?}", algorithm),
        }
    }
}

struct OpenSslDigestContext {
    hasher: openssl::hash::Hasher,
    output: Vec<u8>,
}

impl DigestContext for OpenSslDigestContext {
    fn update(&mut self, data: &[u8]) {
        let _ = self.hasher.update(data);
    }

    fn finish(&mut self) -> Result<&[u8]> {
        let digest = self.hasher.finish().map_err(OpenSslCryptoError::from)?;
        self.output = digest.to_vec();
        Ok(&self.output)
    }
}

struct OpenSslHmacContext {
    signer: OpenSslSigner<'static>,
    output: Vec<u8>,
}

impl HmacContext for OpenSslHmacContext {
    fn update(&mut self, data: &[u8]) {
        let _ = self.signer.update(data);
    }

    fn finish(&mut self) -> Result<&[u8]> {
        self.output = self
            .signer
            .sign_to_vec()
            .map_err(OpenSslCryptoError::from)?;
        Ok(&self.output)
    }
}

#[derive(Debug)]
struct OpenSslRsaSigner {
    key: PKey<Private>,
}

impl Signer for OpenSslRsaSigner {
    fn sign(&self, data: &[u8]) -> Result<Vec<u8>> {
        let mut signer = OpenSslSigner::new(MessageDigest::sha256(), &self.key)
            .map_err(OpenSslCryptoError::from)?;
        signer.update(data).map_err(OpenSslCryptoError::from)?;
        let signature = signer.sign_to_vec().map_err(OpenSslCryptoError::from)?;
        Ok(signature)
    }
}

#[tokio::main(flavor = "current_thread")]
async fn main() -> std::result::Result<(), Box<dyn std::error::Error>> {
    let crypto_provider = Arc::new(OpenSslCryptoProvider);

    let azure = MicrosoftAzureBuilder::new()
        .with_account("mystorageaccount")
        .with_access_key("myaccesskey")
        .with_container_name("mycontainer")
        .with_crypto_provider(crypto_provider)
        .build()?;

    // Write a file
    let path = Path::from("example_openssl.txt");
    let data = bytes::Bytes::from("Hello from OpenSSL crypto provider!");
    azure.put(&path, data.into()).await?;

    // Read it back
    let result = azure.get(&path).await?;
    let bytes = result.bytes().await?;
    println!("Content: {}", String::from_utf8_lossy(&bytes));

    Ok(())
}

```

src/lib.rs
```diff
@@ -515,15 +515,43 @@
...
 //!
+//! # Cryptography
+//!
+//! When using the non `*-no-crypto` features, i.e. `aws`, `gcp`, `azure`, etc... this crate makes use
+//! of [`aws-lc-rs`] for cryptography.
+//!
+//! Alternatively if you wish to use [`ring`], e.g. to support WASM targets, you should instead use the
+//! `*-no-crypto` feature flags, e.g. `aws-no-crypto`, and then enable the `ring` feature.
+//!
+//! Note: for TLS to work you will additionally need to register ring as the default rustls cryptography
```
tustvold, line 526:
```suggestion
//! Note: for TLS to work you will may need to register ring as the default rustls cryptography
```

As whether you need this depends on what HTTPConnector you are using.

Cargo.toml
```diff
@@ -71,13 +72,26 @@ wasm-bindgen-futures = "0.4.18"
...
+ring = ["dep:ring", "rustls-pki-types"]
+
+azure-no-crypto = ["cloud-no-crypto", "httparse"]
+azure = ["cloud", "azure-no-crypto"]
+
 fs = ["walkdir"]
-gcp = ["cloud", "rustls-pki-types"]
-aws = ["cloud", "md-5"]
-http = ["cloud"]
-tls-webpki-roots = ["reqwest?/rustls-tls-webpki-roots"]
```
tustvold, line 80: Given this change is now spicy enough that I don't think it should go in a non-breaking release, I opted to take the opportunity to remove this feature flag. People can register the certificates manually if they wish to.

.github/workflows/ci.yml
```diff
@@ -39,6 +39,10 @@ jobs:
       - uses: actions/checkout@v6
       - name: Setup Clippy
         run: rustup component add clippy
+      - name: Check no crypto crates
+        run:
+          cargo tree --features gcp-no-crypto,aws-no-crypto,azure-no-crypto,http-no-crypto \
+          | grep -qE '\b(ring|openssl)\b' && { echo "❌ disallowed crate found"; exit 1; } || echo "✅ no disallowed crates"
```
crepererum, line 45: It seems that cargo-deny should be able to do that, but I DON'T think it can ban dependencies based on features:

https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html

crepererum (approved):
That's elegant 👍 

Could you apply your own GitHub change recommendations (I think there's one doc type/change)?

Maybe we can aim for another breaking release in Q2.

crepererum:
@tustvold do you have time to rebase this PR, since we're now in the 0.14 merge phase and could merge it? (otherwise someone else can pick this work up)

Cargo.toml
```diff
@@ -54,8 +54,9 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.10.6", default-features = false, optional = true }
 quick-xml = { version = "0.38.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.9", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
-reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "http2"], optional = true }
+reqwest = { version = "0.13", default-features = false, features = ["http2", "rustls-no-provider"], optional = true }
```
Tpt, line 57: Having `rustls-no-provider` as an explicit feature requires users to setup explicitly their crypto backend. What about just:
```suggestion
reqwest = { version = "0.13", default-features = false, features = ["http2"], optional = true }
```
 and letting users that want the `reqwest` auto-setup depend on `reqwest` with the default features enabled?

abonander: Since features are supposed to be purely additive, enabling as a few features as necessary is the best approach for a library.

Tpt: Yes! Hence my suggestion to not add `rustls-no-provider` feature here if it's possible

goffrie:
In the interest of possibly moving this along I've rebased this PR & addressed the `rustls-no-provider` comment in #707; hopefully that doesn't step on anyone's toes

Cargo.toml
tustvold: IIRC this brings in aws-lc-rs which people may be wanting to avoid

tustvold:
Sorry I've been absolutely swamped, I'll try to find time to look into this in the next few days. R.e. the no-provider there was a reason for this, I think it was too avoid requiring aws-lc-rs. It should only need additional setup if not using the no-crypto variants.

Cargo.toml
Tpt: @tustvold Indeed, it's what happens if the `default-tls` default feature is enabled. But if I read https://docs.rs/reqwest/latest/reqwest/tls/index.html#default-tls correctly (please correct me if I am wrong), this does not happen if `default-features = false` is used, and it's what is done here.

(please ignore this message if this discussion is not useful, thank you for your work on this crate)

abonander: The suggested change avoids enabling _any_ TLS backend by default. `rustls-no-provider` would still bring in RusTLS itself.

alamb:
I think this has been superceded by 
- https://github.com/apache/arrow-rs-object-store/pull/707

PR #590, opened 2026-01-03
feat: Add support for AWS_ENDPOINT_URL_S3 environment variable

### LINKED ISSUE #589: Support AWS_ENDPOINT_URL_S3 in aws backend

Support S3 specific endpoint env var - AWS_ENDPOINT_URL_S3 (similar to how we support AWS_ENDPOINT_URL_STS).

If AWS_ENDPOINT_URL_S3 is set, use that as S3 endpoint with highest precedence.

For now I am setting AWS_ENDPOINT_URL but this means I can't use separate endpoint for different services.

rajatgoel:

Closes #589

There are 2 changes - 
* Support `AWS_ENDPOINT_URL_S3` as an alternative env var that sets S3 endpoint.
* Process env vars in the same order as official AWS CLI/SDK does.

If this env var is not set, this should be a no-op. But if this env var is set,
our code will start respecting that. Probably worth calling out in the release notes.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -135,6 +135,8 @@ pub struct AmazonS3Builder {
     bucket_name: Option<String>,
     /// Endpoint for communicating with AWS S3
     endpoint: Option<String>,
+    /// Service-specific S3 endpoint URL (takes precedence over endpoint)
+    s3_endpoint: Option<String>,
     /// Token to use for requests
     token: Option<String>,
     /// Url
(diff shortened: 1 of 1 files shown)


rajatgoel:
Yes, this is a standard aws env var - https://docs.aws.amazon.com/sdkref/latest/guide/feature-ss-endpoints.html

src/aws/builder.rs
```diff
@@ -559,13 +565,23 @@ impl AmazonS3Builder {
...
         let mut builder: Self = Default::default();
 
-        for (os_key, os_value) in std::env::vars_os() {
-            if let (Some(key), Some(value)) = (os_key.to_str(), os_value.to_str()) {
-                if key.starts_with("AWS_") {
-                    if let Ok(config_key) = key.to_ascii_lowercase().parse() {
-                        builder = builder.with_config(config_key, value);
-                    }
-                }
+        // Collect and sort environment variables to ensure deterministic precedence.
```
tustvold: Can we instead add a separate config key for this + member variable, and then implement this priority in the build method. Otherwise this behaviour becomes specific to from_env and will break for methods that use with_config directly - e.g. parse_url

rajatgoel: Done. Now we handle it the same way we do AWS_ENDPOINT_URL_STS.

Issue #582, opened 2025-12-18
Release object store `0.12.5` (maintenance) - Target Dec 2025


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
We are in the process of releasing 0.13.0 (a breaking API change):
- https://github.com/apache/arrow-rs-object-store/issues/367

However, this will likely take several months to percolate through the ecosystem (e.g. to get to a DataFusion user it will require a major release of arrow and then a major release of DataFusion). 

There are several fixes that would be nice to release downstream sooner, such as the fix for https://github.com/apache/arrow-rs-object-store/issues/578   requested by @ion-elgreco 

**Describe the solution you'd like**
Let's make a 0.12.5 release and backport specific fixes

Requested content
- [x]  ~https://github.com/apache/arrow-rs-object-store/issues/578 / https://github.com/apache/arrow-rs-object-store/pull/442 from  @ion-elgreco~
- [x] https://github.com/apache/arrow-rs-object-store/pull/595 / @alamb 
- [x] https://github.com/apache/arrow-rs-object-store/pull/591

Process:
* Cherry pick the relevant changes
* Create a PR targeting the [release/0.12](https://github.com/apache/arrow-rs-object-store/tree/release/0.12) branch

Then I will do my normal process of
- [ ] Version + changelog: https://github.com/apache/arrow-rs-object-store/pull/601
- [ ] Test with arrow-rs: https://github.com/apache/arrow-rs/pull/9127
- [ ] Test with DataFusion: https://github.com/apache/datafusion/pull/19718
- [ ] Make release candidate / start vote thread: https://lists.apache.org/thread/h2n0g943o6ooykrl5dvfy4brm05lz2ms
- [ ] Approval and release to crates.io

crepererum:
We may also want to backport https://github.com/apache/arrow-rs-object-store/pull/591 .

ion-elgreco:
> There hasn't been much motion on this one and I don't have time to spend on it myself. I am happy to run the release process  however. 
> 
> If anyone wants to help this along, please create a PR with the relevant content targeting the [release/0.12](https://github.com/apache/arrow-rs-object-store/tree/release/0.12) branch and flag me for review

I'll see if I can manage to find some time this week!

alamb:
The release content is so far
- [x] https://github.com/apache/arrow-rs-object-store/pull/597
- [ ] https://github.com/apache/arrow-rs-object-store/pull/596
[...]

PR #597, opened 2026-01-08
[release/0.12] Fix new CI issues

AdamGS:
- part of https://github.com/apache/arrow-rs-object-store/issues/582

This PR fixes some issues that have crept into this branch since the last minor release.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -56,9 +56,18 @@ jobs:
         # Necessary because tokio 1.30.0 updates MSRV to 1.63
         # and url 2.5.1, updates to 1.67
         run: |
+          cargo update -p parking_lot --precise 0.12.4
+          cargo update -p parking_lot_core --precise 0.9.11
+          cargo update -p lock_api --precise 0.4.13
+          cargo update -p quote --precise 1.0.41
+          cargo update -p proc-macro2 --precise 1.0.103
+          cargo update -p itoa --precise 1.0.15
+          cargo update -p syn --precise 2.0.106
+          cargo update -p tokio-util --precise 0.7.17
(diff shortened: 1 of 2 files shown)


.github/workflows/rust.yml
```diff
@@ -56,9 +56,18 @@ jobs:
         # Necessary because tokio 1.30.0 updates MSRV to 1.63
         # and url 2.5.1, updates to 1.67
         run: |
+          cargo update -p parking_lot --precise 0.12.4
```
AdamGS, line 59: A lot of crates just recently made the jump from `1.64` to `1.68`, `parking_lot` and tokio moved to `1.71` and even `tracing` made a very significant change to `1.65`.

src/client/retry.rs
```diff
@@ -844,6 +844,8 @@ mod tests {
             // Reset the connection on the first n-1 attempts
             for _ in 0..retry.max_retries {
                 let (stream, _) = listener.accept().await.unwrap();
+                // TcpStream::set_linger is deprecated but this use case is valid to reset the stream
+                #[allow(deprecated)]
                 stream.set_linger(Some(Duration::from_secs(0))).unwrap();
```
AdamGS, line 849: My understanding is that this is more about it being a potential performance issue than anything else, with the thread being potentially blocked on drop.
This specific options will also be now available in new tokio versions as `TcpStream::et_linger_zero`.

alamb: Indeed -- @crepererum fixed the same issue in basically the same way on main in
- https://github.com/apache/arrow-rs-object-store/pull/592

alamb (approved):
Thank you @AdamGS

PR #604, opened 2026-01-14
Fix Azure URL parsing

### LINKED ISSUE #443: az:// <container> not work as expected

tustvold:

- Closes #443

# Rationale for this change
 

https://github.com/apache/arrow-rs-object-store/pull/399 introduced a breaking change to the way that we parse `az://` URLs. This IMO reflects a more sensible URL encoding, but this doesn't match the URL encoding expected by the rest of this crate, which in turn is borrowed from fsspec.

Reverts the URL parsing change in #399

Technically this is a bug fix, but might make sense to wait for a breaking release.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -52,7 +52,7 @@ impl From<Error> for super::Error {
 /// - `memory:///` -> [`InMemory`]
 /// - `s3://bucket/path` -> [`AmazonS3`](crate::aws::AmazonS3) (also supports `s3a`)
 /// - `gs://bucket/path` -> [`GoogleCloudStorage`](crate::gcp::GoogleCloudStorage)
-/// - `az://account/container/path` -> [`MicrosoftAzure`](crate::azure::MicrosoftAzure) (also supports `adl`, `azure`, `abfs`, `abfss`)
+/// - `[az|abfs[s]]://container[@<account>.<host>]/path` -> [`MicrosoftAzure`](crate::azure::MicrosoftAzure)
 /// - `http://mydomain/path` -> [`HttpStore`](crate::http::HttpStore)
 /// - `https://mydomain/path` -> [`HttpStore`](crate::http::HttpStore)
 ///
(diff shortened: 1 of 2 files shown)


src/parse.rs
```diff
@@ -52,7 +52,7 @@ impl From<Error> for super::Error {
 /// - `memory:///` -> [`InMemory`]
 /// - `s3://bucket/path` -> [`AmazonS3`](crate::aws::AmazonS3) (also supports `s3a`)
 /// - `gs://bucket/path` -> [`GoogleCloudStorage`](crate::gcp::GoogleCloudStorage)
-/// - `az://account/container/path` -> [`MicrosoftAzure`](crate::azure::MicrosoftAzure) (also supports `adl`, `azure`, `abfs`, `abfss`)
+/// - `[az|abfs[s]]://container[@<account>.<host>]/path` -> [`MicrosoftAzure`](crate::azure::MicrosoftAzure)
```
tustvold, line 55: I opted to drop `azure` as it is a custom format AFAICT, and `adl` as it is deprecated by azure.

Whilst still supported for compatibility, people shouldn't be using them.

src/parse.rs
```diff
@@ -299,11 +300,35 @@ mod tests {
                 (ObjectStoreScheme::MicrosoftAzure, "path"),
             ),
             (
-                "az://account/container",
-                (ObjectStoreScheme::MicrosoftAzure, ""),
+                "az://container/path",
+                (ObjectStoreScheme::MicrosoftAzure, "path"),
```
tustvold, line 304: This now matches https://docs.rs/object_store/0.12.4/object_store/azure/struct.MicrosoftAzureBuilder.html#method.with_url

src/parse.rs
alamb: to be clear, you stopped documenting that `azure` is supported, but the code still supports it

src/parse.rs
```diff
@@ -299,11 +300,35 @@ mod tests {
...
+            (
+                "abfss://container@account/path",
+                (ObjectStoreScheme::MicrosoftAzure, "path"),
+            ),
+            (
+                "adl://container/path",
+                (ObjectStoreScheme::MicrosoftAzure, "path"),
             ),
             (
-                "az://account/container/path",
```
alamb, line 306: Can we please keep this test so it is clearer what is the difference in behavior? I am not sure what this URL will parse to now

kylebarron (approved):
This seems fine to me; I apologize for the breakage

src/parse.rs
tustvold: It is here https://github.com/apache/arrow-rs-object-store/pull/604#discussion_r2691944687

Just renamed

PR #607, opened 2026-01-15
feat: add redirection handling

### LINKED ISSUE #603: redirection doesn't work in put request
**Describe the bug**

Hi there.

I'm trying to use the object store with Nvidia Aistore, which uses an aws s3 backend.

I've set up aistore with 1 gateway and 1 target (storage instance).

```rust
#[tokio::main]
async fn main() {
    fmt()
        .with_env_filter(
            EnvFilter::try_from_default_env()
                .unwrap_or_else(|_| EnvFilter::new("info,object_store=trace")),
        )
        .init();

    let client = object_store::aws::AmazonS3Builder::from_env()
        .with_allow_http(true)
        .build()
        .unwrap();

    let response = client
        .get(&object_store::path::Path::from_url_path("test").unwrap())
        .await
        .unwrap(); # OK, this file is located in s3. Redirect is also used to get it.

    tracing::info!("get response: {:?}", response);

    client
        .put(
            &object_store::path::Path::from_url_path("test").unwrap(),
            PutPayload::new(),
        )
        .await
        .unwrap();
}
```

Error/Logs:

```sh
2026-01-14T09:53:23.862399Z DEBUG object_store::aws::builder: Using Static credential provider
2026-01-14T09:53:23.898166Z  INFO example: get response: GetResult { payload: GetResultPayload(Stream), meta: ObjectMeta { location: Path { raw: "test" }, last_modified: 2026-01-14T09:46:34Z, size: 0, e_tag: None, version: None }, range: 0..0, attributes: Attributes({ContentType: AttributeValue("application/octet-stream")}) }

thread 'main' panicked at src/main.rs:38:10:
called `Result::unwrap()` on an `Err` value: Generic { store: "S3", source: RetryError(RetryErrorImpl { method: PUT, uri: Some(http://192.168.103.92:8080/s3/oo/test), retries: 0, max_retries: 10, elapsed: 415.533µs, retry_timeout: 180s, inner: Status { status: 307, body: None } }) }
```

**To Reproduce**

1. Setup Aistore: https://github.com/NVIDIA/aistore?tab=readme-ov-file#quick-start

2. Run the code shown above

My env:

```
export AWS_ENDPOINT=http://my-domain:8080/s3
export AWS_REGION=us-east-1
export AWS_ACCESS_KEY_ID=FAKEKEY
export AWS_SECRET_ACCESS_KEY=FAKESECRET
export AWS_BUCKET=oo
```

**Expected behavior**

Put is working successfully.

**Additional context**

```toml
object_store = { version = "0.13.0", features = ["aws"] }
```

if0ne:

Closes #603 

# Rationale for this change
 

The object store’s address can be hidden behind a gateway that forwards requests. The currently used `reqwest` crate is not yet able to handle redirect responses when the request body is a stream.

* Added `RedirectContext` as an argument to the `send` method of the `RetryableRequest` struct

* Added `RedirectConfig` for configuring the maximum number of redirects

### DIFF
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -29,7 +29,8 @@ use crate::multipart::PartId;
 use crate::util::{GetRange, deserialize_rfc1123};
 use crate::{
     Attribute, Attributes, ClientOptions, GetOptions, ListResult, ObjectMeta, Path, PutMode,
-    PutMultipartOptions, PutOptions, PutPayload, PutResult, Result, RetryConfig, TagSet,
+    PutMultipartOptions, PutOptions, PutPayload, PutResult, RedirectConfig, Result, RetryConfig,
+    TagSet,
 };
 use async_trait::async_trait;
 use base64::Engine;
@@ -272,7 +273,7 @@ impl PutRequest<'_> {
             .sensitive(sensitive)
             .idempotent(self.idempotent)
             .payload(Some(self.payload))
-            .send()
+            .send(&RedirectConfig::default())
             .await
             .map_err(|source| {
                 let path = self.path.as_ref().into();
@@ -736,7 +737,7 @@ impl AzureClient {
             .retryable(&self.config.retry_config)
             .sensitive(sensitive)
             .idempotent(overwrite)
-            .send()
+            .send(&RedirectConfig::default())
             .await
             .map_err(|err| err.error(STORE, from.to_string()))?;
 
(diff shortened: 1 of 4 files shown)


tustvold:
> Yes, this will break authentication.

This feels like Nvidia AIStore is doing something that fundamentally isn't S3-Compatible then.

> I’m not sure, but as far as I know, this auth header-handling logic is implemented in reqwest.

reqwest just sends whatever headers it is given, if this Nvidia AIStore thing is relying on redirects (which is dubious to then claim is S3-compatible - see above), its unclear whether the redirected request should then contain the auth header or not.

I'm NGL I am leaning to it being better for you to just implement a custom ObjectStore for this, fundamentally this product doesn't really appear to be S3-compatible...

Edit: The docs for the product seem to describe how boto3 requires patching to make it work - https://aistore.nvidia.com/docs/s3compat#further-reading...

PR #608, opened 2026-01-15
Upgrade reqwest to 0.13

AdamGS:
This PR upgrades the `reqwest` dependency from 0.12 to `0.13`. This changes introduces a few breaking changes in reqwest that I've tried to mitigate here.

1. `reqwest` does not provide features to enable different root cert bundles. It will verify certificates using `rustls-platform-verifier`, which will use the OS native certificate store. This is why the `reqwest/rustls-tls-native-roots` feature was replaced with the `reqwest/rustls` one.
2. For `webpki`, I've kept the feature to maintain compatibility, replacing the `reqwest` feature with pulling the `webpki` certs directly and adding them to the client.
3. `reqwest` has now moved to use `aws-lc` as its underlying crypto provider. Its technically possible to override that but it seems like its technically possible to install a `ring` based one, I'm not sure if that's actually desired. I've also considered removing the `ring` based code so `ring` can be completely removed as a dependency, but I figured I'll try and see if anyone has a more informed opinion before I do.

This PR should have no external impact, but I would like to still try and test it against a real server.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,16 +184,16 @@ jobs:
       - uses: actions/checkout@v6
         with:
           submodules: true
-      - name: Install clang (needed for ring)
+      - name: Install clang (needed for aws-lc-rs)
         run: apt-get update && apt-get install -y clang
       - name: Install wasm32-unknown-unknown
         run: rustup target add wasm32-unknown-unknown
       - name: Build wasm32-unknown-unknown
-        run: cargo build --target wasm32-unknown-unknown
+        run: cargo build --target wasm32-unknown-unknown --features http
       - name: Install wasm32-wasip1
         run: rustup target add wasm32-wasip1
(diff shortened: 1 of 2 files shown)


src/client/mod.rs
```diff
@@ -762,8 +762,19 @@ impl ClientOptions {
             builder = builder.proxy(proxy);
         }
 
-        for certificate in &self.root_certificates {
-            builder = builder.add_root_certificate(certificate.0.clone());
+        builder = builder.tls_certs_merge(
+            self.root_certificates
```
AdamGS, line 766: I moved away from `add_root_certificate` because it now has a note that saying its deprecated

AdamGS:
I've pushed the code replacing ring with aws-lc-rs, LMK if you want to keep it or should I drop it. Its a pretty easy change, the API seems to be almost identical.

PR #609, opened 2026-01-16
Fix EOF retry bug that could cause data duplication when Range header is ignored

sarum90:

N/A - no existing issue. Happy to create one if preferred.

# Rationale for this change

If a reqwest timeout occurs after consuming all bytes from an S3 response body but before receiving the EOF signal, `retry_stream` attempts to retry with a `Range` header like `bytes=5-4` (where 5 is the file size). This is an invalid/backwards range.

Per RFC 7233, servers MUST ignore syntactically invalid Range headers. S3 follows this spec and returns `200 OK` with the **full file** instead of `206 Partial Content` or `416 Range Not Satisfiable`.

Without validation, `retry_stream` would read the full file again, causing data duplication (e.g., `"hellohello"` instead of `"hello"`).

1. **Early EOF check**: If `range.start >= range.end` before retrying, return EOF immediately rather than sending an invalid range request.

2. **206 status validation**: Verify the retry response is `206 Partial Content`. If the server returns `200 OK` (indicating it ignored our Range header), fail the retry and return the original error.

3. **Content-Range validation**: Verify the `Content-Range` header matches the requested range. This catches cases where the server returns a different range than requested.

4. **Warning logs**: Added `warn!` logs for the new validation failures to aid debugging.

No breaking changes. Users may see different error behavior in edge cases where retries were previously succeeding incorrectly (by silently duplicating data).

# Test Coverage

Automated test coverage for this specific bug is limited for two reasons:

1. **Requires real S3**: LocalStack returns `416 Range Not Satisfiable` for invalid ranges, which is actually non-compliant with RFC 7233. Real S3 correctly ignores the invalid range and returns `200 OK` with the full file, which is what triggers the bug.

2. **Hard to trigger**: The bug requires a timeout (or other error) from reqwest *after* all data is read but *before* EOF is signaled. This is timing-dependent and would require integration tests with long sleeps to force a timeout.

A manual reproduction script is available in this gist: https://gist.github.com/sarum90/cb95633750390db690ef701f08aa20ed

The script demonstrates the bug against real S3 and shows the data duplication.

### DIFF
--- a/src/client/get.rs
+++ b/src/client/get.rs
@@ -518,17 +518,22 @@ mod tests {
 }
 #[cfg(all(test, feature = "http", not(target_arch = "wasm32")))]
 mod http_tests {
+    use super::{GetClient, GetClientExt};
+    use crate::client::header::HeaderConfig;
     use crate::client::mock_server::MockServer;
-    use crate::client::{HttpError, HttpErrorKind, HttpResponseBody};
+    use crate::client::retry::{RetryContext, RetryExt};
+    use crate::client::{GetOptionsExt, HttpClient, HttpError, HttpErrorKind, HttpResponseBody};
     use crate::http::HttpBuilder;
     use crate::path::Path;
-    use crate::{ClientOptions, ObjectStoreExt, RetryConfig};
+    use crate::{ClientOptions, GetOptions, ObjectStoreExt, Result, RetryConfig};
+    use async_trait::async_trait;
     use bytes::Bytes;
     use futures::FutureExt;
     use http::header::{CONTENT_LENGTH, CONTENT_RANGE, ETAG, RANGE};
-    use http::{Response, StatusCode};
+    use http::{Method, Response, StatusCode};
     use hyper::body::Frame;
     use std::pin::Pin;
+    use std::sync::Arc;
     use std::task::{Context, Poll, ready};
     use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


tustvold (commented):
Fun bug, well done for tracking this one down. Perhaps we could use the HttpServer machinery used for testing retries to test this?

sarum90:
Thanks for the suggestion! I initially tried using `HttpBuilder` directly, but couldn't get the 206 validation test to fail on main - turns out the HTTP client has defensive checks that the S3 client lacks. So I created a minimal `TestGetClient` structured like the cloud store clients, which also gave me a way to inject client-side errors after content-length bytes (something you can't trigger from the server side without actual timeouts).

I've added tests for:
1. **206 status validation** - verifies we reject retries that return 200 instead of 206
2. **Content-Range validation** - verifies we reject retries with mismatched Content-Range headers
3. **No retry after complete delivery** - injects an error after all Content-Length bytes are received, verifying we don't attempt a spurious retry

Let me know if the `TestGetClient` + `ErrorAfterBody` infrastructure seems like overkill - happy to simplify if there's a lighter-weight approach, or skip some test coverage if we'd rather avoid potentially brittle test complexity.

dentiny:
Hi I met the same issue in our production environment, but instead of returning error directly for unmatched range header, I tried to copy the useful bytes out if applicable.
I made a PR here cause it's a different approach and this PR doesn't seem to be active for >2months, sadly: https://github.com/apache/arrow-rs-object-store/pull/690
Feel free to close my PR if maintainers think it duplicate, thanks!

sarum90:
Yea please feel free to take this over / do a different approach / etc.

Bug is real, probably several different ways to fix.

We wound up having to move away from object store for other reasons, so unlikely I'll have any driving force to fix CI errors + actually try to get this merged.

dentiny:
> Yea please feel free to take this over / do a different approach / etc.

Thank you for the kind reply!

blaginin:
<details>
    <summary>In case it helps, here is how different clouds respond to various `Range` headers</summary>

```
no `Content-Range`
	 s3
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

	 tigris
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

	 r2
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

	 google cloud
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

===
bytes=100-200
	 s3
		Status code: 206
		Content-Range: bytes 100-200/80835272
		Content-Length: 101

	 tigris
		Status code: 206
		Content-Range: bytes 100-200/80835272
		Content-Length: 101

	 r2
		Status code: 206
		Content-Range: bytes 100-200/80835272
		Content-Length: 101

	 google cloud
		Status code: 206
		Content-Range: bytes 100-200/80835272
		Content-Length: 101

===
bytes=0-(max+1)
	 s3
		Status code: 206
		Content-Range: bytes 0-80835271/80835272
		Content-Length: 80835272

	 tigris
		Status code: 206
		Content-Range: bytes 0-80835271/80835272
		Content-Length: 80835272

	 r2
		Status code: 206
		Content-Range: bytes 0-80835271/80835272
		Content-Length: 80835272

	 google cloud
		Status code: 206
		Content-Range: bytes 0-80835271/80835272
		Content-Length: 80835272

===
bytes=-200
	 s3
		Status code: 206
		Content-Range: bytes 80835072-80835271/80835272
		Content-Length: 200

	 tigris
		Status code: 206
		Content-Range: bytes 80835072-80835271/80835272
		Content-Length: 200

	 r2
		Status code: 206
		Content-Range: bytes 80835072-80835271/80835272
		Content-Length: 200

	 google cloud
		Status code: 206
		Content-Range: bytes 80835072-80835271/80835272
		Content-Length: 200

===
bytes=100-
	 s3
		Status code: 206
		Content-Range: bytes 100-80835271/80835272
		Content-Length: 80835172

	 tigris
		Status code: 206
		Content-Range: bytes 100-80835271/80835272
		Content-Length: 80835172

	 r2
		Status code: 206
		Content-Range: bytes 100-80835271/80835272
		Content-Length: 80835172

	 google cloud
		Status code: 206
		Content-Range: bytes 100-80835271/80835272
		Content-Length: 80835172

===
bytes=(max + 100)-
	 s3
		Status code: 416
		Content-Range: None
		Content-Length: None

	 tigris
		Status code: 416
		Content-Range: None
		Content-Length: 255
		Content: `<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRange</Code><Message>Invalid range</Message><Resource>/foo-compact.vortex</Resource><RequestId>1776438374761678270</RequestId><Key>foo-compact.vortex</Key><BucketName>dima-dev</BucketName></Error>`

	 r2
		Status code: 416
		Content-Range: None
		Content-Length: 27150

	 google cloud
		Status code: 416
		Content-Range: None
		Content-Length: 171
		Content: `<?xml version='1.0' encoding='UTF-8'?><Error><Code>InvalidRange</Code><Message>The requested range cannot be satisfied.</Message><Details>bytes=80835373-</Details></Error>`

===
bytes=5-1
	 s3
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

	 tigris
		Status code: 416
		Content-Range: None
		Content-Length: 255
		Content: `<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRange</Code><Message>Invalid range</Message><Resource>/foo-compact.vortex</Resource><RequestId>1776438375764341629</RequestId><Key>foo-compact.vortex</Key><BucketName>dima-dev</BucketName></Error>`

	 r2
		Status code: 400
		Content-Range: None
		Content-Length: 17949

	 google cloud
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

===
bytes=???
	 s3
		Status code: 200
		Content-Range: None
		Content-Length: 80835272

	 tigris
		Status code: 416
		Content-Range: None
		Content-Length: 255
		Content: `<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRange</Code><Message>Invalid range</Message><Resource>/foo-compact.vortex</Resource><RequestId>1776438377339542727</RequestId><Key>foo-compact.vortex</Key><BucketName>dima-dev</BucketName></Error>`

	 r2
		Status code: 400
		Content-Range: None
		Content-Length: 17949

	 google cloud
		Status code: 200
		Content-Range: None
		Content-Length: 80835272
```

</details>

sarum90:
given #690, just going to close this

PR #619, opened 2026-01-23
Implement tests for range and partial content responses

vitoordaz:
Added tests for handling partial content responses and retries for incomplete response bodies.

I didn't file an issue, because this PR only adds unit tests.

# Rationale for this change
 
This improves code coverage and effectively documents how client behaves in edge case situation.

Only unit tests for GetClient

No

### DIFF
--- a/src/client/get.rs
+++ b/src/client/get.rs
@@ -592,7 +592,7 @@ mod http_tests {
 
         let b = store.get_range(&path, 1..5).await.unwrap();
         assert_eq!(b.as_ref(), b"ello");
-        
+
         // NOTE: if debug_assertions is true, hyper panics with response content length header
         // value does not match the length of response body.
         if !cfg!(debug_assertions) {
@@ -638,7 +638,10 @@ mod http_tests {
             );
 
             mock.push_fn(|req| {
-                assert_eq!(req.headers().get(RANGE).unwrap().to_str().unwrap(), "bytes=8-10");
+                assert_eq!(
+                    req.headers().get(RANGE).unwrap().to_str().unwrap(),
+                    "bytes=8-10"
+                );
 
                 Response::builder()
                     .status(StatusCode::PARTIAL_CONTENT)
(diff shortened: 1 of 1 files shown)


src/client/get.rs
```diff
@@ -579,6 +579,83 @@ mod http_tests {
...
+            .body("ello".to_string())
+            .unwrap();
+
+        mock.push(resp);
+
+        let b = store.get_range(&path, 1..5).await.unwrap();
+        assert_eq!(b.as_ref(), b"ello");
+
+        // NOTE: if debug_assertions is true, hyper panics with response content length header
+        // value does not match the length of response body.
```
alamb, line 597: Does this mean hyper would panic in a real world case? If so, is there a ticket tracking it?

 Or does this mean that this particular response is not valid (and thus we shouldn't be testing it)?

alamb (commented):
Thanks @vitoordaz  -- I took a quick look and it seems like a nice addition to me. I have one test question, otherwise I think this is ready to go

alamb:
I think if you can merge up from main the CI tests would also pass cleanly

src/client/get.rs
vitoordaz: hyper panics in test environment which usually have `debug_assertions = true`. In prod `debug_assertions = false` and hyper returns regular response.

I'm wondering if we can update CI to run all tests with `debug_assertions = false` as well?

alamb: > I'm wondering if we can update CI to run all tests with debug_assertions = false as well?

We could add a new test run with that enabled (or perhaps we could run the tests in `--release` mode)?

alamb:
(please just merge up from main so we can get a clean CI run)

src/client/get.rs
vitoordaz: yes, I think that this will work. I was using this command to run this test

```
RUSTFLAGS="-C debug-assertions=false" RUST_BACKTRACE=full cargo test --features http client::get::http_tests::test_stream_retry
```

alamb: Perhaps you can do so as a follow on PR

PR #621, opened 2026-01-24
Avoid metadata lookup for `LocalFileSystem::read_ranges` and `chunked_stream`

### LINKED ISSUE #614: Avoid reading metadata for `LocalFileSystem::read_ranges` (and other methods)

See https://github.com/apache/arrow-rs-object-store/pull/595#issuecomment-3789074698

- open the file
- try getting bytes from it, if succeeds: it was fine (and we avoided reading the metadata), if it fails, we can read the metadata and "reproduce" the original error

**Additional context**

Dandandan:

- Closes #614

# Rationale for this change

This is unnecessary overhead.
 

Only error once the file operation errors.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,8 +139,10 @@ jobs:
       - name: Configure Azurite (Azure emulation)
         # the magical connection string is from
         # https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#http-connection-strings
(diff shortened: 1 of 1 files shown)


alamb:
@Dandandan  -- here is a proposal to simplify the error handling (I think it is somewhat tricky with all the corner cases, so refactoring into some functions makes it easier to see what is happening I think, and convince myself it is correct)
- https://github.com/Dandandan/arrow-rs-object-store/pull/1

PR #623, opened 2026-01-25
Azure ADLS list_with_offset support

### LINKED ISSUE #461: Use reconstructed ListBlobs marker to provide list offset support in `MicrosoftAzure` store
BlobStorage currently does not expose a way to start at a specific location on ListBlobs requests, this makes implementing _GC-like_ routines that lists millions of objects and have to outlive a single process very cumbersome. 

There is a workaround for this limitation which involves reconstructing the `marker` token used for list pagination to specify a start key where listing should begin.

There are no efficient alternatives given that ListBlobs only supports prefix filtering.

omar:

Closes #461.

# Rationale for this change
 
Implements `list_with_offset` support for Azure/ADLS. See #461 for more details.

Besides adding support for `list_with_offset` for Azure, I've also added some tests to codify that `list_with_offset` always assumed the response did not include the offset value provided.

Previously, this was implied because S3/GCP `list_with_offset` operations were exclusive. However, Azure's implementation is inclusive (i.e. it returns the offset value if found).

The Azure implementation handles the exclusivity client side by removing the offset key in the response before handing it off to the crate's internal structs.

No user-facing changes for non-Azure users.

For Azure users, `list_from_offset` behaves differently now.

- Before:  `list_from_offset` defaulted to the underlying `list_from` method which returned everything ignoring `offset` value. 
- After: `list_from_offset` now uses a `startFrom` parameter set to the value of the offset. See https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?view=rest-storageservices-datalakestoragegen2-2019-12-12&tabs=microsoft-entra-id#uri-parameters

### DIFF
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -130,6 +130,8 @@ impl ObjectStore for MicrosoftAzure {
         if self.client.config().is_emulator {
             // Azurite doesn't support the startFrom query parameter,
             // fall back to client-side filtering
+            //
+            // See https://github.com/Azure/Azurite/issues/2619#issuecomment-3660701055
             let offset = offset.clone();
             self.list(prefix)
                 .try_filter(move |f| futures::future::ready(f.location > offset))


omar:
There is no bug. The existing behavior of object_store expects a fully qualified path for `offset` as that's the current behavior of S3 and GCP.

This is ready to be merged.

src/azure/mod.rs
```diff
@@ -121,6 +122,23 @@ impl ObjectStore for MicrosoftAzure {
...
     }
 
+    fn list_with_offset(
+        &self,
+        prefix: Option<&Path>,
+        offset: &Path,
+    ) -> BoxStream<'static, Result<ObjectMeta>> {
+        if self.client.config().is_emulator {
+            // Azurite doesn't support the startFrom query parameter,
+            // fall back to client-side filtering
```
crepererum, line 132:
```suggestion
            // fall back to client-side filtering
            //
            // See https://github.com/Azure/Azurite/issues/2619#issuecomment-3660701055
```

Let's leave some breadcrumbs so that one day we can remove this exception.

PR #625, opened 2026-01-26
fix(wasm): avoid std::time::Instant::now() and expand testing

### LINKED ISSUE #624: Crash in http in WASM due to hitting std::time::Instant::now()
**Describe the bug**
Calling `std::time::Instant::now()` crashes unconditionally on WASM. retry.rs uses tokio::time::sleep() in the retry case, so you may think things are working but see intermittent crashes.

**To Reproduce**
Any http request that is retried in WASM should display the problem.

**Expected behavior**
No crash.

**Additional context**
See for example https://github.com/apache/arrow-rs-object-store/blob/main/src/client/retry.rs#L389

pmarks:
Proposed fix to #624 - opening for discussion, I'm happy to rework this.

Currently retry and throttle can crash on WASM due to hitting std::time::Instant::now(), and MultiPart upload relies on tokio::task::JoinSet, which doesn't work on WASM.

TODO: figure out a unit test that actually exercises `RetryableRequest` in WASM. Existing tests can't run because they need MockServer, which needs TcpListener.

Closes #624 

# Rationale for this change
 Solidify WASM support by fixing a couple small issues and greatly expanding WASM test coverage.
 
- Use n0-futures, which re-exports tokio types on native, and supplies browser-compatible versions on wasm32.  Used to get a replacement for tokio::task::JoinSet, used for MultiPart uploads
- Use sleep() from gloo-timers on wasm instead of tokio::time::sleep which relies on std::Instant::now()
- To test WASM, we use #[wasm_bindgen_test] instead of #[tokio::test] to mark many existing tests.  We can conditionally import the appropriate macro, renaming the macro as `async_test` then decorate the tests with #[async_test]. These tests will now run under the existing WASM CI.

The main Error enum will change for the WASM build - it will now expose `n0_future::task::JoinError` instead of `tokio::task::JoinError` - this is a breaking change.


src/util.rs
```diff
@@ -326,6 +327,23 @@ pub(crate) fn hex_encode(bytes: &[u8]) -> String {
...
+        tokio::time::sleep(duration).await
+    }
+}
+
+/// Sleep only if non-zero duration
+#[cfg(all(target_arch = "wasm32", any(target_os = "unknown", target_os = "none")))]
+pub(crate) async fn sleep(duration: Duration) {
+    use send_wrapper::SendWrapper;
+    if !duration.is_zero() {
+        SendWrapper::new(gloo_timers::future::sleep(duration)).await
```
pmarks, line 343: The inner future is not actually Send because it contains JS objects.  SendWrapper makes it Send, with dynamic checking that it is not used on another thread. This works for standard wasm32-unknown-unknown, and will crash 'safely' if used across threads somehow with Web Workers, so we can figure out how to fix that later.

kylebarron (approved):
Seems reasonable to me

alamb:
I am working on organizing a release

Wha do we want to do with this PR?

src/util.rs
kylebarron: I wonder if perhaps we should create our own type that is `Send` whenever the target arch is not `wasm32`?

I.e.
```rs
//! `MaybeSend` and `MaybeSync` traits for cross-platform async code.
//!
//! Reference: <https://github.com/iced-rs/iced/blob/master/futures/src/maybe.rs>

#[cfg(not(target_arch = "wasm32"))]
mod platform {
    /// A marker trait that enforces `Send` only on native platforms.
    ///
    pub use core::marker::Send as MaybeSend;
    /// A marker trait that enforces `Sync` only on native platforms.
    ///
    pub use core::marker::Sync as MaybeSync;
}

#[cfg(target_arch = "wasm32")]
mod platform {
    /// A marker trait that enforces `Send` only on native platforms.
    pub trait MaybeSend {}

    impl<T> MaybeSend for T {}

    /// A marker trait that enforces `Sync` only on native platforms.
    pub trait MaybeSync {}

    impl<T> MaybeSync for T {}
}

pub use platform::{MaybeSend, MaybeSync};
```

from https://github.com/zarrs/zarrs/blob/77aeab716ea219049f372ea750eecce4b72fbb7f/zarrs_plugin/src/maybe.rs#L1-L28

kevinjqliu:
btw theres a new "feature flags" section in the rust docs 
https://github.com/apache/arrow-rs-object-store/blob/2e859ee8d482d3732c09bb14052b59b75dd11194/src/lib.rs#L522-L548

PR #627, opened 2026-01-27
Unblock emulator based tests

### LINKED ISSUE #626: Emulator tests fail in CI due to an unsupported service version header in Azurite
Azure has released a new client (and I assume backend) while Azurite is lagging behind in supporting this header.
The upstream issue seems to be https://github.com/Azure/Azurite/issues/2623.

AdamGS:

Closes #626.

# Rationale for this change
 
Get CI to green and actually test all emulator based integrations.

Skip API check in azurite to make sure it doesn't break when Azure's tooling causes a difference between Azurite and Azure

None

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -136,17 +136,14 @@ jobs:
           KMS_KEY=$(aws --endpoint-url=http://localhost:4566 kms create-key --description "test key")
           echo "AWS_SSE_KMS_KEY_ID=$(echo $KMS_KEY | jq -r .KeyMetadata.KeyId)" >> $GITHUB_ENV
 
-      - name: Setup UV
-        uses: astral-sh/setup-uv@v7
-
       - name: Configure Azurite (Azure emulation)
         # the magical connection string is from
         # https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#http-connection-strings
-        # We specify an azure-cli version because 2.82 was released before azurite supported service version 2026-02-06
+        # We skip the API version check to prevent breaks related to differences between Azurite, Azure and the azure-cli, 
         # see https://github.com/Azure/Azurite/issues/2623
         run: |
-          echo "AZURITE_CONTAINER=$(docker run -d -p 10000:10000 -p 10001:10001 -p 10002:10002 mcr.microsoft.com/azure-storage/azurite:3.35.0)" >> $GITHUB_ENV
-          uvx --from 'azure-cli==2.81.0' --prerelease=allow az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
+          echo "AZURITE_CONTAINER=$(docker run -d -p 10000:10000 -p 10001:10001 -p 10002:10002 mcr.microsoft.com/azure-storage/azurite azurite -l /data --blobHost 0.0.0.0 --queueHost 0.0.0.0 --tableHost 0.0.0.0 --skipApiVersionCheck)" >> $GITHUB_ENV
+          az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
 
       - name: Setup Rust toolchain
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -139,9 +139,10 @@ jobs:
...
         # the magical connection string is from
         # https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#http-connection-strings
+        # We specify an azure-cli version because 2.82 was released before azurite supported service version 2026-02-06
         run: |
           echo "AZURITE_CONTAINER=$(docker run -d -p 10000:10000 -p 10001:10001 -p 10002:10002 mcr.microsoft.com/azure-storage/azurite)" >> $GITHUB_ENV
-          az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
+          pipx run --spec azure-cli==2.81.0 az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
```
AdamGS: pipx seems pretty slow compared to uv/uvx, but they are not pre-installed on github runners

kylebarron: I think it would be fine to add the `setup-uv` action on the previous line if you'd prefer

.github/workflows/ci.yml
```diff
@@ -139,9 +139,11 @@ jobs:
...
         # https://docs.microsoft.com/en-us/azure/storage/common/storage-use-azurite?tabs=visual-studio#http-connection-strings
+        # We specify an azure-cli version because 2.82 was released before azurite supported service version 2026-02-06
+        # see https://github.com/Azure/Azurite/issues/2623
         run: |
           echo "AZURITE_CONTAINER=$(docker run -d -p 10000:10000 -p 10001:10001 -p 10002:10002 mcr.microsoft.com/azure-storage/azurite)" >> $GITHUB_ENV
-          az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
+          pipx run --spec azure-cli==2.81.0 az storage container create -n test-bucket --connection-string 'DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://localhost:10000/devstoreaccount1;QueueEndpoint=http://localhost:10001/devstoreaccount1;'
```
kylebarron: So we aren't tied to azure-cli 2.81.0 forever? We just need to update azurite and azure-cli in tandem? Should we be using a pinned version of the azurite container too?

AdamGS: As far as I can tell, this is only until azurite adds support for this header value.
Azurite is great and probably as good as it gets, but there are definitely some differences between it and upstream Azure (#334 is another example).

AdamGS: We can pin the Azurite version, but it seems like the client/sdk are changed and released much more frequently, so any issue will always be triggered by them before azurite supports whatever's the change.

AdamGS: was still thinking about it but I guess I pushed a version that pins the server version, 3.35 is the current latest that was released last month.

.github/workflows/ci.yml
AdamGS: when including the installing of uv, it still saves ~40 seconds 🥳

.github/workflows/ci.yml
```diff
@@ -136,12 +136,17 @@ jobs:
           KMS_KEY=$(aws --endpoint-url=http://localhost:4566 kms create-key --description "test key")
           echo "AWS_SSE_KMS_KEY_ID=$(echo $KMS_KEY | jq -r .KeyMetadata.KeyId)" >> $GITHUB_ENV
 
+      - name: Setup UV
+        uses: astral-sh/setup-uv@v7
```
tustvold: I'm not sure this is in the Apache allow list of actions

AdamGS: I'll remove it

AdamGS: seems to be allowed [here](https://github.com/apache/infrastructure-actions/tree/99c4c1b2959b0bf07f9cc7dc34185dc39aacde19), also seems like apache requires actions to be pinned by hash, which I'm happy to do in a separate PR.

AdamGS: not that it matters if we pass skipApiVersionCheck

PR #628, opened 2026-01-29
Use platform specific `read_at` when available

### LINKED ISSUE #622: `LocalFileSystem`: use `read_at` instead of seek + read
Currently, the `LocalFileSystem` uses seek followed by a read, which saves a syscall per request.

We should be able to use `read_at` instead (https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html
It is available for unix (linux/osx / windows):
```
// Platform specific traits needed for read_at
#[cfg(unix)]
use std::os::unix::fs::FileExt; 
#[cfg(windows)]
use std::os::windows::fs::FileExt;

let file = Arc::new(File::open(path)?);
file.read_at(&mut buf, 500).unwrap();
...
```

**Additional context**

AdamGS:

Closes #622.

# Rationale for this change
 
Saves on at least one seek syscall per range

1. On windows and unix, make use of their platform specific `FileExt::read_at` functions.
2. Unifies the read implementation uses in `LocalStore::read_ranges` and for local files in `GetResult::bytes`, so this codepath is shared across both implementation which are currently very similar with one minor change.

No

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -116,6 +116,8 @@
 //!
 //! [`BufReader`]: buffered::BufReader
 //! [`BufWriter`]: buffered::BufWriter
+//! [`Read`]: std::io::Read
+//! [`Seek`]: std::io::Seek
 //!
 //! # Adapters
 //!
@@ -607,8 +609,6 @@ use bytes::Bytes;
 use chrono::{DateTime, Utc};
 use futures::{StreamExt, TryStreamExt, stream::BoxStream};
 use std::fmt::{Debug, Formatter};
-#[cfg(all(feature = "fs", not(target_arch = "wasm32")))]
-use std::io::{Read, Seek, SeekFrom};
 use std::ops::Range;
 use std::sync::Arc;
 
@@ -1650,22 +1650,11 @@ impl GetResult {
             #[cfg(all(feature = "fs", not(target_arch = "wasm32")))]
             GetResultPayload::File(mut file, path) => {
                 maybe_spawn_blocking(move || {
-                    file.seek(SeekFrom::Start(self.range.start as _))
-                        .map_err(|source| local::Error::Seek {
-                            source,
-                            path: path.clone(),
-                        })?;
+                    use crate::local::read_range;
 
-                    let mut buffer = if let Ok(len) = len.try_into() {
-                        Vec::with_capacity(len)
-                    } else {
-                        Vec::new()
-                    };
-                    file.take(len as _)
-                        .read_to_end(&mut buffer)
-                        .map_err(|source| local::Error::UnableToReadBytes { source, path })?;
+                    let buffer = read_range(&mut file, &path, self.range)?;
 
-                    Ok(buffer.into())
+                    Ok(buffer)
                 })
                 .await
             }
--- a/src/local.rs
+++ b/src/local.rs
@@ -19,6 +19,10 @@
 use std::fs::{File, Metadata, OpenOptions, metadata, symlink_metadata};
 use std::io::{ErrorKind, Read, Seek, SeekFrom, Write};
 use std::ops::Range;
+#[cfg(target_family = "unix")]
+use std::os::unix::fs::FileExt;
+#[cfg(target_family = "windows")]
+use std::os::windows::fs::FileExt;
 use std::sync::Arc;
 use std::time::SystemTime;
 use std::{collections::BTreeSet, io};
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -1650,22 +1648,11 @@ impl GetResult {
...
-                        Vec::with_capacity(len)
-                    } else {
-                        Vec::new()
-                    };
-                    file.take(len as _)
-                        .read_to_end(&mut buffer)
-                        .map_err(|source| local::Error::UnableToReadBytes { source, path })?;
-
-                    Ok(buffer.into())
+                    use crate::local::read_range;
```
AdamGS, line 1653: this makes me think that maybe `read_range` belongs elsewhere, maybe a new `file` module?

src/local.rs
```diff
@@ -988,26 +988,79 @@ pub(crate) fn read_range(
...
-    if read != to_read {
-        let error = Error::OutOfRange {
-            path: path.into(),
-            expected: to_read,
-            actual: read,
-        };
+        let mut buf_slice = &mut buf[..];
+        let mut offset = range.start;
+
+        while !buf_slice.is_empty() {
```
AdamGS, line 992: this is roughly the implementation of `read_exact_at`, just allows us to keep the same errors and track how many bytes we've read for the `OutOfRange` error branch.

src/local.rs
```diff
@@ -988,26 +988,79 @@ pub(crate) fn read_range(
...
+        #[cfg(target_family = "windows")]
+        use std::os::windows::fs::FileExt;
+
+        // Safety:
+        // Setting the buffer's length to its capacity is safe as it remains within its allocated memory.
+        // In cases where `read_exact_at` errors, the contents of the buffer are undefined,
+        // but we discard it without using it.
+        unsafe {
+            buf.set_len(to_read as usize);
+        }
```
AdamGS: This is the first use of `unsafe` in the crate, which I'm not thrilled about. The alternative seems to be to pre-fill it with `Vec::resize`, but that seems wasteful.

Dandandan: I think the later `&mut buf[..]` makes this UB.
What is the performance when initializing `buf` with `vec![0u8; 8]`?

Dandandan: The alternative would be reimplementing/vendoring part of this API https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#method.read_buf_at

Dandandan: Or a suggestion by Gemini (raw file IO):

```
use std::os::unix::io::AsRawFd;

let fd = file.as_raw_fd();
let mut buf = Vec::with_capacity(to_read);

let bytes_read = unsafe {
    // libc::pread takes a raw pointer (*mut c_void), avoiding the reference creation UB
    libc::pread(
        fd, 
        buf.as_mut_ptr() as *mut _, 
        to_read, 
        offset as i64
    )
};

if bytes_read >= 0 {
    unsafe { buf.set_len(bytes_read as usize) };
} else {
    // Handle error
}
```

Dandandan: Ah I see it even triggers https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#uninit_vec now

AdamGS: I'll try it out with pre-allocated data

AdamGS: just preallocated the buffer with 0s (`buf.resize(to_read as usize, 0_u8)`), results are:
```
read_ranges/local_fs/10 time:   [24.271 µs 24.605 µs 24.979 µs]
                        thrpt:  [3.0543 GiB/s 3.1008 GiB/s 3.1434 GiB/s]
                 change:
                        time:   [−9.2694% −7.9576% −6.6172%] (p = 0.00 < 0.05)
                        thrpt:  [+7.0861% +8.6456% +10.216%]
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  5 (5.00%) high mild
  4 (4.00%) high severe
read_ranges/local_fs/100
                        time:   [91.591 µs 92.250 µs 92.981 µs]
                        thrpt:  [8.2054 GiB/s 8.2703 GiB/s 8.3299 GiB/s]
                 change:
                        time:   [−19.988% −19.050% −18.121%] (p = 0.00 < 0.05)
                        thrpt:  [+22.131% +23.532% +24.982%]
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) low mild
  5 (5.00%) high mild
  5 (5.00%) high severe
read_ranges/local_fs/1000
                        time:   [877.00 µs 891.02 µs 907.75 µs]
                        thrpt:  [8.4047 GiB/s 8.5626 GiB/s 8.6994 GiB/s]
                 change:
                        time:   [−32.377% −30.518% −28.468%] (p = 0.00 < 0.05)
                        thrpt:  [+39.797% +43.921% +47.879%]
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  4 (4.00%) high mild
  4 (4.00%) high severe

get_opts_whole_file/local_fs
                        time:   [8.2127 ms 8.2487 ms 8.2891 ms]
                        thrpt:  [7.5400 GiB/s 7.5770 GiB/s 7.6101 GiB/s]
                 change:
                        time:   [−3.7223% −2.3673% −1.0502%] (p = 0.00 < 0.05)
                        thrpt:  [+1.0613% +2.4247% +3.8663%]
                        Performance has improved.
Found 12 outliers among 100 measurements (12.00%)
  3 (3.00%) high mild
  9 (9.00%) high severe

get_range_sequential/local_fs/10
                        time:   [223.74 µs 225.81 µs 228.25 µs]
                        thrpt:  [342.28 MiB/s 345.97 MiB/s 349.18 MiB/s]
                 change:
                        time:   [−8.4366% −7.0562% −5.8343%] (p = 0.00 < 0.05)
                        thrpt:  [+6.1958% +7.5919% +9.2139%]
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe
get_range_sequential/local_fs/100
                        time:   [2.2283 ms 2.2480 ms 2.2706 ms]
                        thrpt:  [344.07 MiB/s 347.54 MiB/s 350.61 MiB/s]
                 change:
                        time:   [−8.9263% −7.8699% −6.7098%] (p = 0.00 < 0.05)
                        thrpt:  [+7.1924% +8.5421% +9.8012%]
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  6 (6.00%) high mild
  3 (3.00%) high severe
get_range_sequential/local_fs/1000
                        time:   [23.699 ms 23.843 ms 24.011 ms]
                        thrpt:  [325.37 MiB/s 327.66 MiB/s 329.65 MiB/s]
                 change:
                        time:   [+1.8221% +2.7832% +3.8109%] (p = 0.00 < 0.05)
                        thrpt:  [−3.6710% −2.7079% −1.7895%]
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) low mild
  6 (6.00%) high severe
```

AdamGS: maybe a bit worse for `get_opts` but still shows an improvement, otherwise very similar within the noise of a laptop.

src/local.rs
```diff
@@ -975,8 +979,8 @@ pub(crate) fn read_range(
     range: Range<u64>,
 ) -> Result<Bytes> {
     // If none of the range is satisfiable we should error, e.g. if the start offset is beyond the
-    // extents of the file
-    if range.start >= file_len {
+    // extents of the file, or if its at the end of the file and wants to read a non-empty range.
+    if range.start > file_len || (range.start == file_len && !range.is_empty()) {
```
Dandandan: Isn't the added condition equivalent to `range.end > file_len` ?

AdamGS:
seems like I need some more work on the windows size

src/local.rs
AdamGS: from line 994-995:
```rust
// Don't read past end of file
let to_read = range.end.min(file_len) - range.start;
```

I think its actually technically ok to read with a range that ends past the end of the file (I think HTTP RANGE headers allow that too), the only issue is when trying to start at the end and reading more. I'll try and figure out if this error is actually tested, and add a test if not.

Dandandan: Ok

Dandandan (approved):
To me it looks great - perhaps someone else can take a look.

src/local.rs
Dandandan: Amazing, impressive!

AdamGS:
FWIW - I ran the same benchmark on an AWS box with a gpt-3 drive and the trend is very similar

src/lib.rs
crepererum: TBH I think file handles should be removed from the public API altogether, see #18.

src/local.rs
crepererum, line 977: Technically we could push this even further (maybe in a follow up): `read_at` doesn't modify the current read head at all. `seek_read` does, but the offset is always from the start of the file. So in both cases, you can read offsets from multiple threads in parallel. For a rotating disk this will probably not help (in contrary it might actually be counter-productive), but for an NVMe you can totally do that and in some cases get higher read throughput (esp. when your ranges are so tiny that the read latency is basically the bottleneck).

Dandandan: Yes this would allow it, but I think it probably needs to be controlled from somewhere else to avoid the overhead of creating a lot of threads, for example by allowing to reuse the `File` / `Arc<File>` from multiple threads and across `get_ranges` calls - that would also save a high number of open/close calls.

Dandandan: E.g. https://github.com/apache/datafusion/issues/19983

Dandandan: And there is also the `read_vectored_at` API (nightly) that does vectored IO, i.e. without using separate threads, which probably would be more performant.

https://doc.rust-lang.org/std/os/unix/fs/trait.FileExt.html#method.read_vectored_at

crepererum: Also see https://github.com/apache/arrow-rs-object-store/issues/18#issuecomment-3854039340

src/local.rs
Dandandan: I wondered if it makes sense to do the allocation in bulk in `get_ranges` instead of doing it per range - it seems that is beneficial for locality and saves some allocations.

PR #630, opened 2026-01-30
Add xattr-based attribute support to LocalFileSystem

### LINKED ISSUE #331: Allow attributes in local filesystem (or impove the error msg)
I was writing a multipart upload and it seemed to work fine, but I got an odd error of an operation isn't implemented. I had to clone the repo and search through the trait impls to find out that `object-store` doesn't allow attributes on local files.

I'd like to see support for attributes in the local file system, or if not can the message be more descriptive? Something like in further information that "lfs does not support attributes"

In `src/local.rs` for `put_multipart_opts`

Thanks!

Barre:
Closes #331

Store object attributes as extended attributes on the local filesystem, enabling put_opts and put_multipart_opts to persist attributes. Attributes are set on the staging file before atomic rename to preserve consistency.

- Add xattr dependency (gated by fs feature)
- Map standard attributes to user.* xattr namespace
- Read attributes back in get_opts

# Rationale for this change
 
The LocalFileSystem currently does not support attributes.

The LocalFileSystem now supports attributes.

The LocalFileSystem now supports attributes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,13 @@ description = "A generic object store interface for uniformly interacting with A
 keywords = ["object", "storage", "cloud"]
 repository = "https://github.com/apache/arrow-rs-object-store"
 rust-version = "1.85"
-include = ["src/**/*.rs", "README.md", "LICENSE.txt", "NOTICE.txt", "Cargo.toml"]
+include = [
+    "src/**/*.rs",
+    "README.md",
+    "LICENSE.txt",
+    "NOTICE.txt",
+    "Cargo.toml",
+]
 
 [package.metadata.docs.rs]
 all-features = true
@@ -46,10 +52,14 @@ url = "2.2"
 walkdir = { version = "2", optional = true }
 
 # Cloud storage support
-base64 = { version = "0.22", default-features = false, features = ["std"], optional = true }
+base64 = { version = "0.22", default-features = false, features = [
+    "std",
+], optional = true }
 form_urlencoded = { version = "1.2", optional = true }
 http-body-util = { version = "0.1.2", optional = true }
-httparse = { version = "1.8.0", default-features = false, features = ["std"], optional = true }
+httparse = { version = "1.8.0", default-features = false, features = [
+    "std",
+], optional = true }
 hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.11.0", default-features = false, optional = true }
 quick-xml = { version = "0.39.0", features = ["serialize", "overlapped-lists"], optional = true }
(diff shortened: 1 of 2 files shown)


crepererum:
I think this needs fixes / code adjustments for WASM & Windows.

Barre:
> I think this needs fixes / code adjustments for WASM & Windows.

My most recent changes should do the trick.

src/local.rs
```diff
@@ -882,9 +982,15 @@ impl MultipartUpload for LocalUpload {
...
         let src = self.src.take().ok_or(Error::Aborted)?;
         let s = Arc::clone(&self.state);
+        let attributes = std::mem::take(&mut self.attributes);
         maybe_spawn_blocking(move || {
             // Ensure no inflight writes
             let file = s.file.lock();
+
+            if !attributes.is_empty() {
+                set_xattrs(&src, &attributes)?;
+            }
```
crepererum: As far as I can see, you already handle the `is_empty` case within `set_xattrs` so this check is redundant. I think having the check only in one place (= `set_xattr`) makes the code easier to maintain and reason about.

src/local.rs
```diff
@@ -31,12 +29,14 @@ use std::{collections::VecDeque, path::PathBuf};
...
 use chrono::{DateTime, Utc};
-use futures_util::{FutureExt, TryStreamExt};
-use futures_util::{StreamExt, stream::BoxStream};
+use futures::{FutureExt, TryStreamExt};
+use futures::{StreamExt, stream::BoxStream};
 use parking_lot::Mutex;
 use url::Url;
 use walkdir::{DirEntry, WalkDir};
 
+#[cfg(feature = "xattr")]
```
crepererum: I think this one is wrong: the `#[cfg(not(feature = "xattr"))]` code uses this import.

src/local.rs
```diff
@@ -16,13 +16,11 @@
 // under the License.
 
 //! An object store implementation for a local filesystem
+#[cfg(feature = "xattr")]
+use std::borrow::Cow;
```
crepererum: Could you inline this import into `set_xattrs`? That's somewhat of a personal choice, but I find feature-guarded code always easier to reason about if it doesn't spill all over the place.

Cargo.toml
```diff
@@ -51,45 +51,42 @@ form_urlencoded = { version = "1.2", optional = true }
...
 [target.'cfg(target_family="unix")'.dev-dependencies]
-nix = { version = "0.31.1", features = ["fs"] }
+nix = { version = "0.30.0", features = ["fs"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 web-time = { version = "1.1.0" }
 wasm-bindgen-futures = "0.4.18"
-futures-channel = {version = "0.3", features = ["sink"]}
 
 [features]
```
crepererum, line 89: Isn't this missing the `xattr` feature?

Barre: Ah, I thought you didn't want it by default.

crepererum: Yes, but you still need to define the feature. An optional dependency doesn't automatically define a feature, no?

Barre:
```
$ cargo metadata --format-version 1 --no-deps | jq '.packages[0].features'

[...]
  "xattr": [
    "dep:xattr"
  ]
```

crepererum: Weird, but since it passes the test this is obviously working 🤷

PR #631, opened 2026-02-02
fix(azure): correct Microsoft Fabric blob endpoint domain

kevinjqliu:

Closes #.

# Rationale for this change
 

Fixes a typo in the Azure URL parser where `blob.fabric.microsoft.net` was used instead of the correct `blob.fabric.microsoft.com` domain.

The Microsoft Fabric endpoints use the `.microsoft.com` TLD, not `.microsoft.net`:
- `https://<account>.dfs.fabric.microsoft.com`
- `https://<account>.blob.fabric.microsoft.com`

Reference: https://learn.microsoft.com/en-us/fabric/onelake/onelake-access-api

- Changed `blob.fabric.microsoft.net` → `blob.fabric.microsoft.com` in the `abfs`/`az` URL scheme parsing
- Added test case for `az://container@account.blob.fabric.microsoft.com/` URLs

Yes, this is a user-facing bug fix:

- Fixed Azure URL parsing to recognize `blob.fabric.microsoft.com` (was incorrectly `blob.fabric.microsoft.net`)

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -671,7 +671,7 @@ impl MicrosoftAzureBuilder {
                             self.container_name = Some(validate(parsed.username())?);
                         }
                         Some((a, "dfs.fabric.microsoft.com"))
-                        | Some((a, "blob.fabric.microsoft.net")) => {
+                        | Some((a, "blob.fabric.microsoft.com")) => {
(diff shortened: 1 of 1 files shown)


crepererum (commented):
thank you

PR #633, opened 2026-02-03
Add CRC64NVME checksum support

### LINKED ISSUE #611: Support CRC checksum
We are sinking large dataframes to S3. For AWS, SHA256 is the only supported checksum. When profiling a high-throughput setup (100 Gbps NIC), the checksum adds 25-30% overhead as measured by AWS upload times, compared with not using a checksum.

Support `CRC-64/NVME (CRC64NVME)` (which is the AWS default).

Turning off checksum completely (`"AWS_UNSIGNED_PAYLOAD": "true"`) improves performance but adds unacceptable object integrity risk for large files using multi-part upload.

**Additional context**
Use case: high-end data file sink and read from AWS S3 and other S3 storage back-ends using polars dataframe library.

Note. In addition, it would be nice to see SHA256 offloaded to hardware, even though not all hardware supports this. Currently all SHA256 operations are done on the CPU. This may require swapping the `ring` crate for a more modern crypto crate.

kdn36:

- Closes #611 

# Rationale for this change
 
Improve (AWS) S3 write performance without compromising object integrity.

Key change is that the `CRC64NVME` checksum algorithm, which is the AWS default checksum algorithm, is added as a supported checksum variant.

The `crc-fast` crate is added for maximum performance.

Implementation note: if more checksum algorithms are to be supported, a minor refactor how checksum variants are handled would be in order.

The following `storage_options` key-value pair will be honored:
```
    "aws_checksum_algorithm": "CRC64NVME",
```
A typical use case would combine this with
```
    "aws_unsigned_payload": "true",
```

Kindly review, thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ walkdir = { version = "2", optional = true }
 
 # Cloud storage support
 base64 = { version = "0.22", default-features = false, features = ["std"], optional = true }
+crc-fast = { version = "1.10" }
 form_urlencoded = { version = "1.2", optional = true }
 http-body-util = { version = "0.1.2", optional = true }
 httparse = { version = "1.8.0", default-features = false, features = ["std"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,7 +47,7 @@ walkdir = { version = "2", optional = true }
 
 # Cloud storage support
 base64 = { version = "0.22", default-features = false, features = ["std"], optional = true }
-crc-fast = { version = "1.10" }
+crc-fast = { version = "1.6" , optional = true }
 form_urlencoded = { version = "1.2", optional = true }
 http-body-util = { version = "0.1.2", optional = true }
 httparse = { version = "1.8.0", default-features = false, features = ["std"], optional = true }
@@ -80,7 +80,7 @@ cloud = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream
 azure = ["cloud", "httparse"]
 fs = ["walkdir", "tokio"]
 gcp = ["cloud", "rustls-pki-types"]
-aws = ["cloud", "md-5"]
+aws = ["cloud", "crc-fast", "md-5"]
 http = ["cloud"]
 tls-webpki-roots = ["reqwest?/rustls-tls-webpki-roots"]
 integration = ["rand", "tokio"]
--- a/src/aws/checksum.rs
+++ b/src/aws/checksum.rs
@@ -20,6 +20,7 @@ use std::str::FromStr;
 
 #[allow(non_camel_case_types)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
 /// Enum representing checksum algorithm supported by S3.
 pub enum Checksum {
     /// SHA-256 algorithm.
--- a/src/aws/checksum.rs
+++ b/src/aws/checksum.rs
@@ -24,12 +24,15 @@ use std::str::FromStr;
 pub enum Checksum {
     /// SHA-256 algorithm.
     SHA256,
+    /// CRC64-NVME algorithm.
+    CRC64NVME,
 }
 
 impl std::fmt::Display for Checksum {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match &self {
             Self::SHA256 => write!(f, "sha256"),
+            Self::CRC64NVME => write!(f, "crc64nvme"),
         }
     }
 }
@@ -40,6 +43,7 @@ impl FromStr for Checksum {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s.to_lowercase().as_str() {
             "sha256" => Ok(Self::SHA256),
+            "crc64nvme" => Ok(Self::CRC64NVME),
             _ => Err(()),
         }
     }
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -43,7 +43,6 @@ use async_trait::async_trait;
 use base64::Engine;
 use base64::prelude::BASE64_STANDARD;
 use bytes::{Buf, Bytes};
-use crc_fast;
 use http::header::{
     CACHE_CONTROL, CONTENT_DISPOSITION, CONTENT_ENCODING, CONTENT_LANGUAGE, CONTENT_LENGTH,
     CONTENT_TYPE,
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -399,27 +399,25 @@ impl Request<'_> {
     }
 
     pub(crate) fn with_payload(mut self, payload: PutPayload) -> Self {
-        let needs_sha256_for_signing = !self.config.skip_signature && self.config.sign_payload;
-        let needs_sha256_for_checksum = matches!(self.config.checksum, Some(Checksum::SHA256));
+        use std::cell::LazyCell;
 
-        let sha256_digest = if needs_sha256_for_signing || needs_sha256_for_checksum {
+        let sha256_digest = LazyCell::new(|| {
             let mut sha256 = Context::new(&digest::SHA256);
-            payload.iter().for_each(|x| sha256.update(x));
-            Some(sha256.finish())
-        } else {
-            None
-        };
+            for part in &payload {
+                sha256.update(part);
+            }
+            sha256.finish()
+        });
 
-        if needs_sha256_for_signing {
-            self.payload_sha256 = sha256_digest;
+        if !self.config.skip_signature && self.config.sign_payload {
+            self.payload_sha256 = Some(*sha256_digest);
         }
 
         match self.config.checksum {
             Some(Checksum::SHA256) => {
-                self.builder = self.builder.header(
-                    SHA256_CHECKSUM,
-                    BASE64_STANDARD.encode(sha256_digest.unwrap()),
-                );
+                self.builder = self
+                    .builder
+                    .header(SHA256_CHECKSUM, BASE64_STANDARD.encode(*sha256_digest));
             }
             Some(Checksum::CRC64NVME) => {
                 let crc_algo = crc_fast::CrcAlgorithm::Crc64Nvme;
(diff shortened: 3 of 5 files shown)


src/aws/checksum.rs
```diff
@@ -24,12 +24,15 @@ use std::str::FromStr;
 pub enum Checksum {
     /// SHA-256 algorithm.
     SHA256,
+    /// CRC64-NVME algorithm.
+    CRC64NVME,
```
orlp, line 29: This is *technically* a breaking change because this enum was (erroneously IMO) not marked as `#[non_exhaustive]`. It's up to the maintainer - I personally don't believe anyone is matching on `Checksum`, it would only break code that currently does such a match.

But @kdn36 you should definitely mark this enum as `#[non_exhaustive]`.

orlp (changes requested):

Cargo.toml
```diff
@@ -47,6 +47,7 @@ walkdir = { version = "2", optional = true }
 
 # Cloud storage support
 base64 = { version = "0.22", default-features = false, features = ["std"], optional = true }
+crc-fast = { version = "1.10" }
```
crepererum:
```suggestion
crc-fast = { version = "1.10", optional = true }
```

Cargo.toml
crepererum:
```suggestion
aws = ["cloud", "crc-fast", "md-5"]
```

src/aws/client.rs
```diff
@@ -708,30 +734,52 @@ impl S3Client {
...
+            (e_tag, checksum_sha256, checksum_crc64nvme)
         };
 
-        let content_id = if self.config.checksum == Some(Checksum::SHA256) {
-            let meta = PartMetadata {
-                e_tag,
-                checksum_sha256,
-            };
-            quick_xml::se::to_string(&meta).unwrap()
+        let content_id = if let Some(checksum) = self.config.checksum {
```
crepererum: So `copy` can read both checksum but will only write/propagate one?

src/aws/checksum.rs
alamb: Yeah, we would have to wait for the next breaking object store release. We could discuss making such a change and quickly release 0.14 (next breaking change) if needed

src/aws/checksum.rs
```diff
@@ -24,12 +24,15 @@ use std::str::FromStr;
 pub enum Checksum {
```
alamb, line 25: I double checked and indeed CRC64NVME seems to be the default suggestion:

 https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

Seems like it was added to the official SDK about a year ago:
- https://github.com/awslabs/aws-sdk-rust/issues/1240

alamb (commented):
Thank you @kdn36 and @orlp and @crepererum 

Do we have any benchmarks to show how much faster this is? I ask because I think that would help us tradeoff the breaking API and getting this out. For example, does writing a 4GB object with CRC64-NVME go 1% faster? 10% faster?

For example, I would have thought that the network connection was the bottleneck, but as the speed of networks has increased significantly compared to CPU advances in recent years I no longer know if this intuition is true

So TLDR is that this change seems like a good one to me, the only question is the timeline of the breaking change

orlp:
@alamb I don't have any publicly available benchmarks, but @kdn36 measured that for one of our workflows in Polars Cloud it saved ~25% *end-to-end* on the entire query. Rather significant.

src/aws/checksum.rs
alamb: If we are going to change the enum, perhaps we can add all the other supported checksums as well (we don't have to actually support them, but we can minimize API churn)


I think that would mean adding the following variants (with comments that they aren't yet supported)

CRC32
CRC32C
SHA1
MD5

crepererum:
Regarding the crate of choice: Since we -- I think -- have decided that this is a breaking change, I think we should merge this AFTER #585 and hook up the crc calculation into the crypto provider framework.

src/aws/checksum.rs
orlp: @alamb If the enum is marked as `#[non_exhaustive]` as I recommended above then adding new variants isn't a breaking change. I personally don't think it's a good idea to add variants which aren't actually supported.

alamb: Marking it as `#[non_exhaustive]` would also be fine

(my rationale to add Variants that are not supported was to 🎣  for help supporting them)

Cargo.toml
kdn36: Done. Note I had to choose a lower version (1.6.0) to ensure compatibility with the strict version control by `aws-sdk-s3` (outside of object_store).

src/aws/client.rs
kdn36: I am no expert, but as I understand, we do not send a checksum on copy (write) since we do not have the bytes to calculate one. We do collect the response checksum(s), but only retain them if they match the configured checksum algorithm (we could also raise a warning or error, I think). The use case of specifying a checksum algorithm on copy is to change the checksum metadata as stored in S3.

However, I failed to properly test this locally, as minio does not properly support checksums on multipart copy. Instead, it errors out (https://github.com/minio/minio/issues/17013).

Testing in CI and directly on AWS works as expected.

I added a test case where we modify the checksum algorithm as part of a copy. That works as expected. Unfortunately I could only verify this manually as I could not find an easy way to obtain the checksum metadata for programmatic verification (a simple `head()..meta` does not contain this metdata).

Again, I am new to this, and will happily stand corrected.

src/aws/checksum.rs
kdn36: Great point. Changed to `#[non-exhaustive]`. If this goes through, adding more algorithm variants should be straightforward.

src/aws/client.rs
```diff
@@ -43,6 +43,7 @@ use async_trait::async_trait;
 use base64::Engine;
 use base64::prelude::BASE64_STANDARD;
 use bytes::{Buf, Bytes};
+use crc_fast;
```
orlp: This `use` is... useless 😅

src/aws/client.rs
```diff
@@ -398,19 +400,40 @@ impl Request<'_> {
     }
 
     pub(crate) fn with_payload(mut self, payload: PutPayload) -> Self {
-        if (!self.config.skip_signature && self.config.sign_payload)
-            || self.config.checksum.is_some()
-        {
+        let needs_sha256_for_signing = !self.config.skip_signature && self.config.sign_payload;
+        let needs_sha256_for_checksum = matches!(self.config.checksum, Some(Checksum::SHA256));
+
+        let sha256_digest = if needs_sha256_for_signing || needs_sha256_for_checksum {
```
orlp: An idea but rather than doing this whole check beforehand with an `Option`, I think you can just do
```rust
use std::cell::LazyCell;

let sha256_digest = LazyCell::new(|| {
    let mut sha256 = Context::new(&digest::SHA256);
    for part in &payload {
        sha256.update(part);
    }
    sha256.finish()
});


if !self.config.skip_signature && self.config.sign_payload {
    self.payload_sha256 = *sha256_digest;
}


match self.config.checksum {
    Some(Checksum::SHA256) => {
        self.builder = self.builder.header(
            SHA256_CHECKSUM,
            BASE64_STANDARD.encode(*sha256_digest),
        );
    }
```

orlp (changes requested):
Some nitpicks, other than that looks good.

orlp (approved):

src/aws/client.rs
crepererum: I still don't understand why the match block below just doesn't copy both checksums that you've extracted from the headers (you don't even have to compute it). Why do we strip one of them?

kdn36: Updated. I was (falsely) assuming an unncessary level of strictness for the checkusm configuration setting.

kdn36:
> I'm still confused about the copy behavior, see https://github.com/apache/arrow-rs-object-store/pull/633/changes#r3044086641

Updated after re-review.

Note and fwiw, I oberved different behavior between AWS and Localstack, in that AWS will add a Checksum header for the Copy path, but Localstack will not.

With
```
$ cargo test --features aws copy_multipart_file_with_signature -- --no-capture
```

AWS:
```
$ cat debug_aws.out
running 2 tests
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = Some(
    "kZLCW3NPy62+MtrcKAicYNsOOfkMwgzi5XM/VyYazAw=",
)
[src/aws/client.rs:770:9] &checksum_crc64nvme = None
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = None
[src/aws/client.rs:770:9] &checksum_crc64nvme = Some(
    "m0nypWpUsto=",
)
test aws::tests::copy_multipart_file_with_signature_change_checksum ... ok
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = None
[src/aws/client.rs:770:9] &checksum_crc64nvme = Some(
    "m0nypWpUsto=",
)
test aws::tests::copy_multipart_file_with_signature ... ok
```
Localstack:
```
$ cat debug_localstack.out
running 2 tests
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = None
[src/aws/client.rs:770:9] &checksum_crc64nvme = None
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = None
[src/aws/client.rs:770:9] &checksum_crc64nvme = None
test aws::tests::copy_multipart_file_with_signature_change_checksum ... ok
[src/aws/client.rs:714:9] "start put_part" = "start put_part"
[src/aws/client.rs:769:9] &checksum_sha256 = None
[src/aws/client.rs:770:9] &checksum_crc64nvme = None
test aws::tests::copy_multipart_file_with_signature ... ok
```

kdn36:
Checking in - anything still needed from me for this PR?

alamb:
@crepererum do you have time to review this PR?

crepererum (approved):

PR #636, opened 2026-02-10
fix(aws): Include default headers in signature calculation (#484)

### LINKED ISSUE #484: AWS/S3 Default Headers are not considered for signature calculation
**Describe the bug**
S3 supports quite a few headers in system defined metadata and user defined metadata described in https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#SysMetadata. When setting these headers as default headers, they are not considered in the signature calculation.

As a result, S3 throws an error along the lines of
```
<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n
<Error>
	<Code>AccessDenied</Code>
	<Message>There were headers present in the request which were not signed</Message>
	<HeadersNotSigned>x-amz-meta-s1, x-amz-tagging</HeadersNotSigned>
	...
</Error>
```

**To Reproduce**
```
let s3 = AmazonS3Builder::new()
    .with_bucket_name(bucket)
    .with_client_options({
        ClientOptions::new().with_default_headers({
            let mut headers = HeaderMap::new();
            headers.insert("x-amz-meta-s1", "t1".parse().unwrap());
            headers.insert("x-amz-tagging", "Project=blue&Classification=confidential".parse().unwrap());
            headers
        })
    })
    .build()?;

let store: Arc<dyn ObjectStore> = Arc::new(s3);

store.put(&key.clone().into(), "content".into()).await?;
println!("Uploaded file with key: {}", key);
```

**Expected behavior**
The request should succeed and S3 HeadObject should echo these headers back.

**Additional context**
N/A

singhsaabir:
When using `ClientOptions::with_default_headers()` to set S3 metadata headers (like `x-amz-meta-*` or `x-amz-tagging`), these headers were not included in the AWS SigV4 signature calculation, causing S3 to reject requests with "headers present i
n the request which were not signed".

This fix adds default headers to the request before signing in all three S3 request paths: `S3Client::request()`, `bulk_delete_request()`, and `get_request()`.

Closes #484.

# Rationale for this change

Explained in the issue - default headers set via `ClientOptions::with_default_headers()` were not being signed, causing S3 to reject requests.

- Added `get_default_headers()` getter to `ClientOptions`
- Modified `S3Client::request()`, `bulk_delete_request()`, and `get_request()` to add default headers before signing
- Added test verifying default headers are included in signature

No breaking changes. Default headers set via `ClientOptions::with_default_headers()` will now work correctly with S3.

### DIFF
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -1110,9 +1110,7 @@ mod tests {
 
         let config = default_headers_config(&mock);
         let client = S3Client::new(config, HttpClient::new(reqwest::Client::new()));
-        let result = client
-            .bulk_delete_request(vec![Path::from("test")])
-            .await;
+        let result = client.bulk_delete_request(vec![Path::from("test")]).await;
 
         assert!(result.is_ok());
         mock.shutdown().await;


src/aws/client.rs
```diff
@@ -1010,4 +1020,72 @@ mod tests {
...
+            request_payer: false,
+        };
+
+        let client = S3Client::new(config, HttpClient::new(reqwest::Client::new()));
+        let result = client
+            .request(Method::PUT, &Path::from("test"))
+            .with_payload(PutPayload::default())
+            .do_put()
+            .await;
+
```
crepererum, line 1095: This covers `fn request`. Could you extend the test to also include `bulk_delete_request` & `get_request`?

singhsaabir: Thank you for the quick review! Added tests.

Issue #507, opened 2025-10-14
Introduce common configuration patterns for file/CLI/env-var usage


nyurik:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

Most users of this great crate will usually allow their users to configure bucket access. This includes things like bucket name, region, tokens or automatic environment variable parsing, and lots and lots of other parameters.  Most of the users of this crate will not even be aware of the various configuration options available, nor do they want to be.  Yet, if they get it wrong, some end users who need a very specific authentication mode or other non-so-common setup will suffer. This lib supports it, they know about it, but the specific app that uses this crate is not passing through the right config.

**Describe the solution you'd like**

Introduce a set of reusable components targeting all 3 usecases:
* configuration files - some serde-instrumented structs/enums that can parse yaml/toml/json files in the same way to allow any needed configuration
* CLI - clap-instrumented structs for the same
* env vars - this one seem to be mostly done - but would be good if it works well with the other two methods

**Describe alternatives you've considered**
Each user of this lib implement it by hand

PR #638, opened 2026-02-11
docs: document the `with_config`  options

CommanderStorm:

This PR works towards https://github.com/apache/arrow-rs-object-store/issues/507

# Rationale for this change
 
Currently, a user would not quite know what options exist 

All `with_config` APIs get documentation rewritten.
Due to the amount of documentation required to properly document what this does, I chose markdown.

This has the advantage that dependent projects know what the configuration is and can semi-reli on this. I think everyone wins via this.

I will maintain these docs (I currently already do, but out of tree).
These docs beign in the tree would be more helpfull I assume.

It is a docs-only change 😉

image

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -597,7 +597,18 @@ impl AmazonS3Builder {
         self
     }
 
-    /// Set an option on the builder via a key - value pair.
+    #[doc = concat!(
+      "Set an option by key - value pair\n",
+      "\n",
+      "These are the available options:\n",
+      include_str!("config.md"),
+      "\n",
+      "**TIP:**\n",
+      "below settings are also available under the `aws_` prefix.\n",
+      "This can be useful if you want to have different cloud providers.\n",
+      "\n",
+      include_str!("../client/config.md")
+    )]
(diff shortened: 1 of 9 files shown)


CommanderStorm:
> out-of-sync risk

Shure, that is always a risk.

Cannot have better docs without either
- adding macro-complexity or
- documenting it.

I would be very motivated to take on this maintenance burden.

Obviously, there is no way for you to trust me to keep my word on this one.
What we could do is to revisit this in a year, since I have only been at this for a few months -> https://github.com/maplibre/martin/commits/main/docs/src/pmtiles

This being markdown has the advantage that consumers can more easily consume this documentation into their documentation (aka without relying on rustdoc json).

> Due to the limited nature of markdown, this is gonna be hard to replicate in a table view

Warnings work alright.

The datatypes are also not documented for the Enum way, but I think they should be fairly obvious by having an example.

https://github.com/CommanderStorm/arrow-rs-object-store/blob/a8cf83ee59c001a5529db0cb95a21c13b5a002a8/src/aws/config.md#aws-specific-request-behavior--fallbacks

PR #641, opened 2026-02-13
fix(azure): fix integer overflow in Fabric token expiry check

### LINKED ISSUE #640: Azure Fabric: Unsigned integer underflow when fetching token causes integer overflow panic
**Describe the bug**

`FabricTokenOAuthProvider::fetch_token` performs unsigned integer subtraction when checking token expiry, which can cause an integer underflow panic when the cached token has expired:

```
let exp_in = expiry - Self::get_current_timestamp();
if exp_in > TOKEN_MIN_TTL {
  return Ok(TemporaryToken {
    token: Arc::new(AzureCredential::BearerToken(storage_access_token.clone())),
    expiry: Some(Instant::now() + Duration::from_secs(exp_in)),
  });
}
```

With `opt-level=3`, `exp_in` wraps to ~`u64::MAX` which results in "overflow when adding duration to instant".

**To Reproduce**

This is all from code analysis so take it with a grain of salt 😅 

A user of Daft shared a Fabric notebook that encountered this error in https://github.com/Eventual-Inc/Daft/issues/6007: https://github.com/djouallah/Fabric_Notebooks_Demo/blob/main/ETL/Light_ETL_Python_Notebook.ipynb

This issue **may be** related, but I haven't actually run it because I don't have a Fabric environment.

I can reproduce this with a test case using an expired cached token. Full test code available in PR to follow.

**Expected behavior**

On expiry, we should be able to fetch a new token without issue.

**Proposed Fix**

Use `saturating_sub` instead of `-` for safe subtraction:

```rust
let exp_in = expiry.saturating_sub(Self::get_current_timestamp());
```

When expired, saturating_sub returns 0, which fails the `exp_in > TOKEN_MIN_TTL` check and correctly falls through to the refresh path.

desmondcheongzx:

Closes #640.

# Rationale for this change
 
`FabricTokenOAuthProvider::fetch_token` uses unsigned integer subtraction when checking token expiry. When a cached token has expired (`expiry < current_timestamp`), this causes an overflow when adding duration to instant. This prevents the provider from falling through to the token refresh path as intended.

The issue contains a more detailed breakdown.

Replace `-` with `saturating_sub()` for unsigned integer subtractions in the token expiry checks.

Add a repro test that now passes.

Bug fix only.

### DIFF
--- a/src/azure/credential.rs
+++ b/src/azure/credential.rs
@@ -1252,7 +1252,7 @@ mod tests {
             fabric_workload_host: "fake".to_string(),
             fabric_session_token: "session-token".to_string(),
             fabric_cluster_identifier: "cluster-id".to_string(),
-            storage_access_token: Some(expired_token.clone()),
+            storage_access_token: Some(expired_token),
             token_expiry: Some(expired_timestamp),
         };
 


crepererum:
Just for my understanding: this would trigger if you have a long-running process (e.g. a server) that gets an access token but then doesn't perform any Azure operations for a long time (probably hours?) and then performs a new request (at which point the subtraction underflows), right?

desmondcheongzx:
@crepererum yes I think this underflow would be triggered in the scenario you described.

I also think that simpler scenarios would trigger it. For example:
- Long running user script writing to Azure for longer than token expiry time
- Cache expires (maybe after 60 mins, maybe longer), call `fetch_token` again to do another write
- Underflow

crepererum (approved):
solid fix + test, thank you

PR #643, opened 2026-02-15
feat(local): add fsync to LocalFileSystem for durability

### LINKED ISSUE #661: Reliable file writes for LocalFileSystem: explicit close error checking and opt-in sync

Rust's `File::drop()` silently ignores errors from the underlying `close()` syscall. On network filesystems like NFS, `close()` can fail and indicate data loss, but currently `LocalFileSystem` has no way to detect this. Data may appear to be written successfully while the server actually reports an error that is never surfaced to the caller.

Additionally, for workloads requiring stronger durability guarantees, there is no way to ensure data is flushed to persistent storage before the file is closed and renamed into place.

1. Explicit close with error checking: replace the silent drop with platform-specific close calls that propagate errors back to the caller.
2. Opt-in "sync on close" behaviour: add a builder method on `LocalFileSystem` that calls `File::sync_all()` before closing, providing stronger durability guarantees at the cost of performance.

We could always call `File::sync_all()` but it would be too expensive for local disks where it's unnecessary.

**Additional context**

Polars ran into the same issue: https://github.com/pola-rs/polars/issues/21002.
It got fixed in https://github.com/pola-rs/polars/pull/21588.

Barre:
- closes https://github.com/apache/arrow-rs-object-store/issues/661

Call sync_all() on written files and fsync parent directories at all write-path boundaries (put, copy, rename, multipart complete) so that a successful return guarantees data is durable on disk, matching the implicit contract of cloud object stores.

# Rationale for this change
 
When LocalFileSystem::put (or copy/rename/multipart complete) returns Ok, callers reasonably expect the data to be durable on disk as this is the implicit contract of every cloud object store like S3 or GCS. 

However, LocalFileSystem never called fsync/sync_all, meaning the OS was free to keep the data in its page cache indefinitely. A crash or power loss after a successful put could result in data loss or zero-length files.

This change adds sync_all() calls on written files and fsync on parent directories at every write-path boundary (put_opts, copy_opts, rename_opts, multipart complete), ensuring that when an operation returns success, both the file contents and the directory entry pointing to them are durable on stable storage.

No.

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -114,6 +114,9 @@ pub(crate) enum Error {
     #[error("Filenames containing trailing '/#\\d+/' are not supported: {}", path)]
     InvalidPath { path: String },
 
+    #[error("Unable to sync data to disk for {}: {}", path.display(), source)]
+    UnableToSyncFile { source: io::Error, path: PathBuf },
+
     #[error("Upload aborted")]
     Aborted,
 }
@@ -228,6 +231,8 @@ pub struct LocalFileSystem {
     config: Arc<Config>,
     // if you want to delete empty directories when deleting files
     automatic_cleanup: bool,
+    // if true, fsync written files and their parent directories after writes
+    fsync: bool,
 }
 
 #[derive(Debug)]
@@ -255,6 +260,7 @@ impl LocalFileSystem {
                 root: Url::parse("file:///").unwrap(),
             }),
             automatic_cleanup: false,
+            fsync: false,
         }
     }
 
@@ -273,6 +279,7 @@ impl LocalFileSystem {
                 root: absolute_path_to_url(path)?,
             }),
             automatic_cleanup: false,
+            fsync: false,
         })
     }
 
@@ -284,6 +291,25 @@ impl LocalFileSystem {
     /// Enable automatic cleanup of empty directories when deleting files
     pub fn with_automatic_cleanup(mut self, automatic_cleanup: bool) -> Self {
         self.automatic_cleanup = automatic_cleanup;
+        self
+    }
+
+    /// Enable `fsync` after writes for durability
+    ///
+    /// When enabled, [`LocalFileSystem`] calls [`File::sync_all`] on written files and fsyncs
+    /// the affected parent directories before a write operation
+    /// ([`put_opts`](ObjectStore::put_opts), [`copy_opts`](ObjectStore::copy_opts),
+    /// [`rename_opts`](ObjectStore::rename_opts), and multipart upload completion) returns
+    /// success. This guarantees that both the file contents and the directory entries pointing
+    /// to them are durable on stable storage, matching the implicit durability contract of
+    /// remote object stores such as S3 or GCS.
+    ///
+    /// This trades write throughput for durability and is **disabled by default**.
+    ///
+    /// Note that directory fsync is only performed on Unix; on other platforms (e.g. Windows)
+    /// it is a no-op, as directories cannot be portably opened and synced.
+    pub fn with_fsync(mut self, fsync: bool) -> Self {
+        self.fsync = fsync;
         self
     }
 }
@@ -374,8 +400,9 @@ impl ObjectStore for LocalFileSystem {
         }
 
         let path = self.path_to_filesystem(location)?;
+        let fsync = self.fsync;
         maybe_spawn_blocking(move || {
-            let (mut file, staging_path) = new_staged_upload(&path)?;
+            let (mut file, staging_path) = new_staged_upload(&path, fsync)?;
             let mut e_tag = None;
 
             let err = match payload.iter().try_for_each(|x| file.write_all(x)) {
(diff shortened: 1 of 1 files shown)


crepererum:
I think this is technically what we should do. I wonder however if we want to offer some kind of opt-out mechanism (I think in vein of "safety+sanity by default", it shouldn't be opt-in). WDYT @tustvold?

Barre:
Defaulting to false seems a bit reckless to me, users are most certainly expecting the local "object store" to be durable, as are all other object store implementations in this crate (aside from memory...).

It was definitely a great surprise to me to learn that it was not.

src/local.rs
```diff
@@ -882,11 +1000,21 @@ impl MultipartUpload for LocalUpload {
     async fn complete(&mut self) -> Result<PutResult> {
         let src = self.src.take().ok_or(Error::Aborted)?;
         let s = Arc::clone(&self.state);
+        let fsync = self.fsync;
         maybe_spawn_blocking(move || {
             // Ensure no inflight writes
             let file = s.file.lock();
+            if fsync {
```
crepererum: why is there an fsync before this rename but not before the one in `rename_opts`?

src/local.rs
```diff
@@ -786,23 +855,69 @@ impl LocalFileSystem {
...
+    if fsync {
+        let mut first_existing = parent;
+        while !first_existing.exists() {
+            first_existing = first_existing.parent().unwrap_or(first_existing);
+        }
+
+        std::fs::create_dir_all(parent).map_err(|source| {
+            let path = parent.into();
+            Error::UnableToCreateDir { source, path }
+        })?;
```
crepererum: If you pull this create_dir_all before the `if fsync`, then it's the same code for fsync and no-fsync. That makes it easier to reasons about, since it's really just the fsync-block that differs, not the "actual" I/O operation.

crepererum (commented):
While this gets the job done, I'm getting a bit uneasy about he number of IF-branches in the code and potential "we forgot another fsync here" issue. May I suggest that we pull out the file I/O ops into functions that are more re-usable, e.g.:

```rust
fn rename(file: ..., fsync: bool) -> Result<()> {
    if fsync {
        file.sync_all().map_err(|source| Error::UnableToSyncFile {
            source,
            path: src.clone(),
         })?;
     }

     std::fs::rename(&src, &s.dest)
         .map_err(|source| Error::UnableToRenameFile { source })?;

     if fsync {
         fsync_parent_dir(&s.dest)?;
     }

     Ok(())
}

fn hard_link(..., fsync: bool) -> Result<()> { ... }
```

adamreeve:
Hi @Barre, are you still planning on following up on this and addressing the latest review comments? I'd also like to see this feature added and could help with this if you no longer have time.

alamb:
> 
Hi @Barre, are you still planning on following up on this and addressing the latest review comments? I'd also like to see this feature added and could help with this if you no longer have time.


Thanks -- I am also working on preparing the next release -- if you are interested in pulling this over the line @adamreeve  I will put it on the list of

Barre:
Thanks for the nudge 🙏🏻 

I'll try to push something in few hours.

alamb:
Thank you for bearing with us 🙏

src/local.rs
Barre: `complete()` syncs the freshly-written staging file before publishing it; rename_opts just moves an already-durable object, so there's no new file data to sync.

crepererum:
This critique still stands: https://github.com/apache/arrow-rs-object-store/pull/643#pullrequestreview-4067116956

To keep this code maintainable, I think we need to encapsulate the `std::fs` operations with their respective fsync-calls into helper functions. Otherwise the risk of forgetting the fsync after the file system modification is just too high.

Barre:
> This critique still stands: [#643 (review)](https://github.com/apache/arrow-rs-object-store/pull/643#pullrequestreview-4067116956)
> 
> To keep this code maintainable, I think we need to encapsulate the `std::fs` operations with their respective fsync-calls into helper functions. Otherwise the risk of forgetting the fsync after the file system modification is just too high.

I hope this looks better to you now.

src/local.rs
```diff
@@ -816,23 +840,168 @@ impl LocalFileSystem {
...
+        Some(parent) => fsync_dir(parent),
+        None => Ok(()),
+    }
+}
+
+/// Fsyncs `dir_path` so that changes to its directory entries are durable.
+///
+/// This is only meaningful on Unix; on other platforms (e.g. Windows) directories cannot be
+/// portably opened as a [`File`] and synced, so this is a no-op.
+fn fsync_dir(dir_path: &std::path::Path) -> io::Result<()> {
```
alamb, line 992: I spent some time reviewing this and reviewing the calls to `fsync` -- to be clear the the [File::fsync_all](https://doc.rust-lang.org/beta/std/fs/struct.File.html#method.sync_all) docs despite the name, only flushes the contents of this file descriptor to disk. 

It also might be worthwhile pointing out that fsync_dir is actually doing three syscalls (`open`, `fsync` and  then an implicit `close`). However, apparently this is the same pattern as sqlite uses

https://github.com/sqlite/sqlite/blob/9a5801af3a411c880f556233dda7a3097967f8af/src/os_unix.c#L3945-L3948

so looks good to me

alamb (approved):
I reviewed this PR and I think it looks good now -- I verified test coverage via `llvm-cov` . Thank you so much @Barre and @crepererum 

One thing I did notice is that basically all error paths are untested in this LocalFileSystem as they require IO errors to trigger. I think this is a gap that we should eventually fix, but as it was not made worse by this PR I don't think anything is required here. 

cc @crepererum in case you want a chance to review again before we merge

Update I filed this ticket for error handling
 - https://github.com/apache/arrow-rs-object-store/issues/753

alamb:
I made a small fix in f9e526fe9b60c1b986c1b5d7c93e26ee6d6053c2  for a [rust docs CI failure here](https://github.com/apache/arrow-rs-object-store/actions/runs/27414498468/job/81498984312?pr=643) and then merged up from main to try and get a clean CI run

alamb:
I'll plan to merge this tomorrow unless anyone else would like longer to review

FYI @jgiannuzzi and @adamreeve as I believe you expressed interest in this feature as well

adamreeve (approved):
This looks good to me, thanks for adding this @Barre!

kevinjqliu (approved):
LGTM

Codex flagged one potential issue: `RenameTargetMode::Create` copies with fsync but deletes the source via `delete_location` without fsyncing the source parent dir, so `with_fsync(true)` could make the destination durable while leaving the source deletion non-durable.
Heres the fix that it proposed https://github.com/apache/arrow-rs-object-store/commit/f6e15ffa3e7c0e93e16b9a5a5c42be391a4202ad

alamb:
> f6e15ff

Thanks @kevinjqliu  -- for the case of moving things along, I will merge this PR. can you make the proposed fix a separate PR?

kevinjqliu:
Create https://github.com/apache/arrow-rs-object-store/pull/758 as a follow up, thanks!

PR #648, opened 2026-02-21
Switch TokenCache to RWLock

tustvold:

- Closes https://github.com/apache/arrow-rs-object-store/issues/541.

# Rationale for this change
 

### DIFF
--- a/src/client/token.rs
+++ b/src/client/token.rs
@@ -56,7 +56,7 @@ impl<T> Default for TokenCache<T> {
     }
 }
 
-impl<T: Clone + Send> TokenCache<T> {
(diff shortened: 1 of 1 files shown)


src/client/token.rs
```diff
@@ -56,7 +56,7 @@ impl<T> Default for TokenCache<T> {
     }
 }
 
-impl<T: Clone + Send> TokenCache<T> {
+impl<T: Clone + Send + Sync> TokenCache<T> {
```
tustvold, line 59: This is necessary to make get_or_insert_with Send, which otherwise makes clippy sad

crepererum (approved):

Issue #529, opened 2025-11-05
Support upper-case configuration options in parse_url_opts


christianbudgetthuis:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
When using the from_env function of the object store builders, environmental variables are expected to be upper-case, as is coventional. They are then converted to lower-case to be able to parse them as configuration options (see https://github.com/apache/arrow-rs-object-store/blob/1b8ecc7f41eab2a3c2c373f1c2fe21bebe391727/src/azure/builder.rs#L517).

However, when using parse_url_opts the input is _not_ converted to lower-case (see https://github.com/apache/arrow-rs-object-store/blob/1b8ecc7f41eab2a3c2c373f1c2fe21bebe391727/src/parse.rs#L146) but, of course, still needs to be lower-case as this is what the parse function expects. The documentation even recommends passing std::env::vars() to the parse_url_opts function, which does not work as expected if you are using upper-case environmental variables.

**Describe the solution you'd like**
Make parse_url_opts work with uppercase env. var names, e.g. add to_ascii_lowercase in the builder macro.

**Describe alternatives you've considered**
Add a note in the documentation that the configuration options need to be lower-case for parse_url_opts and uppercase when using from_env() functions for concrete object stores.

tustvold:
Why not just convert the options to lower case on input? The current design was chosen as it allows users to choose if they want case sensitivity or not

christianbudgetthuis:
You are right, that is a possible workaround.

However, it seems a little confusing to me that the from_env() function in the builder does automatically convert to lower-case, while the parse_url_opts() function does not. But if that remains the case, I would suggest the documentation (https://docs.rs/object_store/latest/object_store/fn.parse_url_opts.html) to mention that the input here is case-sensitive and needs to be lower-case. Especially, since the builder documentation (https://docs.rs/object_store/latest/object_store/azure/struct.MicrosoftAzureBuilder.html#method.from_env), does state them in upper-case, and one could assume that both functions work alike.

PR #652, opened 2026-02-27
Unify `from_env` behaviours

miraclx:

Closes https://github.com/apache/arrow-rs-object-store/issues/529.

# Rationale for this change
 
I just ran into this issue myself after generalizing my use of `AmazonS3` via [`from_env`](https://github.com/apache/arrow-rs-object-store/blob/907653e46232edb3a635969bf288324b68d62cdc/src/aws/builder.rs#L575) to `DefaultObjectStoreRegistry`. A cursory look at the code suggests the intended effect is equivalence, but that's not the observed effect.

This is so due to [`AmazonS3Builder::from_env`](https://github.com/apache/arrow-rs-object-store/blob/907653e46232edb3a635969bf288324b68d62cdc/src/aws/builder.rs#L581) calling `.to_ascii_lowercase`, and well enough, 
the [GCP](https://github.com/apache/arrow-rs-object-store/blob/907653e46232edb3a635969bf288324b68d62cdc/src/gcp/builder.rs#L284) and [`Azure`](https://github.com/apache/arrow-rs-object-store/blob/907653e46232edb3a635969bf288324b68d62cdc/src/azure/builder.rs#L517) do the same thing.

`parse_url_opts` being the outlier here suggests it's the wrong one, and therefore to offer some consistency, it should be patched to match.

PS: `AmazonS3Builder` says

https://github.com/apache/arrow-rs-object-store/blob/907653e46232edb3a635969bf288324b68d62cdc/src/aws/builder.rs#L549

Which also contradicts its very own implementation. (It's also the only one to say this)

And finally, I don't think a single dependency should dictate the casing of project-wide env vars. (especially as AWS itself documents these in `SCREAMING_SNAKE_CASE`.

`parse_url_opts`, like the other `from_env` implementations, now also parses env vars case-insensitively.

Yes, documentation updated

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -177,7 +177,7 @@ pub fn parse_url(url: &Url) -> Result<(Box<dyn ObjectStore>, Path), super::Error
 /// * `options`: A list of key-value pairs to pass to the [`ObjectStore`] builder.
 ///   Note different object stores accept different configuration options, so
 ///   the options that are read depends on the `url` value. One common pattern
-///   is to pass configuration information via process variables using
+///   is to pass configuration information via process variables using [`std::env::vars`].
 ///
 /// Returns
 /// - An [`ObjectStore`] of the corresponding type


src/parse.rs
```diff
@@ -144,7 +144,7 @@ macro_rules! builder_opts {
     ($builder:ty, $url:expr, $options:expr) => {{
         let builder = $options.into_iter().fold(
             <$builder>::new().with_url($url.to_string()),
-            |builder, (key, value)| match key.as_ref().parse() {
+            |builder, (key, value)| match key.as_ref().to_ascii_lowercase().parse() {
```
crepererum, line 147: Can we have a test for this code change?

miraclx: Done, updated `test_url_http` to account for this

src/parse.rs
```diff
@@ -178,8 +178,6 @@ pub fn parse_url(url: &Url) -> Result<(Box<dyn ObjectStore>, Path), super::Error
 ///   Note different object stores accept different configuration options, so
 ///   the options that are read depends on the `url` value. One common pattern
 ///   is to pass configuration information via process variables using
-///   [`std::env::vars`]. Keys must be lower-case and match the list of supported
```
kylebarron, line 181: This cuts off the sentence?

src/aws/builder.rs
```diff
@@ -546,7 +546,6 @@ impl AmazonS3Builder {
     ///
     /// All environment variables starting with `AWS_` will be evaluated.
     /// Names must match acceptable input to [`AmazonS3ConfigKey::from_str`].
-    /// Only upper-case environment variables are accepted.
```
kylebarron, line 549: Is there a line like this in the docs for the gcp and azure builders?

src/parse.rs
miraclx: Fixed, thanks!

src/aws/builder.rs
miraclx: No, only AWS incorrectly mentions this.

crepererum:
thank you

PR #670, opened 2026-03-18
fix: Strip prefix in PrefixStore::get_opts

### LINKED ISSUE #664: PrefixStore::head doesn't strip prefix from meta.location
**Describe the bug**

Calling `.head()` on a prefix store returns an ObjectMeta with a `.location` that contains the prefix. Other methods such as `.list()` strip the prefix: https://docs.rs/object_store/latest/src/object_store/prefix.rs.html#146

**To Reproduce**

```rust
let store = PrefixStore::new("client-jail", InMemory::new());
let path = Path::from("my-file");
match store.head(&path).await {
  Ok(meta) => {
    store.get(meta.location).await.expect("to not try open client-jail/client-jail/my-file");
  ...
```

**Expected behavior**

From prefix.rs, I think PrefixStore's `head()` method needs to call strip_meta.

**Current Workaround**
Reusing the path passed into `head()`.

Thank you ❤️

liamdiprose:

Closes #664

Calls strip_meta on the `.meta` field of the `GetResult` from the inner store of the PrefixStore.

The meta.location coming from PrefixStore::head, get_opts, etc no longer has the prefix. I can't imagine anyone relying on the functionality of this bug.

### DIFF
--- a/src/prefix.rs
+++ b/src/prefix.rs
@@ -117,7 +117,16 @@ impl<T: ObjectStore> ObjectStore for PrefixStore<T> {
 
     async fn get_opts(&self, location: &Path, options: GetOptions) -> Result<GetResult> {
         let full_path = self.full_path(location);
-        self.inner.get_opts(&full_path, options).await
+        let slf_prefix = self.prefix.clone();
+        self.inner
+            .get_opts(&full_path, options)
+            .await
+            .map(move |opts| GetResult {
+                payload: opts.payload,
+                range: opts.range,
+                attributes: opts.attributes,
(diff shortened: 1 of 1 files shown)


asubiotto:
Friendly ping, we're running into an issue related to this so would be great to get this PR over the line. I think this breaks listing table in datafusion 53 cc @alamb

crepererum:
@asubiotto feel free to pick this up (filing a new PR, since you don't have edit permissions I think) and add a regression test.

asubiotto:
Done: #686

PR #676, opened 2026-03-23
feat(local): explicit close with error checking for LocalFileSystem

jgiannuzzi:

Implements one part of #661.

# Rationale for this change
 

See issue. This PR is limited to explicitly calling `close` on written files with error checking.

- Add `close_file()` helper that calls `libc::close()` (Unix) / `CloseHandle` (Windows) directly and propagates errors, instead of relying on `File::drop()`. Note that this uses `unsafe`, like in the platforms' respective `drop()` implementations.
- Apply both to the `put` and `put_multipart` (via `LocalUpload`) code paths.
- Add `libc` (Unix) and `windows-sys` (Windows) as optional dependencies behind the `fs` feature for their respective platforms.
- Add unit test verifying `close_file` detects `EBADF` when the file descriptor is closed behind Rust's back. This is just to validate that `close_file` properly returns the OS error.

Close errors are now detected and propagated as `Error::UnableToCopyDataToFile`. Previously these were silently ignored. This is a correctness improvement but could surface errors that were previously hidden.

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -137,6 +137,32 @@ impl From<Error> for super::Error {
     }
 }
 
+/// Explicitly close a file, checking for errors that would be silently ignored by Rust's `File::drop()`.
+///
+/// On network filesystems (e.g. NFS), `close()` can fail and indicate data loss.
+fn close_file(file: File) -> std::result::Result<(), io::Error> {
+    #[cfg(target_family = "unix")]
+    {
+        nix::unistd::close(file).map_err(|e| e.into())
+    }
+    #[cfg(target_family = "windows")]
+    {
+        use std::os::windows::io::IntoRawHandle;
+
+        let handle = file.into_raw_handle();
+        // SAFETY: `handle` is a valid, owned handle obtained from `into_raw_handle()`.
+        match unsafe { windows_sys::Win32::Foundation::CloseHandle(handle) } {
+            0 => Err(io::Error::last_os_error()),
+            _ => Ok(()),
+        }
+    }
+    #[cfg(not(any(target_family = "unix", target_family = "windows")))]
+    {
+        drop(file);
+        Ok(())
+    }
+}
+
 /// Local filesystem storage providing an [`ObjectStore`] interface to files on
 /// local disk. Can optionally be created with a directory prefix
 ///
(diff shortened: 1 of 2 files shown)


src/local.rs
```diff
@@ -137,6 +137,37 @@ impl From<Error> for super::Error {
...
+fn close_file(file: File) -> std::result::Result<(), io::Error> {
+    #[cfg(target_family = "unix")]
+    {
+        use std::os::unix::io::IntoRawFd;
+        let fd = file.into_raw_fd();
+        // SAFETY: `fd` is a valid, owned file descriptor obtained from `into_raw_fd()`.
+        match unsafe { libc::close(fd) } {
+            0 => Ok(()),
+            _ => Err(io::Error::last_os_error()),
+        }
```
crepererum: Since we already depend on `nix`, could we use that library on Unix instead?

https://docs.rs/nix/latest/nix/unistd/fn.close.html

src/local.rs
```diff
@@ -1913,4 +1956,16 @@ mod unix_test {
...
+    fn test_close_file_detects_error() {
+        use std::os::unix::io::AsRawFd;
+
+        let file = tempfile::tempfile().unwrap();
+        let fd = file.as_raw_fd();
+        // Close the fd behind Rust's back so close_file will get EBADF
+        assert_eq!(unsafe { libc::close(fd) }, 0);
+        let err = super::close_file(file).unwrap_err();
+        assert_eq!(err.raw_os_error(), Some(libc::EBADF));
+    }
```
crepererum: This test tests the new function that you've just added. However I somewhat fail to see what actual error case you're trying to prevent here. The file descriptor isn't exposed to the API user of `object_store`. In #661 you reference https://github.com/pola-rs/polars/issues/21002 but that issue is about polar's own file handling, not about `object_store`. Also you say this is an alternative to `fsync`, but that's not true. Closing the file doesn't give you ANY guarantees with regards to durability. So what concrete order of events leads to a case where the additional error check (via the manual FD close) would help?

src/local.rs
jgiannuzzi: `nix` was only used as a dev dependency which is why I went with `libc` instead, cutting down the middleman (`nix` uses `libc` under the hood). I am happy to use `nix` instead though, as it simplifies this piece of code significantly.

src/local.rs
jgiannuzzi: This test is just validating our logic by emulating a failure case.

The actual failure that can happen is when using a remote file system (e.g. NFS) and suffering from an underlying issue with the mount. In this case, the `close` could fail because some previous queued writes failed and checking the `close` return code allows the calling application to know that something failed, _even if `fsync` wasn't called_.

See this comment on the Polars issue for more details: https://github.com/pola-rs/polars/issues/21002#issuecomment-3579381898

I don't think I claimed this was an alternative to `fsync`. It's rather a way to make sure these errors are reported to the caller. For durability guarantees, `fsync` remains required (and I think this should be part of another PR).

crepererum: Ah, NFS our old friend :grin: . This makes sense, thank you!

src/local.rs
crepererum: I think `nix` is a fine choice here and keeps the dev and prod world kinda in-sync :+1:

src/local.rs
crepererum: BTW: you could also call `close_file` twice and expect that the 2nd call fails. That would make this test here also work on Windows.

jgiannuzzi: Good point! I've updated the Unix test and added a Windows version of it. Sadly I still need to use a raw fd/handle in order to not consume the file and be able to close it twice. Thank you Rust for generally preventing misusing APIs! (That's not sarcasm 😅)

jgiannuzzi:
hey @crepererum, could you please take another look at this?

crepererum (approved):
Looks good, thank you!

PR #683, opened 2026-04-05
fix(memory): fix out-of-order multipart upload for in-memory store

dentiny:

Closes https://github.com/apache/arrow-rs-object-store/issues/682

# Rationale for this change
 
IIRC, I think current implement will shrink parts and lose already-uploaded parts.

This PR resize in-memory store part vector only for increasing capacity.

No

### DIFF
--- a/src/memory.rs
+++ b/src/memory.rs
@@ -433,7 +433,7 @@ impl MultipartStore for InMemory {
     ) -> Result<PartId> {
         let mut storage = self.storage.write();
         let upload = storage.upload_mut(id)?;
-        if part_idx <= upload.parts.len() {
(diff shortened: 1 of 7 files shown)


src/memory.rs
```diff
@@ -611,6 +611,43 @@ mod tests {
         assert_eq!(&*read_data, data);
     }
 
+    #[tokio::test]
```
crepererum: could we make this a generic test in https://github.com/apache/arrow-rs-object-store/blob/main/src/integration.rs ? I think this behavior should be tested for all implementations.

dentiny: sure, moved!

crepererum (approved):
thank you

PR #686, opened 2026-04-08
fix[prefix]: strip_meta from get_opts result in PrefixStore

### LINKED ISSUE #664: PrefixStore::head doesn't strip prefix from meta.location

asubiotto:
PrefixStore was incorrectly not stripping the prefix from the returned result.

Closes #664
Closes #670

# Rationale for this change
 

Fix regression

Fix + assertion in integration tests + regression test

### DIFF
--- a/src/prefix.rs
+++ b/src/prefix.rs
@@ -117,7 +117,9 @@ impl<T: ObjectStore> ObjectStore for PrefixStore<T> {
 
     async fn get_opts(&self, location: &Path, options: GetOptions) -> Result<GetResult> {
         let full_path = self.full_path(location);
-        self.inner.get_opts(&full_path, options).await
+        let mut result = self.inner.get_opts(&full_path, options).await?;
+        result.meta = self.strip_meta(result.meta);
+        Ok(result)
     }
 
     async fn get_ranges(&self, location: &Path, ranges: &[Range<u64>]) -> Result<Vec<Bytes>> {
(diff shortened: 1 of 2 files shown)


asubiotto:
cc @crepererum

src/prefix.rs
```diff
@@ -299,6 +301,26 @@ mod tests {
         assert_eq!(&*read_data, data)
     }
 
+    // Regression test for
```
crepererum, line 304: the extensions of the integration test is actually sufficient, but having an additional regression test also doesn't hurt.

crepererum (approved):
thank you

asubiotto:
Thanks for the review! Can I/how do I request that this gets backported to 0.13?

crepererum:
None of the breaking changes for 0.14.x have landed yet, so backporting is technically not even required. I will however note this in #672 so we don't forget.

PR #689, opened 2026-04-09
fix(core): add missing extension assignment

dentiny:
# Rationale for this change
 
Disclaimer: this issues was discovered by opus when I was leverage LLM to read through the code.

This PR adds missing extension through BufWriter single put shutdown path.
Reading through the codebase, it reads to me "attributes", "tags" and "extensions" should all be assigned for `PutOptions`.

Another example for multipart upload options https://github.com/apache/arrow-rs-object-store/blob/e89f62b6508f6f65873ce6a5017cd0d5d7395184/src/buffered.rs#L341-L344

This PR passes the missing `extension` field to `PutOption`.

No

### DIFF
--- a/src/buffered.rs
+++ b/src/buffered.rs
@@ -446,6 +446,7 @@ impl AsyncWrite for BufWriter {
                     let opts = PutOptions {
                         attributes: self.attributes.take().unwrap_or_default(),
                         tags: self.tags.take().unwrap_or_default(),
+                        extensions: self.extensions.take().unwrap_or_default(),
                         ..Default::default()
                     };
                     let store = Arc::clone(&self.store);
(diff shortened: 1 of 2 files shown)


crepererum:
Is it possible to have a regression test for this?

dentiny:
> Is it possible to have a regression test for this?

Hmmm I think it's not very straightforward; different from attributes, extensions won't appear in the response.

dentiny:
> error: explicit call to `.into_iter()` in function argument accepting `IntoIterator`
   --> src/client/list.rs:120:36
    |
120 |             common_prefixes.extend(response.common_prefixes.into_iter());
    |                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
note: this parameter accepts any `IntoIterator`, so you don't need to call `.into_iter()`

hmmm I might be on an old toolchain version

dentiny:
Hi @crepererum , sorry for the interruption and thanks for the previous reviews, I'm wondering if you mind taking another look maybe when you have some time? Thank you so much!

PR #691, opened 2026-04-10
fix(core): add missing version assignment

dentiny:

This PR adds missing `version` field assignment for `ObjectMetadata`.

No.

### DIFF
--- a/src/prefix.rs
+++ b/src/prefix.rs
@@ -92,11 +92,11 @@ fn strip_meta(prefix: &Path, meta: ObjectMeta) -> ObjectMeta {
         version,
     } = meta;
     ObjectMeta {
-        last_modified: last_modified,
-        size: size,
+        last_modified,
+        size,
         location: strip_prefix(prefix, location),
-        e_tag: e_tag,
-        version: version,
+        e_tag,
+        version,
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/prefix.rs
crepererum, line 94: It would probably help if we destruct `meta` so we don't forget about variables in the future, i.e.:

```rust
let ObjectMeta {
  last_modified,
  size,
  location,
  e_tag,
  version,
} = meta;

ObjectMeta {
  last_modified,
  size,
  location: strip_prefix(...)
  e_tag,
  version,
}
```

dentiny: Yeah that's much better! Thanks for the suggestion! Updated.

crepererum:
I wonder if we could have a regression test for this? :thinking:

dentiny:
> I wonder if we could have a regression test for this? 🤔

I didn't do that initially since it's an easy change, your code suggestion already prevents bugs for future field changes. And unit tests themselves are maintenance overhead. But let me think about it.

dentiny:
Hi @crepererum , sorry for the interruption and thanks for the previous reviews, I'm wondering if you mind taking another look maybe when you have some time? Thank you so much!

src/prefix.rs
```diff
@@ -321,6 +328,26 @@ mod tests {
...
+        let prefix = Path::from("prefix");
+        let meta = ObjectMeta {
+            location: Path::from("prefix/foo"),
+            last_modified: chrono::DateTime::from_timestamp(1_700_000_000, 0).unwrap(),
+            size: 42,
+            e_tag: Some("etag-value".to_string()),
+            version: Some("version-value".to_string()),
+        };
+
+        let stripped = strip_meta(&prefix, meta.clone());
```
crepererum: similar to the actual `strip_meta` function, I would welcome if we future-proof this test a bit so new fields aren't forgotten:


```suggestion
        let ObjectMeta {location, last_modified, size, e_tag, version} = strip_meta(&prefix, meta.clone());
```

dentiny: That sounds great! Updated

crepererum (approved):
thank you

dentiny:
> thank you

Thanks for the review!

PR #694, opened 2026-04-17
fix: clippy

crepererum:
# Which issue does this PR close?
\-

# Rationale for this change
`clippy` broken on `main` after Rust 1.95 release.

# What changes are included in this PR?
Lint fixes.

# Are there any user-facing changes?
\-

PR #696, opened 2026-04-23
fix(azure): fall back to client-side filtering for list_with_offset on OneLake

### LINKED ISSUE #695: MicrosoftAzure::list_with_offset returns empty on OneLake since 0.13.0 (regression from #623)
# `MicrosoftAzure::list_with_offset` returns empty on OneLake since 0.13.0 (regression from #623)

## Describe the bug

Against Microsoft Fabric OneLake (`*.dfs.fabric.microsoft.com`), `ObjectStore::list_with_offset(prefix, offset)` returns **zero** entries even when the prefix contains files strictly greater than `offset`. The equivalent `list(prefix)` on the same store returns the correct files, so the data is reachable — only the offset-based listing is broken.

This regressed in `object_store` 0.13.0 via #623, which replaced the default fallback with an Azure-specific implementation that uses the ADLS Gen2 `startFrom` URI parameter. OneLake's REST surface does not handle `startFrom` the same way the standard ADLS Gen2 endpoint does.

## Impact

Every downstream that uses `list_with_offset` against OneLake is broken on `object_store >= 0.13.0`:

- `delta-kernel-rs` (used by DuckDB's delta extension, delta-rs): loading a Delta table with a `_last_checkpoint` hint fails with `Invalid Checkpoint: Had a _last_checkpoint hint but didn't find any checkpoints`. See [delta-io/delta-kernel-rs#2433](https://github.com/delta-io/delta-kernel-rs/issues/2433) and the (now-closed) workaround attempt [#2437](https://github.com/delta-io/delta-kernel-rs/pull/2437).
- `lakehq/sail` (does NOT use delta-kernel-rs; independently hits the same bug): [lakehq/sail#1730](https://github.com/lakehq/sail/issues/1730).

## To Reproduce

Minimal, no-delta-kernel reproducer below. The only thing swapped between the two runs is the `object_store` pin.

`Cargo.toml`:
```toml
[package]
name = "onelake-repro"
version = "0.0.1"
edition = "2021"

[dependencies]
# Swap between "=0.12.5" (works) and "=0.13.2" (broken)
object_store = { version = "=0.13.2", features = ["azure"] }
futures = "0.3"
tokio = { version = "1", features = ["rt-multi-thread", "macros"] }
url = "2"
anyhow = "1"
```

`src/main.rs`:
```rust
use std::env;

use anyhow::{anyhow, Context, Result};
use futures::stream::StreamExt;
use object_store::azure::{AzureConfigKey, MicrosoftAzureBuilder};
use object_store::path::Path;
use object_store::{ObjectMeta, ObjectStore};

#[tokio::main(flavor = "multi_thread", worker_threads = 2)]
async fn main() -> Result<()> {
    let args: Vec<String> = env::args().collect();
    if args.len() != 5 {
        return Err(anyhow!("usage: onelake-repro <workspace> <lakehouse> <table> <checkpoint_version>"));
    }
    let workspace = &args[1];
    let lakehouse = &args[2];
    let table = &args[3];
    let ckpt_version: u64 = args[4].parse()?;

    let token = env::var("AZURE_STORAGE_TOKEN")
        .context("AZURE_STORAGE_TOKEN not set")?;

    let url = format!(
        "abfss://{workspace}@onelake.dfs.fabric.microsoft.com/{lakehouse}.Lakehouse/Tables/{table}/"
    );

    let store = MicrosoftAzureBuilder::new()
        .with_url(url.as_str())
        .with_config(AzureConfigKey::Token, token)
        .build()?;

    let prefix_str = format!("{lakehouse}.Lakehouse/Tables/{table}/_delta_log");
    let prefix = Path::from(prefix_str.as_str());
    let offset = Path::from(format!("{prefix_str}/{ckpt_version:020}").as_str());

    let a = collect(store.list(Some(&prefix))).await?;
    println!("A) list(prefix): {} entries", a.len());
    for loc in &a { println!("   {loc}"); }

    let b = collect(store.list_with_offset(Some(&prefix), &offset)).await?;
    println!("\nB) list_with_offset(prefix, offset): {} entries", b.len());
    for loc in &b { println!("   {loc}"); }

    Ok(())
}

async fn collect<S>(mut s: S) -> Result<Vec<String>>
where S: futures::Stream<Item = object_store::Result<ObjectMeta>> + Unpin
{
    let mut out = vec![];
    while let Some(m) = s.next().await { out.push(m?.location.to_string()); }
    out.sort();
    Ok(out)
}
```

Run:
```bash
export AZURE_STORAGE_TOKEN=$(az account get-access-token --resource https://storage.azure.com/ --query accessToken -o tsv)
cargo run --release -- <workspace> <lakehouse> <table> <checkpoint_version>
```

## Expected behavior

`list_with_offset(prefix, offset)` should return exactly the files in `list(prefix)` whose location is lexicographically greater than `offset`.

## Actual behavior

Against the same OneLake table (a Delta table with `_last_checkpoint` at v10):

**With `object_store = "=0.12.5"`** (works):
```
A) list(prefix): 11 entries
   _delta_log/00000000000000000005.json
   _delta_log/00000000000000000006.json
   _delta_log/00000000000000000007.json
   _delta_log/00000000000000000008.json
   _delta_log/00000000000000000009.json
   _delta_log/00000000000000000010.checkpoint.parquet
   _delta_log/00000000000000000010.json
   _delta_log/00000000000000000011.json
   _delta_log/00000000000000000012.json
   _delta_log/00000000000000000013.json
   _delta_log/_last_checkpoint

B) list_with_offset(prefix, _delta_log/00000000000000000010): 6 entries
   _delta_log/00000000000000000010.checkpoint.parquet
   _delta_log/00000000000000000010.json
   _delta_log/00000000000000000011.json
   _delta_log/00000000000000000012.json
   _delta_log/00000000000000000013.json
   _delta_log/_last_checkpoint
```

**With `object_store = "=0.13.2"`** (broken):
```
A) list(prefix): 11 entries       <-- identical to above
   ...

B) list_with_offset(prefix, _delta_log/00000000000000000010): 0 entries
```

(Only `list_with_offset` differs between the two runs.)

## Suspected cause

[#623](https://github.com/apache/arrow-rs-object-store/pull/623) added a direct `list_with_offset` implementation for Azure that sends `startFrom=<offset>` per the [ADLS Gen2 list-blobs API](https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?view=rest-storageservices-datalakestoragegen2-2019-12-12&tabs=microsoft-entra-id#uri-parameters). OneLake's endpoint apparently does not implement `startFrom` compatibly — it returns an empty list regardless of the offset value.

This matches `lonless9`'s analysis on [lakehq/sail#1730](https://github.com/lakehq/sail/issues/1730) and the related [Azurite#2619](https://github.com/Azure/Azurite/issues/2619#issuecomment-3660701055).

## Environment

- `object_store` 0.13.2 (and 0.13.0, 0.13.1 — all contain #623)
- OneLake endpoint `onelake.dfs.fabric.microsoft.com`
- Service-principal / Azure CLI bearer token (same auth in both runs; auth is not the issue)
- Observed on Windows 11 / rustc 1.95.0, but not platform-dependent

---

Repro and report co-drafted with [Claude Code](https://claude.com/claude-code) (Claude Opus 4.7).

kevinjqliu:

Closes #695.

# Rationale for this change
 

`list_with_offset` returns empty results on OneLake when **both** of the following conditions are met:

1. **Friendly-name URLs** — the endpoint addresses the workspace and lakehouse by display name (e.g. `onelake.blob.fabric.microsoft.com/MyWorkspace/lakehouse.Lakehouse/...`) rather than by GUID
2. **`startFrom` query parameter** — the `List Blobs` request includes `startFrom`, which OneLake silently ignores in this case, returning 200 OK with zero results rather than rejecting the request

(Note that when using GUID-based URLs (`onelake.blob.fabric.microsoft.com/{workspace-guid}`), `startFrom` works correctly.)

This regression was introduced in #623 which implemented optimizations using `startFrom` pushdown.

We (The OneLake team) are actively fixing `startFrom` with friendly-name URLs on our side. 
In the meantime, `object_store` can fall back to client-side filtering for OneLake endpoints.
I created a tracking issue (#697) to re-enable this optimization for OneLake once the fix is complete

- Auto-detect Fabric/OneLake endpoints (`*.fabric.microsoft.com`) and skip `startFrom`, falling back to client-side filtering -- same approach already used for Azurite
- Add `#[ignore]` integration test (`test_onelake_list_with_offset`) that verifies offset exclusivity, boundary cases, and ordering against a live OneLake endpoint

### Tested locally
GUID-based URI works:
```
export AZURE_STORAGE_TOKEN=$(az account get-access-token --resource https://storage.azure.com/ --query accessToken -o tsv) && \
ONELAKE_URL="https://msit-onelake.blob.fabric.microsoft.com/OLSTeamWorkspace/lh.Lakehouse/Files/test_startfrom" \
cargo test --features azure test_onelake_list_with_offset -- --ignored --no-capture
```

Friendly name based URI fails:
```
export AZURE_STORAGE_TOKEN=$(az account get-access-token --resource https://storage.azure.com/ --query accessToken -o tsv) && \
ONELAKE_URL="https://msit-onelake.blob.fabric.microsoft.com/45a753bc-c074-42cf-8b30-5dfa920b241f/3e56af9e-3832-47ed-b18c-dcc58b562e87/Files/test_startfrom" \
cargo test --features azure test_onelake_list_with_offset -- --ignored --no-capture
```

No, this is a behavioral change for `list_with_offset` against OneLake endpoints

### DIFF
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -127,11 +127,20 @@ impl ObjectStore for MicrosoftAzure {
         prefix: Option<&Path>,
         offset: &Path,
     ) -> BoxStream<'static, Result<ObjectMeta>> {
-        if self.client.config().is_emulator {
-            // Azurite doesn't support the startFrom query parameter,
+        let disable_start_from = self.client.config().is_emulator
+            || self
+                .client
+                .config()
+                .service
+                .host_str()
+                .is_some_and(|h| h.ends_with(".fabric.microsoft.com"));
+
+        if disable_start_from {
+            // Azurite and OneLake don't support the startFrom query parameter,
             // fall back to client-side filtering
             //
             // See https://github.com/Azure/Azurite/issues/2619#issuecomment-3660701055
+            // See https://github.com/apache/arrow-rs-object-store/issues/695
             let offset = offset.clone();
             self.list(prefix)
                 .try_filter(move |f| futures_util::future::ready(f.location > offset))
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -419,6 +419,114 @@ mod tests {
         assert_eq!(data, loaded);
     }
 
+    /// Verifies that `list_with_offset` works against OneLake (Fabric) endpoints.
+    ///
+    /// OneLake silently ignores the `startFrom` query parameter, so the client
+    /// must fall back to client-side filtering.
+    ///
+    /// Set these env vars before running:
+    /// - `AZURE_STORAGE_TOKEN`: bearer token (e.g. from `az account get-access-token`)
+    /// - `ONELAKE_URL`: full OneLake table URL, e.g.
+    ///   `https://onelake.dfs.fabric.microsoft.com/<ws>/<item>/Tables/<table>`
+    ///
+    /// See <https://github.com/apache/arrow-rs-object-store/issues/695>
+    #[ignore = "Used for manual testing against a real OneLake endpoint."]
+    #[tokio::test]
+    async fn test_onelake_list_with_offset() {
+        let url = std::env::var("ONELAKE_URL").unwrap();
+        let token = std::env::var("AZURE_STORAGE_TOKEN").unwrap();
+
+        let store = MicrosoftAzureBuilder::new()
+            .with_url(&url)
+            .with_config(AzureConfigKey::Token, token)
+            .build()
+            .unwrap();
+
+        // Derive a writable path prefix from the URL
+        // (skip workspace segment which becomes the container)
+        let parsed: Url = url.parse().unwrap();
+        let mut segments = parsed.path_segments().unwrap();
+        let _workspace = segments.next().unwrap();
+        let base: String = segments.collect::<Vec<_>>().join("/");
+        let test_dir = format!("{base}/test_onelake_offset");
+
+        // Create test files with predictable ordering
+        let prefix = Path::from(test_dir.as_str());
+        let files: Vec<Path> = (b'a'..=b'e')
+            .map(|c| Path::from(format!("{test_dir}/file_{}.txt", c as char)))
+            .collect();
+        let data = Bytes::from("test data");
+        for file in &files {
+            store.put(file, data.clone().into()).await.unwrap();
+        }
+
+        // Test 1: Offset at file_b → should return c, d, e (not b)
+        let offset = Path::from(format!("{test_dir}/file_b.txt"));
+        let result: Vec<Path> = store
+            .list_with_offset(Some(&prefix), &offset)
+            .map_ok(|m| m.location)
+            .try_collect()
+            .await
+            .unwrap();
+        assert!(
+            !result.contains(&offset),
+            "offset file_b should be excluded, got: {result:?}"
+        );
+        assert_eq!(result.len(), 3, "expected c/d/e after file_b, got: {result:?}");
+
+        // Test 2: Offset at file_a → should return b, c, d, e
+        let offset = Path::from(format!("{test_dir}/file_a.txt"));
+        let result: Vec<Path> = store
+            .list_with_offset(Some(&prefix), &offset)
+            .map_ok(|m| m.location)
+            .try_collect()
+            .await
+            .unwrap();
+        assert!(!result.contains(&offset), "offset file_a should be excluded");
+        assert_eq!(result.len(), 4, "expected b/c/d/e after file_a, got: {result:?}");
+
+        // Test 3: Offset at file_e (last) → should return empty
+        let offset = Path::from(format!("{test_dir}/file_e.txt"));
+        let result: Vec<Path> = store
+            .list_with_offset(Some(&prefix), &offset)
+            .map_ok(|m| m.location)
+            .try_collect()
+            .await
+            .unwrap();
+        assert!(result.is_empty(), "offset at last file should return empty, got: {result:?}");
+
+        // Test 4: Offset before all files → should return all 5
+        let offset = Path::from(format!("{test_dir}/file"));
+        let result: Vec<Path> = store
+            .list_with_offset(Some(&prefix), &offset)
+            .map_ok(|m| m.location)
+            .try_collect()
+            .await
+            .unwrap();
+        assert_eq!(result.len(), 5, "offset before all files should return all, got: {result:?}");
+
+        // Test 5: Every returned entry is strictly greater than offset
+        let offset = Path::from(format!("{test_dir}/file_c.txt"));
+        let result: Vec<ObjectMeta> = store
+            .list_with_offset(Some(&prefix), &offset)
+            .try_collect()
+            .await
+            .unwrap();
+        for meta in &result {
+            assert!(
+                meta.location > offset,
+                "entry {} should be > offset {}",
+                meta.location,
+                offset
+            );
+        }
+
+        // Cleanup
+        for file in &files {
+            let _ = store.delete(file).await;
+        }
+    }
+
     #[test]
     fn azure_test_config_get_value() {
         let azure_client_id = "object_store:fake_access_key_id".to_string();
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -421,13 +421,15 @@ mod tests {
 
     /// Verifies that `list_with_offset` works against OneLake (Fabric) endpoints.
     ///
-    /// OneLake silently ignores the `startFrom` query parameter, so the client
-    /// must fall back to client-side filtering.
+    /// OneLake silently ignores the `startFrom` query parameter when using
+    /// friendly-name URLs (e.g. `.../MyWorkspace/lakehouse.Lakehouse/...`),
+    /// returning 200 OK with zero results.
+    /// GUID-based URLs handle `startFrom` correctly.
     ///
     /// Set these env vars before running:
     /// - `AZURE_STORAGE_TOKEN`: bearer token (e.g. from `az account get-access-token`)
-    /// - `ONELAKE_URL`: full OneLake URL, e.g.
-    ///   `https://onelake.blob.fabric.microsoft.com/<ws>/<item>/Files/`
+    /// - `ONELAKE_URL`: full OneLake URL with friendly names, e.g.
+    ///   `https://onelake.blob.fabric.microsoft.com/<workspace>/<item>.Lakehouse/`
     ///
     /// See <https://github.com/apache/arrow-rs-object-store/issues/695>
     #[ignore = "Used for manual testing against a real OneLake endpoint."]
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -472,7 +472,11 @@ mod tests {
             !result.contains(&offset),
             "offset file_b should be excluded, got: {result:?}"
         );
-        assert_eq!(result.len(), 3, "expected c/d/e after file_b, got: {result:?}");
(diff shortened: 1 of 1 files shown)


kevinjqliu:
Need to rebase #694 to pass CI

kevinjqliu:
@tustvold could you take a look?

rtyler (approved):

kevinjqliu:
rebased to pull in #694

rtyler (approved):
The only suggestion that I have is to provide a true listing of the OneLake location used for the manual testing in case anybody motivated enough to come along and try that manual test later needs to 😄 

Is there a publicly linkable reference to the bug witht he OneLake service that can be referenced as well?

I am cautiously optimistic that this could get into a 0.13.3 so some downstream breaks can be resolved ahead of 0.14

tustvold:
Apologies for the delay on getting to this, been absolutely swamped at work, is this still required or has onelake added support for this?

shehabgamin:
> is this still required or has onelake added support for this?

@tustvold It's still required for OneLake!

alamb:
@crepererum and I are working on on a release -- I put this PR on the list and will try and move it along

kevinjqliu:
I verified that this is fixed and deployed on the OneLake side. So this PR is no longer needed; we want users to take advantage of `startFrom` whenever possible

shehabgamin:
> I verified that this is fixed and deployed on the OneLake side. So this PR is no longer needed; we want users to take advantage of `startFrom` whenever possible

@kevinjqliu Delta reads still fail (I last tested on Sunday), were you perhaps testing with an unreleased version of Object Store?

Or, was there a very recent deploy/fix in OneLake within the last day?

EDIT:
Confirmed Delta reads are still broken as of today. Delta writes work though.

kevinjqliu:
The fix is getting rolled out to production environments. It might not be there yet for some regions. Give it another try in a few days. I'll follow up next week and test it again.

kevinjqliu:
@shehabgamin i checked internally, this is now deployed

shehabgamin:
> @shehabgamin i checked internally, this is now deployed

@kevinjqliu Still doesn't work for me. Here's a reproducible example:
```
>>> import os
... os.environ["SAIL_CATALOG__LIST"] = (
...     '[{type="onelake", name="ls-onelake", '
...     'url="SOME_ONELAKE_URL/LakeSail.Lakehouse", '
...     'bearer_token="' + BEARER_TOKEN + '"}]'
... )
... os.environ["AZURE_STORAGE_TOKEN"] = BEARER_TOKEN
...
>>> from pysail.spark import SparkConnectServer
... from pyspark.sql import SparkSession
... server = SparkConnectServer(); server.start()
... _, port = server.listening_address
... spark = SparkSession.builder.remote(f"sc://localhost:{port}").getOrCreate()
...
[2026-06-23T21:41:59Z INFO sail_python::spark::server] Starting the Spark Connect server on 127.0.0.1:49728...
>>> spark.sql("SHOW NAMESPACES").show()
[2026-06-23T21:42:04Z INFO sail_session::session_manager::actor::handler] creating session bef5cd8d-c4fe-4bac-bf25-c98e7233c9dd

+----+----------+-----------+-----------+
|name|   catalog|description|locationUri|
+----+----------+-----------+-----------+
| dbo|ls-onelake|       NULL|       NULL|
+----+----------+-----------+-----------+

>>>
>>> spark.sql("SHOW TABLES IN dbo").show()

+--------+--------------------+-----------+
|database|           tableName|isTemporary|
+--------+--------------------+-----------+
|     dbo|iceberg_dim_products|      false|
|     dbo|  delta_dim_products|      false|
+--------+--------------------+-----------+

>>>
>>> df = spark.read.format("csv").option("header", "true").load(
...     "abfss://SOME_ONELAKE_URL@onelake.dfs.fabric.microsoft.com/"
...     "LakeSail.Lakehouse/Files/Dim_Products.csv"
... )
>>>
>>> df.write.format("delta").save(
...     "abfss://SOME_ONELAKE_URL@onelake.dfs.fabric.microsoft.com/"
...     "LakeSail.Lakehouse/Tables/dbo/delta_dim_products_test"
... )

[2026-06-23T21:42:46Z WARN sail_delta_lake::kernel::transaction] Post-commit: failed to load state for version 0 (post-commit activities skipped): Missing commit file: expected final version 0, replay reached -1
>>>
>>> spark.sql("SHOW TABLES IN dbo").show()
+--------+--------------------+-----------+
|database|           tableName|isTemporary|
+--------+--------------------+-----------+
|     dbo|delta_dim_product...|      false|
|     dbo|iceberg_dim_products|      false|
|     dbo|  delta_dim_products|      false|
+--------+--------------------+-----------+

>>> spark.read.format("delta").load(
...     "abfss://SOME_ONELAKE_URL@onelake.dfs.fabric.microsoft.com/"
...     "LakeSail.Lakehouse/Tables/dbo/delta_dim_products_test"
... ).show(5)
Traceback (most recent call last):
... [TRUNCATED]
pyspark.errors.exceptions.connect.AnalysisException: Invalid table location: No commit files found in _delta_log
>>>
```

maartenhubrechts-anb:
Just hit this too, so adding another data point. Still broken for me as of 2026-06-30, different tenant from LakeSail, North Europe region.

`deltalake` 1.6.0 (object_store Azure backend) fails with `Generic delta kernel error: No files in log segment`. I put a TLS-intercepting proxy in front of it to see what it actually sends. The log listing request is this:

```
GET /<workspace>?restype=container&comp=list
    &prefix=<lakehouse>/Tables/<schema>/<table>/_delta_log/
    &startFrom=<lakehouse>/Tables/<schema>/<table>/_delta_log/00000000000000000000
```

I replayed that same List Blobs call by hand with the same token, only changing `startFrom`:

| `startFrom` | blobs returned |
|---|---|
| (omitted) | `…/_delta_log/00000000000000000000.json`, `…0001.json` |
| `…/_delta_log/00000000000000000000` (what object_store sends) | none |
| `…/_delta_log/` (the prefix itself) | none |

So any non-empty `startFrom` comes back empty, even values that sort before the actual files (`…00000000000000000000` is less than `…00000000000000000000.json`, so those should still be returned). And the empty response is HTTP 200 with `<Blobs />` and an empty `<NextMarker />`, so it's a final page, not a pagination thing that object_store is failing to follow.

Same result on two table types: a Lakehouse schema-shortcut table (GUID abfss URL) and a Warehouse table (friendly-name abfss URL).

So whatever got deployed, `startFrom` on `_delta_log` is still returning nothing here. Can grab more captures if that helps.


Update 2026-07-03: Further narrowed this down. Same code, same auth, both GUID abfss URLs:
- native Lakehouse table → reads fine
- schema-shortcut table → No files in log segment

Also reproduced independently of delta-rs with DuckDB's delta extension (which uses delta-kernel-rs / object_store as well).

kevinjqliu:
Thank you both for verifying again. Seems like there was a deployment issue and we had to roll back. I will update again once the new deployment is complete

kevinjqliu:
Thank you both for engaging with this @shehabgamin @maartenhubrechts-anb 

The deployment has been rolled out. I have verified the issue has been resolved, using the lakesail repro steps above (thanks!)
I also tried with schema shortcut from warehouse, everything is ✅ 

Please let me know if you see otherwise. Im happy to help investigate

alamb:
Thank you for keeping it real @kevinjqliu  -- so much going on to keepup with

PR #700, opened 2026-04-25
fix: fix incorrect splitting with line delimited streaming

### LINKED ISSUE #650: newline_delimited_stream incorrect processing `Generic { store: "LineDelimiter", source: UnterminatedString }` on valid CSVs (impacts datafusion)
**Describe the bug**

We've had issue with datafusion schema inferrence which uses newline_delimited_stream to chunk the input stream.

It manifested as `Generic { store: "LineDelimiter", source: UnterminatedString }` errors on what was a valid CSV file (csv parser would be fine, but datafusion would fail due to the incorrect chunking).

**To Reproduce**

Minimal testcases that can be added to deliminated.rs:
```
   #[tokio::test]
    async fn test_delimiter_quotes_stream() {
        let input = vec!["x,y,z\n,\"new\nline\",\"with ", "space\""];
        let input_stream =
            futures_util::stream::iter(input.into_iter().map(|s| Ok(Bytes::from(s))));
        let stream = newline_delimited_stream(input_stream);

        let results: Vec<_> = stream.try_collect().await.unwrap();
        assert_eq!(
            results,
            vec![
                Bytes::from("x,y,z\n"),
                Bytes::from("\"new\nline\",\"with space\"")
            ]
        )
    }

    #[tokio::test]
    async fn test_delimiter_escape_stream() {
        let input = vec!["hello\n\n\"\\ttabulated\"", "world"];
        let input_stream =
            futures_util::stream::iter(input.into_iter().map(|s| Ok(Bytes::from(s))));
        let stream = newline_delimited_stream(input_stream);

        let results: Vec<_> = stream.try_collect().await.unwrap();
        assert_eq!(
            results,
            vec![
                Bytes::from("hello\n"),
                Bytes::from("\"\ttabulated\"\"world\n")
            ]
        )
    }

```

**Additional context**

Both bugs have the same root cause: `records_end` is a double ended iterator, so when we call `next_back()`, the escaping/quote logic is run in reverse instead of forward, and the internal state during the filter_map gets corrupted.

One obvious fix is to materialize all the split points into a Vec, e.g.:
```
        let mut record_ends = val.iter().enumerate().filter_map(|(idx, v)| {
            ...
        }).collect::<Vec<_>>();
```

bboissin:
In some cases, valid CSV in datafusion would return: `Generic { store: "LineDelimiter", source: UnterminatedString }` due to incorrect logic.

records_ends is a double ended iterator, so when calling next_back() the quoting/escaping logic would run in reverse, corrupting the internal state.

Closes #650

### DIFF
--- a/src/delimited.rs
+++ b/src/delimited.rs
@@ -77,22 +77,28 @@ impl LineDelimiter {
 
         let is_escape = &mut self.is_escape;
         let is_quote = &mut self.is_quote;
-        let mut record_ends = val.iter().enumerate().filter_map(|(idx, v)| {
-            if *is_escape {
-                *is_escape = false;
-                None
-            } else if *v == ESCAPE {
-                *is_escape = true;
-                None
-            } else if *v == QUOTE {
-                *is_quote = !*is_quote;
-                None
-            } else if *is_quote {
-                None
-            } else {
-                (*v == NEWLINE).then_some(idx + 1)
-            }
-        });
+        let mut record_ends = val
+            .iter()
+            .enumerate()
+            .filter_map(|(idx, v)| {
+                if *is_escape {
+                    *is_escape = false;
+                    None
+                } else if *v == ESCAPE {
+                    *is_escape = true;
+                    None
+                } else if *v == QUOTE {
+                    *is_quote = !*is_quote;
+                    None
+                } else if *is_quote {
+                    None
+                } else {
+                    (*v == NEWLINE).then_some(idx + 1)
+                }
+            })
+            // Materialize the split points, records_ends is a double ended iterator, so when calling next_back() the quoting/escaping logic would run in reverse, corrupting the internal state.
+            .collect::<Vec<_>>()
+            .into_iter();
 
         let start_offset = match self.remainder.is_empty() {
             true => 0,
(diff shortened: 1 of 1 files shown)


src/delimited.rs
```diff
@@ -77,22 +77,27 @@ impl LineDelimiter {
...
+                    *is_quote = !*is_quote;
+                    None
+                } else if *is_quote {
+                    None
+                } else {
+                    (*v == NEWLINE).then_some(idx + 1)
+                }
+            })
+            .collect::<Vec<_>>()
+            .into_iter();
```
crepererum:
```suggestion
            // records_ends is a double ended iterator, so when calling next_back() the quoting/escaping logic would run in reverse, corrupting the internal state.
            .collect::<Vec<_>>()
            .into_iter();
```

(might need an additional line break, just copied your explanation from the PR description)

It took me a while to understand why `collect().into_iter()` would be a sensible choice here (because usually it isn't), so adding a code comment will likely help a future reader.

crepererum (approved):

tustvold (approved):
Thank you for this, added a small tweak to use .last instead of collecting into a vec

Issue #657, opened 2026-03-11
Release object store `0.14.0` (breaking) - Target April 2026


crepererum:
Next breaking release after `0.13.2` (#393).

# Breaking Changes
- [ ] #633
- [ ] #585

crepererum:
Kinda depends on the two major PRs that we wanna get in, but from my PoV that should be possible.

bendu:
Thank you for the update @crepererum Please let me know if there's anything I can do to help.

alamb:
Given we are nwo past April (and into June) are we planning on this one? It looks like 
- https://github.com/apache/arrow-rs-object-store/pull/633

Was merged with API changes so we'll have to do a 0.14 release (rather than 0.13.x)

alamb:
It would be great to also get an update to reqwest in as well, for exmaple
- https://github.com/apache/arrow-rs-object-store/pull/707

kevinjqliu:
Thanks for starting this @alamb 

I have 2 related PRs for OneLake (an ADLS compatible storage product) 
- https://github.com/apache/arrow-rs-object-store/pull/696
- https://github.com/apache/arrow-rs-object-store/pull/711

both are well-tested and changes isolated to OneLake only

alamb:
I added to the list above

kevinjqliu:
im going through the PRs above and some that are marked ready for review. please lmk if theres anything else i can help with

kevinjqliu:
Flagging **Makes reqwest optional** (#724), its similar to **Pluggable Crypto / Update reqwest** 0.13 (#707)

alamb:
Also added the feature flag discussion to the release check list (well really its resolution)
- https://github.com/apache/arrow-rs-object-store/issues/744

alamb:
BTW I am out tomorrow (Fri June 12, 2026) but will be back after that and help drive this release next week

alamb:
Update:
- I just reviewed / approved https://github.com/apache/arrow-rs-object-store/pull/643 from @Barre 
- https://github.com/apache/arrow-rs-object-store/pull/707 is ready for a final review and then we can merge and cut the release

alamb:
I'll plan to merge both PRs and prepare a release candidate tomorrow

kevinjqliu:
would be great to include https://github.com/apache/arrow-rs-object-store/pull/742 as well

kevinjqliu:
[...]

PR #707, opened 2026-04-28
Pluggable Crypto / Update reqwest 0.13

### LINKED ISSUE #413: Change crypto provider from ring to aws-rust-lc

goffrie:
NOTE: this is a rebased version of #585

- Closes #413.

# Rationale for this change
 

Adds the ability to compile object_store without a dependency on ring, and instead use aws-lc-rs or a user-provided crypto implementation. Also changes the default to aws-lc-rs, which the ecosystem seems to be moving toward as a default.

Also upgrades `reqwest` to 0.13, and makes it possible to configure reqwest with any TLS backend by selecting a `-no-crypto` feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,10 @@ jobs:
       - uses: actions/checkout@v6
       - name: Setup Clippy
         run: rustup component add clippy
+      - name: Check no crypto crates
+        run:
+          cargo tree --features gcp-no-crypto,aws-no-crypto,azure-no-crypto,http-no-crypto \
+          | grep -qE '\b(ring|openssl)\b' && { echo "❌ disallowed crate found"; exit 1; } || echo "✅ no disallowed crates"
       # Run different tests for the library on its own as well as
       # all targets to ensure that it still works in the absence of
       # features that might be enabled by dev-dependencies of other
@@ -49,12 +53,20 @@ jobs:
         run: cargo clippy --no-default-features -- -D warnings
       - name: Run clippy with fs features
         run: cargo clippy --no-default-features --features fs -- -D warnings
+      - name: Run clippy with aws-no-crypto feature
+        run: cargo clippy --features aws-no-crypto -- -D warnings
       - name: Run clippy with aws feature
         run: cargo clippy --features aws -- -D warnings
+      - name: Run clippy with gcp-no-crypto feature
+        run: cargo clippy --features gcp-no-crypto -- -D warnings
       - name: Run clippy with gcp feature
         run: cargo clippy --features gcp -- -D warnings
+      - name: Run clippy with azure-no-crypto feature
+        run: cargo clippy --features azure-no-crypto -- -D warnings
       - name: Run clippy with azure feature
         run: cargo clippy --features azure -- -D warnings
+      - name: Run clippy with http-no-crypto feature
+        run: cargo clippy --features http-no-crypto -- -D warnings
       - name: Run clippy with http feature
         run: cargo clippy --features http -- -D warnings
       - name: Run clippy with integration feature
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,9 +70,9 @@ jobs:
       - name: Run clippy with http-base feature
         run: cargo clippy --no-default-features --features http-base -- -D warnings
       - name: Run clippy with base features, reqwest and ring crypto
-        run: cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest,reqwest/rustls,ring -- -D warnings
+        run: cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest,ring -- -D warnings
       - name: Run clippy with base features and reqwest but no bundled crypto
-        run: cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest,reqwest/rustls -- -D warnings
+        run: cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest -- -D warnings
       - name: Verify base features do not enable reqwest
         run: |
           if cargo tree \
@@ -188,6 +188,12 @@ jobs:
           AWS_CONDITIONAL_PUT: etag
           AWS_COPY_IF_NOT_EXISTS: multipart
 
+      # Exercise the `ring` crypto provider at runtime (the other jobs only
+      # compile it; the default `aws`/etc. features use `aws-lc-rs`).
+      # object_store's signing uses ring here (the `aws-lc-rs` feature is off);
+      - name: Run object_store tests with ring crypto provider
+        run: cargo test --lib --tests --no-default-features --features fs,aws-base,azure-base,gcp-base,http-base,reqwest,ring
+
       - name: GCS Output
         if: ${{ !cancelled() }}
         run: docker logs $GCS_CONTAINER
@@ -220,6 +226,10 @@ jobs:
         run: rustup target add wasm32-unknown-unknown
       - name: Build wasm32-unknown-unknown
         run: cargo build --target wasm32-unknown-unknown
+      - name: Build wasm32-unknown-unknown with ring crypto provider
+        # aws-lc-rs does not build for wasm32, so use the ring crypto provider here.
+        # Includes aws/gcp/azure-base so the RingCryptoProvider is compiled.
+        run: cargo build --no-default-features --features aws-base,gcp-base,azure-base,http-base,reqwest,ring --target wasm32-unknown-unknown
       - name: Install wasm32-wasip1
         run: rustup target add wasm32-wasip1
       - name: Use reqwest 0.13.2 for wasm32-wasip1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,25 +39,35 @@ jobs:
       - uses: actions/checkout@v6
       - name: Setup Clippy
         run: rustup component add clippy
-      - name: Check no crypto crates
+      - name: Verify base features do not enable TLS/signing crypto crates
         # `--prefix none -f '{p}'` prints one `name vX.Y.Z` per line, so the
         # crate name can be anchored with `^`. Avoid `\b`, which is not
         # portable (e.g. BSD grep) and would also substring-match crates like
         # `iri-string`.
         run: |
-          found=$(cargo tree --no-default-features --features gcp-base,aws-base,azure-base,http-base --prefix none -f '{p}' \
-            | grep -E '^(ring|openssl|aws-lc-rs) v' | sort -u || true)
+          found=$(cargo tree --no-default-features --features aws-base,azure-base,gcp-base,http-base --edges normal --prefix none -f '{p}' \
+            | grep -E '^(aws-lc-rs|aws-lc-sys|native-tls|openssl|ring|rustls) v' | sort -u || true)
           if [ -n "$found" ]; then
             echo "❌ disallowed crate(s) found:"
             echo "$found"
             exit 1
           fi
-          echo "✅ no ring/openssl/aws-lc-rs crates"
+          echo "✅ no TLS/signing crypto crates"
+      - name: Verify base features do not enable reqwest
+        run: |
+          found=$(cargo tree --no-default-features --features aws-base,azure-base,gcp-base,http-base --edges normal --prefix none -f '{p}' \
+            | grep -E '^reqwest v' | sort -u || true)
+          if [ -n "$found" ]; then
+            echo "❌ reqwest must not be enabled by *-base features:"
+            echo "$found"
+            exit 1
+          fi
+          echo "✅ *-base features do not enable reqwest"
       - name: Check ring build does not pull aws-lc-rs
         # A `ring` build (with a non-aws-lc-rs reqwest TLS backend) must be free
         # of aws-lc-rs entirely, so users can opt out of it on every target.
         run: |
-          found=$(cargo tree --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest,reqwest/native-tls,ring --prefix none -f '{p}' \
+          found=$(cargo tree --no-default-features --features aws-base,azure-base,gcp-base,http-base,reqwest,reqwest/native-tls,ring --edges normal --prefix none -f '{p}' \
             | grep -E '^(aws-lc-rs|aws-lc-sys) v' | sort -u || true)
           if [ -n "$found" ]; then
             echo "❌ aws-lc-rs pulled into a ring-only build:"
@@ -65,6 +75,19 @@ jobs:
             exit 1
           fi
           echo "✅ ring build is free of aws-lc-rs"
+      - name: Check batteries-included features do not pull ring
+        # The batteries-included features use reqwest/rustls, which uses aws-lc-rs
+        # for TLS by default. Keep object_store's default bundled crypto provider
+        # aligned with that choice by ensuring these features don't pull ring.
+        run: |
+          found=$(cargo tree --no-default-features --features aws,azure,gcp,http --edges normal --prefix none -f '{p}' \
+            | grep -E '^ring v' | sort -u || true)
+          if [ -n "$found" ]; then
+            echo "❌ batteries-included features pulled ring:"
+            echo "$found"
+            exit 1
+          fi
+          echo "✅ batteries-included features are free of ring"
       # Run different tests for the library on its own as well as
       # all targets to ensure that it still works in the absence of
       # features that might be enabled by dev-dependencies of other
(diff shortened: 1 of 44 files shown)


djc:
Ping? Would be nice to avoid having duplicate reqwest versions in downstream projects.

goffrie:
I've given this another rebase, I think it's more blocked on review/maintainer capacity though?

goffrie:
(cc @creperum since you were asking for a rebase in #585)

djc:
Maintainers, could someone take a look? Would be nice to avoid duplicate dependencies in downstream applications that are up to date with the rest of the ecosystem (and it's been 5 months since reqwest 0.13 was released).

.github/workflows/ci.yml
```diff
@@ -39,6 +39,10 @@ jobs:
       - uses: actions/checkout@v6
       - name: Setup Clippy
         run: rustup component add clippy
+      - name: Check no crypto crates
+        run:
+          cargo tree --features gcp-no-crypto,aws-no-crypto,azure-no-crypto,http-no-crypto \
+          | grep -qE '\b(ring|openssl)\b' && { echo "❌ disallowed crate found"; exit 1; } || echo "✅ no disallowed crates"
```
Tpt: I guess this test should also add aws-lc-rs?
```suggestion
          | grep -qE '\b(ring|openssl|aws-lc-rs)\b' && { echo "❌ disallowed crate found"; exit 1; } || echo "✅ no disallowed crates"
```

src/client/mod.rs
```diff
@@ -260,7 +301,7 @@ impl FromStr for ClientConfigKey {
             "timeout" => Ok(Self::Timeout),
             "user_agent" => Ok(Self::UserAgent),
             _ => Err(super::Error::UnknownConfigurationKey {
-                store: "HTTP",
+                store: "http-no-crypto",
```
Tpt: find/replace error?
```suggestion
                store: "HTTP",
```

src/gcp/credential.rs
```diff
@@ -123,44 +104,64 @@ pub struct GcpSigningCredential {
...
-        use rustls_pki_types::PrivateKeyDer;
-        use rustls_pki_types::pem::PemObject;
-
-        match PrivateKeyDer::from_pem_slice(encoded) {
-            Ok(PrivateKeyDer::Pkcs8(key)) => Self::from_pkcs8(key.secret_pkcs8_der()),
-            Ok(PrivateKeyDer::Pkcs1(key)) => Self::from_der(key.secret_pkcs1_der()),
-            Ok(_) => Err(Error::MissingKey),
-            Err(source) => Err(Error::ReadPem { source }),
-        }
+    #[cfg(feature = "aws-lc-rs")]
```
Tpt, line 122: hyper-nit (please ignore): moving the `cfg` switch inside of the function might help ensuring the signature does not diverge between aws-lc and rustls (same for the other functions)

Tpt (commented):
Not a maintainer but I would love to see this MR get merged.

I allowed myself 3 suggestions.

tustvold:
Apologies this is still on my list, things have just been extremely busy at work and maintaining this and arrow-rs is no longer my day job :sweat_smile:. I'll try to take a look this evening.

Cargo.toml
```diff
@@ -55,8 +55,9 @@ hyper = { version = "1.2", default-features = false, optional = true }
 md-5 = { version = "0.11.0", default-features = false, optional = true }
 quick-xml = { version = "0.39.0", features = ["serialize", "overlapped-lists"], optional = true }
 rand = { version = "0.10", default-features = false, features = ["std", "std_rng", "thread_rng"], optional = true }
-reqwest = { version = "0.12", default-features = false, features = ["rustls-tls-native-roots", "http2"], optional = true }
+reqwest = { version = "0.13", default-features = false, features = ["http2"], optional = true }
```
alamb, line 58: Can I request we make a separate PR to upgrade the version of `reqwest` to `0.13`

That way we won't tie the fate of that upgrade to the pluggable crypto stuff (and it will maek the PR easier / faster to review)

alamb:
Given it looks like we will make a new major version anyways I think it would be good to try and update reqwest as well

Cargo.toml
tustvold: Unfortunately the reqwest upgrade ends up tied to this, because there were changes to the way reqwest handles encryption

kevinjqliu (commented):
I think `*-custom` would be clearer than `*-no-crypto`.

From a user’s perspective, `aws-no-crypto`, `gcp-no-crypto`, and `azure-no-crypto` sound like the stores can operate without crypto. But these stores may still need crypto for request signing, signed URLs, service-account JWTs, SAS signing, etc.

The feature really means: enable the store without selecting object_store’s default crypto provider.

`*-custom` better matches that expectation:

- `aws`, `gcp`, `azure`, `http`: normal/default path
- `aws-custom`, `gcp-custom`, `azure-custom`, `http-custom`: advanced path for users who want to choose their own crypto/TLS setup
- `crypto-aws-lc`, `crypto-ring`: explicit provider choices

Example:

```toml
# default path
object_store = { features = ["aws"] }

# custom provider setup
object_store = { features = ["aws-custom", "crypto-ring", "tls-rustls"] }
```

This keeps the existing feature names for normal users, while making the advanced path match the actual use case: customization, not disabling crypto.

kevinjqliu:
Create an issue to facilitate with discussion on feature naming
https://github.com/apache/arrow-rs-object-store/issues/744

Cargo.toml
```diff
@@ -82,13 +83,26 @@ futures-channel = {version = "0.3", features = ["sink"]}
 
 [features]
 default = ["fs"]
-cloud = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand", "ring", "http-body-util", "form_urlencoded", "serde_urlencoded", "tokio"]
-azure = ["cloud", "httparse"]
+cloud-no-crypto = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand","http-body-util", "form_urlencoded", "serde_urlencoded", "tokio"]
```
kevinjqliu: I think we should align the feature naming here with the optional `reqwest` work in #724.
Feel free to chime in on the discussion here: https://github.com/apache/arrow-rs-object-store/issues/744

Instead of adding a separate `*-no-crypto` naming pattern, I think `*-base` can be the common escape hatch for both optional transport and optional crypto:

```text
cloud-base = shared cloud implementation without default reqwest or bundled crypto

aws-base   = cloud-base + AWS/S3-specific implementation deps
azure-base = cloud-base + Azure-specific implementation deps
gcp-base   = cloud-base + GCS-specific implementation deps
http-base  = cloud-base + HTTP-specific implementation deps

reqwest    = built-in reqwest HTTP transport

ring       = bundled ring crypto provider
aws-lc-rs  = bundled aws-lc-rs crypto provider

cloud      = cloud-base + reqwest + default crypto provider
             compatibility alias for the old cloud behavior

aws        = aws-base + reqwest + default crypto provider
azure      = azure-base + reqwest + default crypto provider
gcp        = gcp-base + reqwest + default crypto provider
http       = http-base + reqwest + default crypto provider
```

That gives users explicit composition:

```toml
# S3 implementation only; user provides HTTP connector and crypto provider
object_store = { default-features = false, features = ["aws-base"] }

# S3 implementation + built-in reqwest, but user controls crypto provider
object_store = { default-features = false, features = ["aws-base", "reqwest"] }

# S3 implementation + custom HTTP connector, but object_store provides bundled ring crypto
object_store = { default-features = false, features = ["aws-base", "ring"] }

# S3 implementation + built-in reqwest + explicit aws-lc-rs crypto provider
object_store = { default-features = false, features = ["aws-base", "reqwest", "aws-lc-rs"] }
```

This keeps the default user path simple:

```toml
object_store = { features = ["aws"] }
```

while making the advanced cases composable. 

The remaining compatibility/default decision is whether the batteries-included features should keep `ring` as the default crypto provider for compatibility, or intentionally switch to `aws-lc-rs`.

Tpt: > The remaining compatibility/default decision is whether the batteries-included features should keep ring as the default crypto provider for compatibility, or intentionally switch to aws-lc-rs.

My suggestion would be to follow `reqwest` and use `aws-lc-rs`. This way users of both `object_store` and `reqwest` do not end up with two crypto stacks.

alamb:
As others have noted, this is going to conflict / need feature flag alignment with 
- https://github.com/apache/arrow-rs-object-store/pull/724

I recommend we try and merge https://github.com/apache/arrow-rs-object-store/pull/724 in first, and then merge this one up and clean up the flags / dependencies

kylebarron:
#724 was just merged. This branch can now be updated for the resulting merge conflicts

alamb:
I am going to take a crack at updating this PR with the new regime

alamb:
I merged up from main. I went through the rust code which looks good. 

However, I am not convinced that we can actually disable aws-lc-rs when using ring. Looking into adding a test to verify

alamb:
Ok, I merged up from main and did some polishing. I think this PR is now almost ready as it could be, though there are some non trivial gaps. I suggest we merge the PR as is, and then address the gaps as a follow on:

## fix testing gaps
- Custom CryptoProvider / `crypto_provider` is entirely untested. 
- NoSystemCertificates not in the config round-trip test. client_test_config_from_map (src/client/mod.rs:1107) enumerates the other keys but omits the new one, so its parse→serialize round-trip and with_no_system_certificates are untested.

## No documentation
Docs say we support a custom provider but show no example.  "You can also implement a custom client::CryptoProvider" with no example. Ideally we would document this via, a doctest.


However, I don't think either of those things would block the release

alamb (approved):
I think this PR is good to go, but it would be good if someone took a final look at it (specifically the few commits I pushed yesterday)

It is kind of weird for me to approve something where I also did a bunch of hte work

alamb:
I'll plan to merge this PR tomorrow unless anyone else would like some additional time to review

Cargo.toml
```diff
@@ -84,29 +85,33 @@ futures-channel = {version = "0.3", features = ["sink"]}
...
-http = ["http-base", "reqwest"]
+# "Batteries-included" implementation features.
+# Each enables the default `reqwest` transport (with rustls) and `aws-lc-rs`
+azure = ["azure-base", "reqwest", "reqwest/rustls", "aws-lc-rs"]
+gcp = ["gcp-base", "reqwest", "reqwest/rustls", "aws-lc-rs"]
+aws = ["aws-base", "reqwest", "reqwest/rustls", "aws-lc-rs"]
+http = ["http-base", "reqwest", "reqwest/rustls", "aws-lc-rs"]
 
 fs = ["walkdir", "tokio", "nix", "windows-sys"]
-tls-webpki-roots = ["reqwest?/rustls-tls-webpki-roots"]
```
kevinjqliu, line 109: 👍 looks like `rustls-tls-webpki-roots` was removed in `reqwest` 0.13

kevinjqliu (commented):
Generally LGTM, codex helped flagged 2 bugs, 1 with azure and another with gcp. And I had a few minor nits. 
All the changes are in this commit, with corresponding comments explaining the changes: https://github.com/apache/arrow-rs-object-store/commit/69a0825

Feel free to take a look and take any that you think is needed. 

The default bundled crypto provider changed from `ring` to `aws-lc-rs`, so maybe we should add "api-change" label

src/client/crypto.rs
```diff
@@ -0,0 +1,410 @@
...
+#[non_exhaustive]
+pub enum SigningAlgorithm {
+    /// RSASSA-PKCS1-v1_5 using SHA-256
+    RS256,
+}
+
+/// Provides cryptographic primitives
+pub trait CryptoProvider: std::fmt::Debug + Send + Sync {
+    /// Compute a digest
+    fn digest(&self, algorithm: DigestAlgorithm) -> Result<Box<dyn DigestContext>>;
```
Tpt, line 39: Seems like all the callers already have something that could be seen as a sequence of byte slices so maybe a signature like could work: 
```rust
    fn digest(&self, algorithm: DigestAlgorithm, content: impl IntoIterator<Item=impl AsRef<[u8]>>) -> Result<Vec<u8>>;
```

This would drop the need for a dyn-safe trait and hopefully might help slightly performances.

However, this looks like a follow-up and not something for this MR.

Tpt (commented):
Just read this MR again. Everything looks good to me.

alamb:
> Generally LGTM, codex helped flagged 2 bugs, 1 with azure and another with gcp. And I had a few minor nits. All the changes are in this commit, with corresponding comments explaining the changes: [69a0825](https://github.com/apache/arrow-rs-object-store/commit/69a0825)
> 
> Feel free to take a look and take any that you think is needed.
> 
> The default bundled crypto provider changed from `ring` to `aws-lc-rs`, so maybe we should add "api-change" label

Thank you @kevinjqliu 

I had one question about the no-fips feature flag: https://github.com/kevinjqliu/arrow-rs-object-store/commit/69a0825#r189078918

I also can't but help thinking we may be making things too complicated here by trying to help people use reqwest without one of its built in crypto providers -- I fear in so doing we are going to overfit and people won't actually be able to test with all the feature flags.

I have pulled  in 69a0825 and reverted the no-fips flag in https://github.com/apache/arrow-rs-object-store/pull/707/commits/dee6094ee99b9716d0d0f1e8d1790f5fbfe4884a -- thank you

kevinjqliu:
So I added `non-fips` so object_store’s `aws-lc-rs` feature could be used directly with the `*-base` features. Since `object_store` disables `aws-lc-rs` default features, one of `aws-lc-sys`, `non-fips`, or `fips` must be enabled; otherwise `aws-lc-rs` fails to compile.

This is whats [breaking CI currently](https://github.com/apache/arrow-rs-object-store/actions/runs/27684854585/job/81882682610?pr=707) when running `cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,aws-lc-rs -- -D warnings`:
```
Run cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,aws-lc-rs -- -D warnings
  cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,aws-lc-rs -- -D warnings
  shell: sh -e {0}
   Compiling aws-lc-rs v1.17.0
    Checking rustls-pki-types v1.14.1
error: failed to run custom build command for `aws-lc-rs v1.17.0`

Caused by:
  process didn't exit successfully: `/__w/arrow-rs-object-store/arrow-rs-object-store/target/debug/build/aws-lc-rs-a4641309a02b1d13/build-script-build` (exit status: 101)
  --- stdout
  cargo:rustc-check-cfg=cfg(aws_lc_rs_docsrs)
  cargo:rustc-check-cfg=cfg(disable_slow_tests)
  cargo:rustc-check-cfg=cfg(dev_tests_only)
  cargo:rerun-if-env-changed=AWS_LC_RS_DISABLE_SLOW_TESTS
  cargo:rerun-if-env-changed=AWS_LC_RS_DEV_TESTS_ONLY

  --- stderr

  thread 'main' (3575) panicked at /usr/local/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/aws-lc-rs-1.17.0/build.rs:59:9:
  one of the following features must be specified: `aws-lc-sys`, `non-fips`, or `fips`.
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
Error: Process completed with exit code 101.
```

One of the following aws-lc-rs features must be specified: `aws-lc-sys`, `non-fips`, or `fips`. Maybe we should use `aws-lc-rs/aws-lc-sys` instead. This was previously hidden because `reqwest/rustls` pulls in `aws-lc-sys`, but the new CI check exercises `*-base + aws-lc-rs` without `reqwest`.

Cargo.toml
```diff
@@ -84,29 +85,36 @@ futures-channel = {version = "0.3", features = ["sink"]}
...
 
 # Built-in reqwest-based HTTP transport.
 reqwest = ["dep:reqwest", "reqwest/stream"]
 
-# Provider base features.
-# These compile provider logic without forcing reqwest.
+# Bundled crypto providers.
+# Selecting one provides the default `CryptoProvider`; otherwise a custom
+# provider must be supplied at runtime.
+aws-lc-rs = ["dep:aws-lc-rs", "rustls-pki-types"]
```
kevinjqliu, line 97:
```suggestion
aws-lc-rs = ["dep:aws-lc-rs", "aws-lc-rs/aws-lc-sys", "rustls-pki-types"]
```

kevinjqliu:
> I also can't but help thinking we may be making things too complicated here by trying to help people use reqwest without one of its built in crypto providers -- I fear in so doing we are going to overfit and people won't actually be able to test with all the feature flags.

This is actually for users who wants to provide a custom HTTP client instead of `reqwest`. We want `*-base + aws-lc-rs` to be a standalone feature combination; the new CI check catches that case directly.

alamb:
> > I also can't but help thinking we may be making things too complicated here by trying to help people use reqwest without one of its built in crypto providers -- I fear in so doing we are going to overfit and people won't actually be able to test with all the feature flags.
> 
> This is actually for users who wants to provide a custom HTTP client instead of `reqwest`. We want `*-base + aws-lc-rs` to be a standalone feature combination; the new CI check catches that case directly.

I feel one difference is that actual customers will have their own Cargo.toml file where they can directly activate the features in the subcrates, so they won't be compiling **just** the object_store. So therefore I think the new CI check is invalid as it is verifying something that won't actually be done in practice. 

After some thought, I think a better way to test these features would be to create some targeted examples that are not part of the workspace and demonstrate (and verify they compile, etc) how to do the various usecases. I will file a follow on ticket to track this work

Cargo.toml
alamb: If we do this, doesn't that mean that downstream crates could not use `fips` or `non-fipts` 🤔  as the object store crate would activate them

kevinjqliu:
> After some thought, I think a better way to test these features would be to create some targeted examples that are not part of the workspace and demonstrate (and verify they compile, etc) how to do the various usecases. I will file a follow on ticket to track this work

Yea thats a great idea. 

I'm happy to move on with the PR as is. I can follow up with documentations on using `aws-lc-rs` without `reqwest`

alamb:
Thanks @kevinjqliu  -- I filed a ticket to track improved testing:
- https://github.com/apache/arrow-rs-object-store/issues/759

I will also fix the CI and add a comment link

Cargo.toml
alamb: I did this change slightly differently in https://github.com/apache/arrow-rs-object-store/pull/707/commits/53817ddfa8ed7037388c37968f0aad0ec612b800 

I activated it directly on the clippy command line, rather than as part of the `aws-lc-rs` object store feature (not the dependency)

alamb:
Okay -- hoping for a clean run now

Cargo.toml
kevinjqliu: >         run: cargo clippy --no-default-features --features aws-base,azure-base,gcp-base,http-base,aws-lc-rs,aws-lc-rs/aws-lc-sys -- -D warnings

yea that makes sense! that mirrors what the user of the crate would do 😄 thanks!

alamb:
merged up to resolve conflicts

alamb:
I am not entirely sure I got all the pluggable crypto hooked up correctly in the azure client, but assuming I can get CI to pass on this PR I am going to merge it in. We can work on improving / testing the pluggable crypto as a follow on PR

alamb:
I also filed a ticket to track adding more tests / docs for the pluggable crypto feature
- https://github.com/apache/arrow-rs-object-store/issues/760

alamb:
Let's get this in and iterate / test on main as this is the most potentially disruptive update. I will start preparing for a release with a version bump and release notes

kevinjqliu:
awesome thanks @alamb, i'll do some testing on main branch as well. 
And thank you @goffrie for the PR! 😄

alamb:
> awesome thanks @alamb, i'll do some testing on main branch as well. And thank you @goffrie for the PR! 😄

Yes, 100% -- thank you @kevinjqliu and @goffrie (and @djc and @tustvold and @Tpt ) -- quite a team effort

alamb:
I had a small follow on doc PR:
- https://github.com/apache/arrow-rs-object-store/pull/764

PR #713, opened 2026-05-07
GCP: fix copy (no) overwrite semantics

### LINKED ISSUE #712: GoogleCloudStorage::copy overwrite semantics are flipped.
**Describe the bug**
`GoogleCloudStorage::copy_opts` gets the CopyMode behavior backwards - it forbids overwrites on `CopyMode::Overwrite` and allows them on `CopyMode::Create`.

**To Reproduce**
Create two objects in a GCS bucket, create a google cloud storage client, and copy from one to the other with `CopyOptions { copy_mode: CopyMode::Overwrite, ..Default::default() }`. This should succeed, but it fails with `Error::AlreadyExists`.

**Expected behavior**
one object is copied over the other.

**Additional context**

james-rms:

Closes #712

# Rationale for this change
 
On `main`, `GoogleCloudStorage::copy(src, dst)` fails with `Error::AlreadyExists` if `dst` is the path to an existing object. This is the wrong default behavior - CopyMode defaults to Overwrite, not Create. This happens because of a bug in `GoogleCloudStorage::copy_opts`, where CopyMode::Overwrite gets mapped to `if_not_exists = true` instead of `false`. 

This PR fixes and moves the boolean mapping closest to where it gets used, to minimise confusion.

Yes. Anyone using GoogleCloudStorage::copy today who is relying on the incorrect behavior will see objects getting overwritten instead of an error.

### DIFF
--- a/src/gcp/client.rs
+++ b/src/gcp/client.rs
@@ -32,8 +32,8 @@ use crate::multipart::PartId;
 use crate::path::Path;
 use crate::util::hex_encode;
 use crate::{
-    Attribute, Attributes, ClientOptions, GetOptions, MultipartId, PutMode, PutMultipartOptions,
-    PutOptions, PutPayload, PutResult, Result, RetryConfig,
+    Attribute, Attributes, ClientOptions, CopyMode, GetOptions, MultipartId, PutMode,
+    PutMultipartOptions, PutOptions, PutPayload, PutResult, Result, RetryConfig,
 };
 use async_trait::async_trait;
 use base64::Engine;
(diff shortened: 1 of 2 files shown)


james-rms:
Note: I believe the GCS emulator used in tests does not support object versioning, so I didn't try to add a test here. Up to reviewers judgement though.

james-rms:
@tustvold @crepererum  sorry to bother - this feels reasonably high-impact to me as well as very simple to review. Are either of you available to take a look?

tustvold (approved):
Makes sense to me, thank you

PR #717, opened 2026-05-09
feat(gcp): support explicit bearer token config in parse flow

### LINKED ISSUE #716: GCS: support bearer-token auth in parse_url_opts without bypassing default option handling

We need to access `gs://` paths using an explicit OAuth bearer token.

Today, the normal GCS path goes through `object_store::parse_url_opts`. However, there does not appear to be a supported way to pass a bearer token through that standard option-parsing flow. Because of that, downstream users have to bypass the default path and manually construct a `GoogleCloudStorageBuilder` with a custom credential provider.

The problem with that workaround is that it no longer preserves full behavior parity with the default `gs` handling. The manual builder path can forward recognized `GoogleConfigKey` options, but anything else handled by the standard `parse_url_opts` path may be lost or silently ignored.

So the challenge is: we want bearer-token authentication for GCS without having to replace the default GCS option-handling path.

A supported way to provide a GCS bearer token while still using the standard `parse_url_opts` flow.

For example, one of these would solve the problem:

- support a bearer-token option in the GCS parsing path
- allow a credential override to be passed into the GCS builder/parser
- provide a standard API for overriding credentials without bypassing the normal option handling

The key requirement is that bearer-token auth should integrate with the existing `gs` option parsing instead of forcing downstream code into a custom builder path.

The alternative we considered was manually creating a `GoogleCloudStorageBuilder` and injecting a credential provider with the bearer token.

That works for authentication, but it has downsides:

- it bypasses `parse_url_opts`
- it requires downstream code to reimplement part of the standard GCS setup
- it may silently ignore extra options that would otherwise be handled by the default path

So it works as a workaround, but not as a clean or robust long-term solution.

**Additional context**

This came up in a downstream integration where we added custom handling for a `google_bearer_token` option only for `gs://` URLs.

That downstream implementation highlighted the API gap: once we switch to a custom builder to support bearer-token auth, we are no longer guaranteed to preserve all behavior from the default `parse_url_opts` path.

siddharthmittal13:

Closes #716.

# Rationale for this change

GCS already supported static bearer-token authentication through the builder, but there was no clean way to use it through the standard `parse_url_opts` configuration flow. That forced downstream users to bypass the normal GCS parsing path and manually construct a builder, which risks drifting from the default option-handling behavior.

This change adds explicit bearer-token support to the normal GCS config path so callers can keep using the standard parse/config flow.

This PR adds support for configuring GCS with an explicit OAuth bearer token through the standard GCS config parsing path.

It also documents that an explicitly provided bearer token is treated as a static credential and is not refreshed automatically.

In addition, it adds test coverage for bearer-token configuration aliases and the parse-layer GCS flow.

Yes.

Users can now provide an explicit bearer token for GCS using supported config keys such as `google_bearer_token` and `bearer_token` while continuing to use the normal `parse_url_opts` path.

This token is treated as a static credential override and will not refresh automatically.

There are no breaking API changes.

### DIFF
--- a/src/gcp/builder.rs
+++ b/src/gcp/builder.rs
@@ -670,7 +670,6 @@ impl GoogleCloudStorageBuilder {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures_executor::block_on;
     use std::collections::HashMap;
     use std::io::Write;
     use tempfile::NamedTempFile;
@@ -712,8 +711,8 @@ mod tests {
         assert_eq!(builder.bucket_name.unwrap(), google_bucket_name.as_str());
     }
 
-    #[test]
-    fn gcs_test_config_aliases() {
+    #[tokio::test]
+    async fn gcs_test_config_aliases() {
         // Service account path
         for alias in [
             "google_service_account",
@@ -752,7 +751,7 @@ mod tests {
                 .build()
                 .unwrap();
 
-            let credential = block_on(gcs.credentials().get_credential()).unwrap();
+            let credential = gcs.credentials().get_credential().await.unwrap();
             assert_eq!(credential.bearer, "test-token");
         }
     }
@@ -863,7 +862,9 @@ mod tests {
             google_bucket_name
         );
         assert_eq!(
-            builder.get_config_value(&GoogleConfigKey::BearerToken).unwrap(),
+            builder
+                .get_config_value(&GoogleConfigKey::BearerToken)
+                .unwrap(),
             google_bearer_token
         );
     }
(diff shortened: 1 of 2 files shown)


src/gcp/builder.rs
```diff
@@ -714,6 +744,17 @@ mod tests {
...
         }
+
+        for alias in ["google_bearer_token", "bearer_token"] {
+            let gcs = GoogleCloudStorageBuilder::new()
+                .with_config(alias.parse().unwrap(), "test-token")
+                .with_bucket_name("fake_bucket")
+                .build()
+                .unwrap();
+
+            let credential = block_on(gcs.credentials().get_credential()).unwrap();
```
tustvold: Could we use tokio::test here instead?

tustvold (approved):
Makes sense to me, just a minor nit

src/gcp/builder.rs
siddharthmittal13: Sure thanks addressed.

PR #718, opened 2026-05-11
refactor: unify internal naming of `allow_invalid_certificates`

CommanderStorm:

None, because should not be mentioned in a changelog.
Is internal-only

# Rationale for this change
 
Inconsistent between external api and internal api.

Rename internal variable

No user facing or breaking changes

### DIFF
--- a/src/client/mod.rs
+++ b/src/client/mod.rs
@@ -331,7 +331,7 @@ pub struct ClientOptions {
     proxy_ca_certificate: Option<String>,
     proxy_excludes: Option<String>,
     allow_http: ConfigValue<bool>,
-    allow_insecure: ConfigValue<bool>,
+    allow_invalid_certificates: ConfigValue<bool>,
     timeout: Option<ConfigValue<Duration>>,
     connect_timeout: Option<ConfigValue<Duration>>,
     read_timeout: Option<ConfigValue<Duration>>,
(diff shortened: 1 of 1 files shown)


kevinjqliu (approved):
LGTM

make sense to align with the builder method `with_allow_invalid_certificates`

btw ci is failing due to formatting issue

kylebarron (approved):

kevinjqliu (approved):
LGTM

CommanderStorm:
🎉

kylebarron:
This may have broken CI on main 🥲  https://github.com/apache/arrow-rs-object-store/actions/runs/27298445894

kylebarron:
I re-ran the CI and it passed the second time, so it must just be a flakey test

kevinjqliu:
its green! looks like it might be an intermittent issue

CommanderStorm:
The way in which that failed does not make sense to me as far as I undestand the rules of cargo test.

It is as if `cargo test` narrowed the feature set to what is actually being used in the testcase..
I checked release notes and commit logs and did not find any smoking guns of mayne a new version rolling out or something. strange.

kylebarron:
🤷‍♂️ we'll see if it happens again! Let's just say it's GitHub's fault haha

PR #721, opened 2026-05-11
feat(azure): expose `split_sas` for custom SAS credential providers

### LINKED ISSUE #720: Expose azure::split_sas for custom SAS credential providers

MicrosoftAzureBuilder already parses raw SAS strings into the Vec<(String, String)> accepted by AzureCredential::SASToken, but the parser (split_sas) is private. Users implementing their own CredentialProvider<Credential = AzureCredential> — e.g. fetching short-lived SAS tokens from a secret broker — have to reimplement it. Naive reimplementations using form_urlencoded mis-decode + as space, corrupting base64-encoded SAS signatures.

Make split_sas pub and re-export it from object_store::azure.

**Additional context**

dbatomic:
Closes #720.

# Rationale for this change
 Users implementing their own `CredentialProvider<Credential = AzureCredential>` that fetches SAS tokens at request time (e.g. tokens refreshed from a secret manager or coordination service) currently have to reimplement this parser to feed pairs into
`AzureCredential::SASToken`. Reimplementations are easy to get subtly wrong: in particular, using `form_urlencoded` decodes `+` as space, which corrupts base64-encoded SAS signatures.

Make `split_sas` public, change its return type to the public `crate::Result<Vec<(String, String)>>` (the inner `Error` variants already convert into `crate::Error` via the existing `From` impl, so the two existing internal call sites are unchanged), document it, and re-export it from `object_store::azure`. Add tests for the empty/`?` edge cases and the missing-`=` error path.

Plumbing to make `split_sas` public and additional test coverage. I tried to keep changes minimal - let me know if a broader API change is needed here, instead of a free function export.

No.

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -1074,7 +1074,8 @@ fn url_from_env(env_name: &str, default_url: &str) -> Result<Url> {
     Ok(url)
 }
 
-fn split_sas(sas: &str) -> Result<Vec<(String, String)>, Error> {
+/// Parse a SAS token string into the query pairs expected by [`AzureCredential::SASToken`].
+pub fn split_sas(sas: &str) -> Result<Vec<(String, String)>> {
     let sas = percent_decode_str(sas)
         .decode_utf8()
         .map_err(|source| Error::DecodeSasKey { source })?;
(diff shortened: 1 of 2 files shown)


kylebarron (approved):

kevinjqliu (approved):
LGTM

dbatomic:
@kylebarron - Is there anything else that needs to be done before merging this?

kylebarron:
I generally like to get a second opinion of someone who has merge access before merging myself (especially when it pertains to the public API)

tustvold (approved):
Seems reasonable to me - in hindsight there probably should be a SasToken type, and then this implements FromStr, but that ship has sailed

Issue #392, opened 2025-05-30
[EPIC] Release Tracker


alamb:
This is a list of planned releases
- [x] https://github.com/apache/arrow-rs-object-store/issues/306
- [ ] https://github.com/apache/arrow-rs-object-store/issues/307
- [ ] https://github.com/apache/arrow-rs-object-store/issues/367

PR #728, opened 2026-05-20
feat: Support custom DNS resolver

### LINKED ISSUE #726: Lightweight DNS cache with shuffle to prevent DNS flooding
Using object_store from polars with high-concurrency IO in distributed mode on EKS environment using S3 storage may lead to DNS flooding. Subsequently, this results in significant slowdown of the overall query.

We would like to see object_store to have a lightweight built-in DNS cache with shuffling, exposed through configuration option(s). Much smaller than a full-blown DNS resolver such as `hickory-resolver`.

Alternatively, expose a `with_dns_resolver` client option which accepts a custom resolver (`Resolve` trait). This would allow us to inject our own resolver. Or expose a public `options.client_builder()` so we can integrate into the `options.client()` logic.

(1) Out-of-the-box, object_store give shuffling, but no caching.
(2) One can add `reqwest` with the `hickory-resolver` feature, and turn off `RandomizeAddresses`, but this only gives caching and no shuffling. The cost (importing a full DNS resolver) is high, too high for a dataframe library. The configuration feels hacky.
(3) Once can add a custom client and connector via `with_http_connector` which takes on the extra DNS responsibilities, but then the logic to handle `ClientOptions` is technically not viable since the fields nor the `options.client()` method are not publicly accessible.

For now, option (3) with some extra steps to re-implement `ClientOptions` may be our best stopgap solution, but it would be cleaner and easier to maintain if all dependent functionality is contained within and configurable as part of the crate (or pass a custom DNS resolver).

We can submit a (draft) PR if that helps.

**Additional context**
The following setup can trigger DNS flooding:
- default object_store and reqwest as used by polars
- polars distributed with high-concurrency (large TCP pool to servce many concurrent `get_range` requests)
- polars distributed running on default EKS, e.g., on 32 instances (note, EKS shares DNS, unlike a typical standalone OS)
- (further aggravated by the default `ndots:5` and dual-stack IPv4+v6)
The above results in DNS flooding to the pod that serves CoreDNS.

kdn36:

Mitigates but does not close #726.

# Rationale for this change
 
This PR enables users of object_store to provide a custom DNS resolver. Use case: the custom DNS resolver may have different properties (e.g., caching), and a different footprint (e.g., substantially smaller) than the default options.

Enables override of the DNS resolver as used by the `reqwest::HttpClient`.

Yes. This PR adds `ClientOptions::with_dns_resolver(Arc<dyn reqwest::dns::Resolve>)`.

### DIFF
--- a/src/client/dns.rs
+++ b/src/client/dns.rs
@@ -17,15 +17,10 @@
 
 //! Customizable DNS resolution for remote object stores
 
-use std::net::ToSocketAddrs;
-
-use rand::prelude::SliceRandom;
 use std::fmt::Debug;
 use std::future::Future;
-use std::net::{IpAddr, SocketAddr};
+use std::net::IpAddr;
 use std::pin::Pin;
-use std::sync::Arc;
-use tokio::task::JoinSet;
 
 /// Error returned by a [`DnsResolver`]
 pub type DnsError = Box<dyn std::error::Error + Send + Sync>;
@@ -43,7 +38,7 @@ pub type DnsFuture = Pin<Box<dyn Future<Output = Result<Vec<IpAddr>, DnsError>> 
 /// HTTP transport.
 ///
 /// Configure via [`ClientOptions::with_dns_resolver`]. The built-in
-/// [`reqwest`]-based transport honors this automatically; custom
+/// reqwest-based transport honors this automatically; custom
 /// [`HttpConnector`] implementations should retrieve it via
 /// [`ClientOptions::dns_resolver`] and apply it themselves.
 ///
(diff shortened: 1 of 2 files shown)


tustvold (commented):
I think this is probably fine, but generally we have tried to keep reqwest types out of our public API, especially in light of things like #724, and the fact we now support pluggable HTTP clients. The fact reqwest now has a resolve2 method, also doesn't really bode well for this API's stability longer-term 😅 It appears to have just been the legacy hyper API repackaged.

What do you think about defining our own resolve trait?

alamb:
Can someone help me out -- is it possible to use a custom DNS resolver with a custom https://docs.rs/object_store/latest/object_store/client/trait.HttpConnector.html and wrapper around another client?

alamb:
Marking as draft as I think this PR is no longer waiting on feedback and I am trying to make it easier to find PRs in need of review. Please mark it as ready for review when it is ready for another look

kdn36:
Apologies for the delay.

The custom client `HttpConnector` trait has the `connect(&self, options: &ClientOptions)` method, yet the fields of `ClientOptions` are not public. So to implement the trait and honor the options (which object_store may set or modify), one would have to extract them one by one and reapply. Technically not impossible, but error-prone and difficult to maintain. Unless I am missing something.

Object_store having its own `Resolve` trait (details to be worked out) should be viable and would enable object_store not to leak implementation details. This is something I can look into. At first glance, this would be the cleanest way forward for our use case.

Fwiw, the code in the PR is in use already with a lightweight shuffling cache and works as expected.

kdn36:
The latest commit introduces a local `DnsResolver` trait, which gets mapped to the `reqwest::Resolve` trait for the built-in `reqwest`-based path.

The proposed implementation has been validated end-to-end from our client-side library. Kindly review and let me know any needed changes.

kdn36:
Hi, any chance someone can look into this again? We would like to converge back to latest main including this functionality. Thanks!

tustvold (approved):
Makes sense to me, sorry for the delay in review

kdn36:
Thanks :). I rebased and did some minor refactor to fixup the feature gate wasm32 and reqwest. Any chance you could trigger a new CI run? Or can I do so myself?

alamb:
I kciked off the CI runs

kdn36:
Thanks. It looks like CI is green now :). Any chance this can get merged? When can we expect the next release?

alamb:
> Thanks. It looks like CI is green now :). Any chance this can get merged? When can we expect the next release?

- The current release schedule: https://github.com/apache/arrow-rs-object-store/issues/392

It seems like maybe we should plan a new incremental release perhaps 🤔

alamb:
FYI this caused a slight potential API breakage (due to auto trait bounds)
- https://github.com/apache/arrow-rs-object-store/issues/835 (simple fix in https://github.com/apache/arrow-rs-object-store/pull/836)

alamb:
> > Thanks. It looks like CI is green now :). Any chance this can get merged? When can we expect the next release?
> 
> * The current release schedule: [[EPIC] Planned Release Schedule #392](https://github.com/apache/arrow-rs-object-store/issues/392)
> 
> It seems like maybe we should plan a new incremental release perhaps 🤔

- I am planning for one https://github.com/apache/arrow-rs-object-store/issues/832

PR #732, opened 2026-05-23
[capabilities] add capabilities method to ObjectStore trait

vitoordaz:
I'm starting with a single capability for ordered list results.

Memory, GCP, and Azure list objects always return ordered results. Local filesystem storage uses `WalkDir` which relies on OS `readdir` syscall that does not guarantee lexicographic order because it depends on file system.

For AWS, it depends on the bucket type; for directory buckets, the results are not ordered.

I'm thinking about adding a new config option for indicating whether the S3 bucket is a directory bucket or not. But for now, we can say that AWS results are not ordered.

- Closes https://github.com/apache/arrow-rs-object-store/issues/675.

# Rationale for this change
 
This will allow object store users to write more efficient code by leveraging underlying object store features.

A new `capabilities` method is added to the ObjectStore trait.

This change is backward compatible.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -1119,11 +1119,6 @@ impl AmazonS3Builder {
     }
 
     /// Override the [`Capabilities`] advertised by this store.
-    ///
-    /// By default the store reports `ordered_listing: true` because S3
-    /// `ListObjectsV2` returns results in lexicographic order. Use this
-    /// method if you are connecting to an S3-compatible endpoint whose
-    /// behaviour differs from the standard S3 API.
     pub fn with_capabilities(mut self, capabilities: Capabilities) -> Self {
         self.capabilities = Some(ConfigValue::Parsed(capabilities));
         self
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -196,7 +196,7 @@ pub struct AmazonS3Builder {
     /// The [`HttpConnector`] to use
     http_connector: Option<Arc<dyn HttpConnector>>,
     /// Capabilities to advertise for this store instance
-    capabilities: Option<Capabilities>,
+    capabilities: Option<ConfigValue<Capabilities>>,
 }
 
 /// Configuration keys for [`AmazonS3Builder`]
@@ -467,6 +467,9 @@ pub enum AmazonS3ConfigKey {
 
     /// Encryption options
     Encryption(S3EncryptionConfigKey),
+
+    /// Override the capabilities advertised by this store.
+    Capabilities,
 }
 
 impl AsRef<str> for AmazonS3ConfigKey {
@@ -501,6 +504,7 @@ impl AsRef<str> for AmazonS3ConfigKey {
             Self::RequestPayer => "aws_request_payer",
             Self::Client(opt) => opt.as_ref(),
             Self::Encryption(opt) => opt.as_ref(),
+            Self::Capabilities => "aws_capabilities",
         }
     }
 }
@@ -561,6 +565,7 @@ impl FromStr for AmazonS3ConfigKey {
             "aws_sse_customer_key_base64" | "sse_customer_key_base64" => Ok(Self::Encryption(
                 S3EncryptionConfigKey::CustomerEncryptionKey,
             )),
+            "aws_capabilities" => Ok(Self::Capabilities),
             _ => match s.strip_prefix("aws_").unwrap_or(s).parse() {
                 Ok(key) => Ok(Self::Client(key)),
                 Err(_) => Err(Error::UnknownConfigurationKey { key: s.into() }.into()),
@@ -713,6 +718,9 @@ impl AmazonS3Builder {
                     self.encryption_customer_key_base64 = Some(value.into())
                 }
             },
+            AmazonS3ConfigKey::Capabilities => {
+                self.capabilities = Some(ConfigValue::Deferred(value.into()))
+            }
         };
         self
     }
@@ -769,6 +777,7 @@ impl AmazonS3Builder {
             AmazonS3ConfigKey::DisableTagging => Some(self.disable_tagging.to_string()),
             AmazonS3ConfigKey::DisableBulkDelete => Some(self.disable_bulk_delete.to_string()),
             AmazonS3ConfigKey::RequestPayer => Some(self.request_payer.to_string()),
+            AmazonS3ConfigKey::Capabilities => self.capabilities.as_ref().map(ToString::to_string),
             AmazonS3ConfigKey::Encryption(key) => match key {
                 S3EncryptionConfigKey::ServerSideEncryption => {
                     self.encryption_type.as_ref().map(ToString::to_string)
@@ -1116,7 +1125,7 @@ impl AmazonS3Builder {
     /// method if you are connecting to an S3-compatible endpoint whose
     /// behaviour differs from the standard S3 API.
     pub fn with_capabilities(mut self, capabilities: Capabilities) -> Self {
-        self.capabilities = Some(capabilities);
+        self.capabilities = Some(ConfigValue::Parsed(capabilities));
         self
     }
 
@@ -1303,7 +1312,7 @@ impl AmazonS3Builder {
 
         Ok(AmazonS3 {
             client,
-            capabilities: self.capabilities,
+            capabilities: self.capabilities.map(|x| x.get()).transpose()?,
         })
     }
 }
@@ -1553,6 +1562,7 @@ impl From<S3EncryptionHeaders> for HeaderMap {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::Capability;
     use std::collections::HashMap;
 
     #[test]
@@ -1570,6 +1580,7 @@ mod tests {
             ("aws_session_token", aws_session_token.clone()),
             ("aws_unsigned_payload", "true".to_string()),
             ("aws_checksum_algorithm", "sha256".to_string()),
+            ("aws_capabilities", "ordered-listing".to_string()),
         ]);
 
         let builder = options
@@ -1589,6 +1600,14 @@ mod tests {
             Checksum::SHA256
         );
         assert!(builder.unsigned_payload.get().unwrap());
+        assert!(
+            builder
+                .capabilities
+                .unwrap()
+                .get()
+                .unwrap()
+                .has(Capability::OrderedListing)
+        );
     }
 
     #[test]
@@ -1643,7 +1662,8 @@ mod tests {
             .with_config(
                 "aws_sse_customer_key_base64".parse().unwrap(),
                 "some_customer_key",
-            );
+            )
+            .with_config(AmazonS3ConfigKey::Capabilities, "ordered-listing");
 
         assert_eq!(
             builder
@@ -1702,6 +1722,12 @@ mod tests {
                 .get_config_value(&"aws_sse_customer_key_base64".parse().unwrap())
                 .unwrap(),
             "some_customer_key"
+        );
+        assert_eq!(
+            builder
+                .get_config_value(&"aws_capabilities".parse().unwrap())
+                .unwrap(),
+            "ordered-listing"
         );
     }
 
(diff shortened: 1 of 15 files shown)


vitoordaz:
@alamb does this change seems reasonable to you?

vitoordaz:
cc: @tustvold

alamb:
Thank you for this -- I am not likely to have a chance to review this in the next week or so -- when I am trying to get the next release ready I hopefully can find time

Between now and then perhaps someone else will be able to help review this -- specifically what I think is needed is someone to think through the implications of adding this API

src/lib.rs
```diff
@@ -1131,6 +1133,14 @@ pub trait ObjectStore: std::fmt::Display + Send + Sync + Debug + 'static {
         self.delete(from).await?;
         Ok(())
     }
+
+    /// Return the [`Capabilities`] supported by this store.
+    ///
+    /// All capability fields default to `false`. Individual store
```
kevinjqliu: nit: empty instead of `false`?

src/capabilities.rs
```diff
@@ -0,0 +1,174 @@
...
+            Self::OrderedListing => write!(f, "{}", ORDERED_LISTING),
+        }
+    }
+}
+
+/// Optional features supported by an [`ObjectStore`] implementation.
+///
+/// Obtain the capabilities of a store by calling [`ObjectStore::capabilities`].
+/// All fields default to `false`; a store sets a field to `true` when it
+/// natively supports that feature.
```
kevinjqliu: nit: set false/true is confusing. maybe something like "has the capability"

src/aws/builder.rs
```diff
@@ -1071,6 +1084,17 @@ impl AmazonS3Builder {
         self
     }
 
+    /// Override the [`Capabilities`] advertised by this store.
+    ///
+    /// By default the store reports `ordered_listing: true` because S3
+    /// `ListObjectsV2` returns results in lexicographic order. Use this
+    /// method if you are connecting to an S3-compatible endpoint whose
+    /// behaviour differs from the standard S3 API.
```
kevinjqliu:
```suggestion
```
ordering_listing capability is not set for aws. 

also for this and other builders, would be good to move the documentation of capabilities away from `with_capabilities` and perhaps into `get_default_capabilities`

src/capabilities.rs
```diff
@@ -0,0 +1,174 @@
...
+///
+/// The struct is `#[non_exhaustive]` so that new capability flags can be added
```
kevinjqliu, line 70: The `Capabilities` docs say the struct is `#[non_exhaustive]`, but the attribute is not present. 

`Capability` is a public exhaustive enum. Since the motivation for this API includes future capabilities like negative ranges, adding a new variant later would be a semver concern. Could we either mark the enum `#[non_exhaustive]`, add a `Custom(...)` variant, or use a string-like capability representation?

src/integration.rs
```diff
@@ -398,6 +399,18 @@ pub async fn put_get_delete_list(storage: &DynObjectStore) {
...
+        let actual: Vec<_> = storage
+            .list(None)
+            .map_ok(|x| x.location)
+            .try_collect::<Vec<_>>()
+            .await
+            .unwrap();
+        let mut sorted_files = files.clone();
+        sorted_files.sort();
+        assert_eq!(actual, sorted_files);
+    }
```
kevinjqliu, line 421: could we test `list_with_offset` here as well? its explicitly called out in the OrderedListing docs

src/aws/builder.rs
```diff
@@ -1071,6 +1084,17 @@ impl AmazonS3Builder {
         self
     }
 
+    /// Override the [`Capabilities`] advertised by this store.
+    ///
+    /// By default the store reports `ordered_listing: true` because S3
```
kevinjqliu:
```suggestion
    /// By default the store reports `ordered-listing: true` because S3
```

kevinjqliu (commented):
Thanks for the PR! The general direction LGTM. A couple of comments

I also looked up the ordering documentation and implementation details for the three stores that advertise OrderedListing in this PR:

* GCS: The Cloud Storage docs explicitly say objects are ordered lexicographically by name:
  https://docs.cloud.google.com/storage/docs/listing-objects

* Azure: The List Blobs REST API docs say blobs are listed alphabetically in the response body, with uppercase letters listed first:
  https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs

  However, there is an important caveat for accounts with hierarchical namespace (HNS) enabled. Azure documents that `/` is treated as the lowest sort order, and that this difference only applies to recursive listing. Since `ObjectStore::list` is recursive, this may not be exactly equivalent to `object_store::Path::Ord` for all valid paths. For example, paths containing characters that sort before `/` in Rust string ordering, such as space, !, %, -, or ., may be ordered differently by Azure HNS recursive listing.

  So I think Azure should either only advertise OrderedListing when the account is not HNS-enabled, or the capability documentation should be weakened to mean “ordered according to the backend’s documented object-name ordering” rather than “ordered by Path::Ord.”

* Memory: the implementation uses BTreeMap<Path, Entry> and list() iterates the map with range((prefix)..). Since BTreeMap iterates keys in order, the memory store should return ordered results by implementation.

Based on this, advertising ordered listing for GCS and Memory seems straightforward. Azure also provides ordered results, but HNS-enabled accounts have different documented recursive listing semantics, so I would avoid advertising this capability for Azure HNS if the contract is strict Path::Ord ordering.

vitoordaz:
@kevinjqliu Here is how I think about this. If user is using vanilla object store implementation than default capabilities for each type of ObjectStore should be enough and as you pointed out these APIs rarely change.

If user is using ObjectStore that implements some of the public object store (for example ObjectStore that implements AWS S3 API) then they can override default values using config.

There are edge cases even in case of S3, object list order depends on bucket type (for directory buckets list is not ordered). For these cases we can use conservative default values.

kevinjqliu:
> There are edge cases even in case of S3, object list order depends on bucket type (for directory buckets list is not ordered). For these cases we can use conservative default values.

yea that makes sense. i think azure blob with HNS enabled is another edge case, maybe we can be conservative here and unset OrderedListing for azure. 
Alternatively, we can make capabilities take into account specific configurations like HNS for azure. For example, Azure would have the `OrderedListing` capability when HNS is disabled.

src/gcp/mod.rs
```diff
@@ -66,6 +66,11 @@ mod credential;
 
 const STORE: &str = "GCS";
 
+// OrderedListing is supported by all GCP bucket types.
```
kevinjqliu, line 69:
```suggestion
// OrderedListing is supported by all GCP bucket types.
// https://docs.cloud.google.com/storage/docs/listing-objects
```

src/azure/mod.rs
```diff
@@ -58,10 +58,16 @@ pub use credential::AzureCredential;
 
 const STORE: &str = "MicrosoftAzure";
 
+// OrderedListing capability is not supported by with Azure Storage Hierarchical Namespace is enabled.
```
kevinjqliu, line 61:
```suggestion
// OrderedListing capability is not supported by with Azure Storage Hierarchical Namespace is enabled.
// https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs
```

src/aws/mod.rs
```diff
@@ -79,10 +79,16 @@ use crate::client::parts::Parts;
 use crate::list::{PaginatedListOptions, PaginatedListResult, PaginatedListStore};
 pub use credential::{AwsAuthorizer, AwsCredential};
 
+/// OrderedListing capability depends on the bucket type, it's not enabled for directory bucket.
```
kevinjqliu:
```suggestion
/// OrderedListing capability depends on the bucket type, it's not enabled for directory bucket.
/// https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
```

src/aws/builder.rs
```diff
@@ -1500,6 +1522,7 @@ impl From<S3EncryptionHeaders> for HeaderMap {
 #[cfg(test)]
```
kevinjqliu, line 1557: nit: should we add the same tests for azure/gcp/memory?

kevinjqliu (approved):
LGTM! 

I found a few good documentation pages that i think we should link for the respective object stores. 
To summarize the PR, we're setting `OrderedListing` capability for `memory` and `gcs`, but not for `aws` and `azure` to be conservative. 
- `aws` directory bucket does not guarantee ordering. 
- `azure` HNS enabled account does not guarantee lexicographical ordering.

We can address those as follow ups. Would be great to set the capability once we determine its not one of the edge cases

src/capabilities.rs
```diff
@@ -0,0 +1,174 @@
...
+    ///
+    /// Returns `None` if `s` does not correspond to any known capability.
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            ORDERED_LISTING => Ok(Self::OrderedListing),
+            cap => Err(Error::Generic {
+                store: "Config",
+                source: format!("invalid capability: {cap}").into(),
+            }),
+        }
```
kevinjqliu: nit: align behavior with doc string or vice versa, one says returns None the other throws

src/capabilities.rs
```diff
@@ -0,0 +1,174 @@
...
+        let mut iter = self.supported.iter();
+        if let Some(cap) = iter.next() {
+            write!(f, "{}", cap)?;
+        }
+        for cap in iter {
+            write!(f, ", {}", cap)?;
+        }
+        Ok(())
+    }
+}
```
kevinjqliu, line 134: nit:

Capabilities stores capabilities in a HashSet, and Display iterates over self.supported.iter(). That is fine with one capability today, but once more capabilities are added, the string form could become non-deterministic. Since this is now exposed through config round-tripping like get_config_value, I’d make the output stable now.  
```
let mut caps: Vec<_> = self.supported.iter().collect();

caps.sort_by_key(|cap| cap.to_string());
```

src/capabilities.rs
```diff
@@ -0,0 +1,174 @@
...
+    ///
+    /// When this capability is present, callers may rely on the ordering and
+    /// avoid buffering all results solely for sorting purposes.
+    OrderedListing,
+}
+
+impl std::str::FromStr for Capability {
+    type Err = Error;
+
+    /// Parses a capability from its kebab-case string representation.
```
kevinjqliu: nit: 
thoughts on kebab-case vs snake_case? Most of the repo’s user-facing config strings appear to use `snake_case`, including `aws_capabilities`.

Since `Capabilities` is now parsed from config values as well, `aws_capabilities=ordered-listing` mixes a `snake_case` config key with a `kebab-case` capability value. I think `aws_capabilities=ordered_listing` would be more consistent with the rest of the config surface.

If we prefer `kebab-case`, maybe we should document that capability names intentionally use kebab-case and possibly accept both `ordered-listing` and `ordered_listing` to avoid bikeshedding / future compatibility.

kevinjqliu (commented):
A couple things flagged by AI

src/aws/mod.rs
```diff
@@ -79,10 +79,17 @@ use crate::client::parts::Parts;
...
+/// https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
```
kevinjqliu:
```suggestion
// OrderedListing capability depends on the bucket type, it's not enabled for directory bucket.
// https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
```

saw this in the commit diff, use `//` to align with other `get_default_capabilities` functions

src/capabilities.rs
```diff
@@ -0,0 +1,173 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let mut caps: Vec<_> = self.supported.iter().collect();
+        caps.sort_by_key(|cap| cap.to_string());
+        if let Some(cap) = caps.first() {
+            write!(f, "{}", cap)?;
+        }
+        for cap in caps[1..].iter() {
+            write!(f, ", {}", cap)?;
+        }
+        Ok(())
```
kevinjqliu:
```suggestion
        let mut caps: Vec<_> = self.supported.iter().map(ToString::to_string).collect();
        caps.sort();
        write!(f, "{}", caps.join(", "))
```

kevinjqliu (approved):
LGTM!

kevinjqliu:
i saw Andrew's comment on the original issue thread, https://github.com/apache/arrow-rs-object-store/issues/675#issuecomment-4674781494

lets discuss there before proceeding with this PR

vitoordaz:
@kevinjqliu do we have decision regarding this feature?

kevinjqliu:
i like the current state of the repo but i would defer to #675

in general i like the idea. a couple things i'd like to see:
* capabilities are conservative, instance-specific guarantees
* Absence of a capability means “unknown or unsupported,” not “definitely unsupported.”

alamb:
Thank you for the discussion and feedback @vitoordaz

PR #734, opened 2026-05-27
feat: add option to disable bulk delete for aws

### LINKED ISSUE #731: S3: delete() now requires DeleteObjects (bulk delete) support, breaks S3-compatible providers (Alibaba Cloud OSS, etc.)
### Describe the bug

[#549](https://github.com/apache/arrow-rs-object-store/pull/549) refactored `AmazonS3::delete()` so that it now delegates to `delete_stream()`, which issues a `POST /?delete` (S3 `DeleteObjects` / bulk delete) request instead of the previous single-object `DELETE /key`.

While the S3 *core* protocol is widely implemented by third-party providers, the `DeleteObjects` multi-object API is **not** universally supported. As a result, any user of `object_store` 0.x against a non-AWS S3-compatible endpoint that doesn't implement `POST /?delete` can no longer delete objects at all — even single-object deletes — because the only public API (`delete`) now routes through bulk delete.

Concretely, this breaks:

- **Alibaba Cloud OSS** — its S3-compatible endpoint does not implement `DeleteObjects`. Calls fail with an XML error (`NotImplemented` / `MethodNotAllowed`, depending on region).
- Likely affected: older MinIO versions, Cloudflare R2 in some configurations, Tencent COS legacy endpoints, and various on-prem S3 gateways. (Can collect more reports if helpful.)

Before #549, `delete()` issued a plain `DELETE /key`, which is part of the core S3 API and is supported by essentially every S3-compatible provider. After #549, there is no way for downstream users to opt out.

### To Reproduce

Using `object_store` configured with an Alibaba Cloud OSS S3-compatible endpoint:

```rust
let store = AmazonS3Builder::new()
    .with_endpoint("https://oss-<region>.aliyuncs.com")
    .with_bucket_name("my-bucket")
    .with_access_key_id(...)
    .with_secret_access_key(...)
    .build()?;

store.put(&path, bytes).await?;          // OK
store.delete(&path).await?;              // FAILS: NotImplemented / MethodNotAllowed
```

The same call succeeded on the previous release that used single-object `DELETE`.

### Expected behavior

`AmazonS3::delete(path)` should issue a single-object `DELETE /key` request, which is part of the core S3 API and works against every S3-compatible provider.

Bulk delete (`POST /?delete`, the `DeleteObjects` API) is an *optimization* and should only be used by `delete_stream` / batched APIs — not by the per-object `delete()`. This also mirrors the AWS SDKs, which expose `DeleteObject` and `DeleteObjects` as distinct operations precisely because the latter is optional / extended.

### Suggested fix

Two options, not mutually exclusive:

1. **Revert `delete()` to single-object DELETE** (preferred). `delete_stream` continues to use `DeleteObjects` for throughput; `delete()` stays compatible with every S3 provider. This is the lowest-risk change and matches the contract users had pre-#549.

2. **Add an opt-out config**, e.g. `S3ConfigKey::DisableBulkDelete` (or a positive `with_bulk_delete(false)` builder method), so users targeting incompatible providers can force single-object DELETEs throughout. Useful even with option 1, for `delete_stream` callers on providers that don't support `DeleteObjects`.

Happy to send a PR implementing either or both — would appreciate guidance on which direction maintainers prefer.

### Environment

- `object_store` version: 0.13.x
- Provider: Alibaba Cloud OSS (S3-compatible endpoint)

### Additional context

- Original PR: [#549](https://github.com/apache/arrow-rs-object-store/pull/549)
- AWS API reference: [[DeleteObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html)](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html) vs [[DeleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html)](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html) — distinct operations in the upstream API.

---

The error like this:

```
2026-05-22T02:00:03.274139257+00:00 ERROR infra::storage: Failed to delete object: Generic S3 error: Error performing POST http://openobserve.oss-cn-beijing-internal.aliyuncs.com/openobserve?delete in 9.033372ms - Server returned non-2xx status code: 405 Method Not Allowed: <?xml version="1.0" encoding="UTF-8"?>
<Error>
  <Code>MethodNotAllowed</Code>
  <Message>The specified method is not allowed against this resource.</Message>
  <RequestId>6A0FB8A330259633318A4958</RequestId>
  <HostId>openobserve.oss-cn-beijing-internal.aliyuncs.com</HostId>
  <Method>POST</Method>
  <ResourceType>Object</ResourceType>
  <EC>0017-00000001</EC>
  <RecommendDoc>https://api.aliyun.com/troubleshoot?q=0017-00000001</RecommendDoc>
</Error>
```

@crepererum @alamb

hengfeiyang:

- Closes #731

# Rationale for this change
 
We discussed in #731 and agree to add an option to disable `bulk_delete` for AWS.

Add a new option for AWS s3 client `with_disable_bulk_delete`.

User can use `client.with_disable_bulk_delete(true)` to disable bulk delete if his provider doesn't compatible for `s3 bulk delete API`.

### DIFF
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -205,6 +205,7 @@ pub(crate) struct S3Config {
     pub sign_payload: bool,
     pub skip_signature: bool,
     pub disable_tagging: bool,
+    pub disable_bulk_delete: bool,
     pub checksum: Option<Checksum>,
     pub copy_if_not_exists: Option<S3CopyIfNotExists>,
     pub conditional_put: S3ConditionalPut,
@@ -613,6 +614,16 @@ impl S3Client {
         Ok(results)
     }
 
+    /// Make a single-object S3 Delete request <https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObject.html>
+    ///
+    /// Unlike [`bulk_delete_request`](Self::bulk_delete_request), this issues a
+    /// plain `DELETE /key` request, which is part of the core S3 API and is
+    /// supported by every S3-compatible provider.
+    pub(crate) async fn delete_request(&self, path: &Path) -> Result<()> {
+        self.request(Method::DELETE, path).send().await?;
+        Ok(())
+    }
+
     /// Make an S3 Copy request <https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html>
     pub(crate) fn copy_request<'a>(&'a self, from: &Path, to: &'a Path) -> Request<'a> {
         let source = format!("{}/{}", self.config.bucket, encode_path(from));
@@ -1006,10 +1017,13 @@ fn encode_path(path: &Path) -> PercentEncode<'_> {
 mod tests {
     use super::*;
     use crate::GetOptions;
+    use crate::ObjectStore;
+    use crate::aws::AmazonS3;
     use crate::client::HttpClient;
     use crate::client::get::GetClient;
     use crate::client::mock_server::MockServer;
     use crate::client::retry::RetryContext;
+    use futures_util::{StreamExt, TryStreamExt};
     use http::Response;
     use http::header::{AUTHORIZATION, CONTENT_LENGTH};
     use hyper::Request;
@@ -1047,6 +1061,7 @@ mod tests {
             retry_config: Default::default(),
             sign_payload: false,
             disable_tagging: false,
+            disable_bulk_delete: false,
             checksum: None,
             copy_if_not_exists: None,
             conditional_put: Default::default(),
@@ -1102,6 +1117,7 @@ mod tests {
             retry_config: Default::default(),
             sign_payload: false,
             disable_tagging: false,
+            disable_bulk_delete: false,
             checksum: None,
             copy_if_not_exists: None,
             conditional_put: Default::default(),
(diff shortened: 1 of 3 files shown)


hengfeiyang:
@alamb @tustvold Can you help review?

alamb:
What do you think about @PsiACE suggestion here: https://github.com/apache/arrow-rs-object-store/issues/731#issuecomment-4550744611

> @Xuanwo 's suggestion is correct. Your request was sent to the wrong endpoint. Adding an option to disable bulk delete is not the right solution to this problem.

Adding an option to disable bulk delete does seem reasonable however

src/aws/mod.rs
```diff
@@ -264,6 +264,25 @@ impl ObjectStore for AmazonS3 {
...
+            return locations
+                .map(move |location| {
+                    let client = Arc::clone(&client);
+                    async move {
+                        let location = location?;
+                        client.delete_request(&location).await?;
+                        Ok(location)
+                    }
+                })
+                .buffered(20)
```
alamb, line 282: I wondered how you picked `20` as the concurrency but it seems to mirror the `buffered(20)` below

src/aws/client.rs
```diff
@@ -1153,6 +1169,55 @@ mod tests {
...
+        let config = default_headers_config(&mock);
+        let client = S3Client::new(config, HttpClient::new(reqwest::Client::new()));
+        let result = client.delete_request(&Path::from("test")).await;
+
+        assert!(result.is_ok());
+        mock.shutdown().await;
+    }
+
+    #[tokio::test]
+    async fn test_disable_bulk_delete_uses_single_object_delete() {
```
alamb: is there an existing test that shows that bulk_delete actually uses the POST /delete API? I didn't see one immediately

If not I think we should add one to document in test the different behaviors

alamb (commented):
Thanks @hengfeiyang  -- this code looks good to me -- it is clearly a compatible and opt-in option for the API

Two issues in my mind:
1. I had a question about one more test 
2. @PSIAce's question

src/aws/mod.rs
hengfeiyang: Yes. it is from bellow.

src/aws/client.rs
hengfeiyang: let me check this test case.

hengfeiyang:
@alamb about `@PsiACE's question`, I replied in the issue #731

hengfeiyang:
@alamb i added 4 new unit tests, the key point is different option will generate different result:
- with_virtual_hosted_style_request
- with_disable_bulk_delete

## Unit tests

  | # | Test name | virtual_hosted | disable_bulk_delete | Expected request |
  |---|-----------|----------------|---------------------|------------------|
  | 1 | `test_delete_default` | `false` | `false` | `POST /test-bucket?delete` |
  | 2 | `test_delete_default_with_disable_bulk` | `false` | `true` | `DELETE /test-bucket/foo`, `DELETE /test-bucket/bar` (no query) |
  | 3 | `test_delete_virtual_hosted` | `true` | `false` | `POST /?delete` |
  | 4 | `test_delete_virtual_hosted_with_disable_bulk` | `true` | `true` | `DELETE /foo`, `DELETE /bar` (no query) |

tustvold (approved):

alamb:
thanks @hengfeiyang and @tustvold

PR #736, opened 2026-06-01
Shaswot/cloudlfare native

shaswot16:

Closes #XXX

(Create an issue first if one does not already exist.)

# Rationale for this change

Cloudflare R2 is a widely used object storage service that exposes both an S3-compatible API and a native Cloudflare REST API. While R2 can already be accessed through the existing `AmazonS3` backend, a dedicated Cloudflare R2 implementation provides several advantages:

- Simplified authentication using Cloudflare API tokens instead of AWS-style HMAC credentials
- Native integration with Cloudflare's API ecosystem
- A foundation for Cloudflare Workers R2 bindings and `wasm32` support
- A more intuitive user experience for developers already using Cloudflare services

This follows the existing pattern of providing dedicated backends for major cloud providers such as AWS, Azure, and GCP.

### Core Implementation

- Added `CloudflareR2` object store implementation
  - Implements `ObjectStore`
  - Implements `MultipartStore`
  - Implements `PaginatedListStore`

### Builder and Configuration

- Added `CloudflareR2Builder`
- Added `CloudflareConfigKey`
- Added support for:
  - Environment-based configuration (`from_env`)
  - URL-based configuration
  - `r2://` URL scheme parsing

### Authentication

- Added `CloudflareCredential`
- Added static credential provider implementation based on Cloudflare API tokens

### HTTP Client

- Added Cloudflare REST API v4 client supporting:
  - Object upload
  - Object download
  - Object deletion
  - Object copy
  - Object listing
  - Multipart uploads

### Cloudflare Workers Support

- Added `CloudflareR2Workers`
- Added support for Cloudflare Workers runtimes (`wasm32`)
- Feature-gated behind `cloudflare-workers`

### Library Integration

- Added `ObjectStoreScheme::CloudflareR2`
- Added `r2://` scheme support in URL parsing
- Exported Cloudflare modules behind feature flags

### Cargo Features

Added two new optional features:

```toml
cloudflare = ["cloud"]
cloudflare-workers = ["worker"]

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,7 @@ nix = { version = "0.31.1", features = ["fs"] }
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 web-time = { version = "1.1.0" }
 wasm-bindgen-futures = "0.4.18"
+worker = { version = "0.8", optional = true }
 futures-channel = {version = "0.3", features = ["sink"]}
 
 [features]
@@ -85,6 +86,7 @@ default = ["fs"]
 cloud = ["serde", "serde_json", "quick-xml", "hyper", "reqwest", "reqwest/stream", "chrono/serde", "base64", "rand", "ring", "http-body-util", "form_urlencoded", "serde_urlencoded", "tokio"]
 azure = ["cloud", "httparse"]
 cloudflare = ["cloud"]
+cloudflare-workers = ["worker"]
 fs = ["walkdir", "tokio", "nix", "windows-sys"]
 gcp = ["cloud", "rustls-pki-types"]
 aws = ["cloud", "crc-fast", "md-5"]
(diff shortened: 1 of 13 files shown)


kevinjqliu:
Thanks for the PR. I think it makes sense to add cloudflare r2 to this repo since its an object storage implementation. What is the purpose of adding cloudflare workers? Looking at the docs, workers are compute nodes https://developers.cloudflare.com/r2/api/workers/workers-api-usage/

kylebarron:
I think we should have an issue discussion about this before making a PR. It's a significant new amount of code.

> * Simplified authentication using Cloudflare API tokens instead of AWS-style HMAC credentials
> * Native integration with Cloudflare's API ecosystem
> * A foundation for Cloudflare Workers R2 bindings and `wasm32` support
> * A more intuitive user experience for developers already using Cloudflare services

None of those really seem like big enough pros for me to have a whole new implementation, when the AWS implementation already works for Cloudflare. Is there a performance difference?

> What is the purpose of adding cloudflare workers?

Agreed I don't see the benefits of `worker` integration

alamb:
Thank you for this PR @shaswot16 . 

I am not sure this crate has the bandwidth to take on the maintenance of another object store implementation unless one of the other maintainers would like to volunteer (basically echoing what @kylebarron says)

Given that the implementations are easy to implement using the core traits, perhaps we can implement cloudflare support as another crate (e.g. `object-store-cloudflare` or something) -- if it gets enough users / attracts enough maintainer bandwidth we could potentially add it here

PR #742, opened 2026-06-09
Supporting CPK (Customer Provided Keys) in Azure Blob Storage requests

### LINKED ISSUE #741: Supporting CPKs for Azure Blob Requests

I want to be able to access paraquet / arrow files stored in blob storage that are encrypted with CPKs, this requires passing along some headers to azure.

Note: I've got a PR for this; https://github.com/apache/arrow-rs-object-store/pull/742

Adding a new configuration parameter that accepts an encryption key then it calls azure with that key i.e.

```
X-Ms-Encryption-Key: 
X-Ms-Encryption-Key-Sha256: 
X-Ms-Encryption-Algorithm: AES256
```
(note: the algorithm is always AES256)

A hacky custom one, but I feel up-sourcing it makes sense.

**Additional context**

Braedon-Wooding-Displayr:

- Closes #741

Add customer-provided key support to the Azure backend by wiring CPK headers through all blob reads, writes, heads, and deletes, with builder-level configuration and test coverage for ranged reads and real-account behavior.

# Rationale for this change
 

I think this is the cleanest way to solve this, matches similar style for other azure configuration.

New configuration key `encryption_key`.

New configuration key `encryption_key` but I've already updated the docs comment for it, so it should be okay.

### DIFF
--- a/src/azure/builder.rs
+++ b/src/azure/builder.rs
@@ -15,7 +15,7 @@
 // specific language governing permissions and limitations
 // under the License.
 
-use crate::azure::client::{AzureClient, AzureConfig};
+use crate::azure::client::{AzureClient, AzureConfig, AzureEncryptionHeaders};
 use crate::azure::credential::{
     AzureAccessKey, AzureCliCredential, ClientSecretOAuthProvider, FabricTokenOAuthProvider,
     ImdsManagedIdentityProvider, WorkloadIdentityOAuthProvider,
@@ -84,6 +84,11 @@ enum Error {
 
     #[error("Missing component in SAS query pair")]
     MissingSasComponent {},
+
+    #[error("Invalid encryption key: {source}")]
+    InvalidEncryptionKey {
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
+    },
 
     #[error("Configuration key: '{}' is not known.", key)]
     UnknownConfigurationKey { key: String },
@@ -178,6 +183,8 @@ pub struct MicrosoftAzureBuilder {
     fabric_session_token: Option<String>,
     /// Fabric cluster identifier
     fabric_cluster_identifier: Option<String>,
+    /// Base64-encoded 256-bit customer-provided encryption key
+    encryption_key: Option<String>,
     /// The [`HttpConnector`] to use
     http_connector: Option<Arc<dyn HttpConnector>>,
 }
@@ -380,6 +387,13 @@ pub enum AzureConfigKey {
     /// - `fabric_cluster_identifier`
     FabricClusterIdentifier,
 
+    /// Base64-encoded customer-provided encryption key
+    ///
+    /// Supported keys:
+    /// - `azure_storage_encryption_key`
+    /// - `encryption_key`
+    EncryptionKey,
+
     /// Client options
     Client(ClientConfigKey),
 }
@@ -410,6 +424,7 @@ impl AsRef<str> for AzureConfigKey {
             Self::FabricWorkloadHost => "azure_fabric_workload_host",
             Self::FabricSessionToken => "azure_fabric_session_token",
             Self::FabricClusterIdentifier => "azure_fabric_cluster_identifier",
+            Self::EncryptionKey => "azure_storage_encryption_key",
             Self::Client(key) => key.as_ref(),
         }
     }
@@ -466,6 +481,7 @@ impl FromStr for AzureConfigKey {
             "azure_fabric_cluster_identifier" | "fabric_cluster_identifier" => {
                 Ok(Self::FabricClusterIdentifier)
             }
+            "azure_storage_encryption_key" | "encryption_key" => Ok(Self::EncryptionKey),
             // Backwards compatibility
             "azure_allow_http" => Ok(Self::Client(ClientConfigKey::AllowHttp)),
             _ => match s.strip_prefix("azure_").unwrap_or(s).parse() {
@@ -594,6 +610,7 @@ impl MicrosoftAzureBuilder {
             AzureConfigKey::FabricClusterIdentifier => {
                 self.fabric_cluster_identifier = Some(value.into())
             }
+            AzureConfigKey::EncryptionKey => self.encryption_key = Some(value.into()),
         };
         self
     }
@@ -635,6 +652,7 @@ impl MicrosoftAzureBuilder {
             AzureConfigKey::FabricWorkloadHost => self.fabric_workload_host.clone(),
             AzureConfigKey::FabricSessionToken => self.fabric_session_token.clone(),
             AzureConfigKey::FabricClusterIdentifier => self.fabric_cluster_identifier.clone(),
+            AzureConfigKey::EncryptionKey => self.encryption_key.clone(),
         }
     }
 
@@ -938,6 +956,25 @@ impl MicrosoftAzureBuilder {
         self
     }
 
+    /// Set the customer-provided encryption key (CPK) used to encrypt blob content.
+    ///
+    /// `key` must be a base64-encoded 256-bit AES key (the decoded value must be
+    /// exactly 32 bytes). The same key must be supplied on every subsequent read,
+    /// write, or copy of any blob created with it; if the key is lost or omitted
+    /// the data is unrecoverable. CPK material is sent to Azure on every request,
+    /// so the configured endpoint must use HTTPS.
+    ///
+    /// Only a subset of Blob storage operations support CPK
+    /// (see the [Azure documentation][cpk-ops]). When CPK is enabled, `copy`
+    /// switches from the asynchronous `Copy Blob` API to `Put Blob From URL`,
+    /// which is synchronous and limits the source blob to 5,000 MiB.
+    ///
+    /// [cpk-ops]: https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#blob-storage-operations-supporting-customer-provided-keys
+    pub fn with_encryption_key(mut self, key: impl Into<String>) -> Self {
+        self.encryption_key = Some(key.into());
+        self
+    }
+
     /// The [`HttpConnector`] to use
     ///
     /// On non-WASM32 platforms uses [`reqwest`] by default, on WASM32 platforms must be provided
@@ -1079,6 +1116,16 @@ impl MicrosoftAzureBuilder {
             (false, url, credential, account_name)
         };
 
+        let encryption_headers =
+            AzureEncryptionHeaders::try_new(self.encryption_key).map_err(|source| {
+                Error::InvalidEncryptionKey {
+                    source: match source {
+                        crate::Error::Generic { source, .. } => source,
+                        other => Box::new(other),
+                    },
+                }
+            })?;
+
         let config = AzureConfig {
             account,
             is_emulator,
@@ -1089,6 +1136,7 @@ impl MicrosoftAzureBuilder {
             client_options: self.client_options,
             service: storage_url,
             credentials: auth,
+            encryption_headers,
         };
 
         let http_client = http.connect(&config.client_options)?;
@@ -1137,6 +1185,8 @@ pub fn split_sas(sas: &str) -> Result<Vec<(String, String)>> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use base64::Engine;
+    use base64::prelude::BASE64_STANDARD;
     use std::collections::HashMap;
 
     #[test]
@@ -1355,6 +1405,59 @@ mod tests {
                 "use_fabric_endpoint not set for URL: {url}"
             );
         }
+    }
+
+    #[test]
+    fn azure_encryption_key_roundtrip() {
+        let key = BASE64_STANDARD.encode([7_u8; 32]);
+        let builder = MicrosoftAzureBuilder::new().with_encryption_key(&key);
+
+        assert_eq!(
+            builder
+                .get_config_value(&AzureConfigKey::EncryptionKey)
+                .as_deref(),
+            Some(key.as_str())
+        );
+    }
+
+    #[test]
+    fn azure_encryption_key_rejects_malformed_base64() {
+        let err = MicrosoftAzureBuilder::new()
+            .with_account("account")
+            .with_container_name("container")
+            .with_access_key(EMULATOR_ACCOUNT_KEY)
+            .with_encryption_key("not-base64!!!")
+            .build()
+            .unwrap_err();
+
+        let msg = err.to_string();
+        assert!(msg.contains("Invalid encryption key") || msg.contains("Invalid byte"));
+    }
+
+    #[test]
+    fn azure_encryption_key_rejects_short_decoded_key() {
+        let err = MicrosoftAzureBuilder::new()
+            .with_account("account")
+            .with_container_name("container")
+            .with_access_key(EMULATOR_ACCOUNT_KEY)
+            .with_encryption_key(BASE64_STANDARD.encode([7_u8; 31]))
+            .build()
+            .unwrap_err();
+
+        assert!(err.to_string().contains("must decode to 32 bytes, got 31"));
+    }
+
+    #[test]
+    fn azure_encryption_key_rejects_long_decoded_key() {
+        let err = MicrosoftAzureBuilder::new()
+            .with_account("account")
+            .with_container_name("container")
+            .with_access_key(EMULATOR_ACCOUNT_KEY)
+            .with_encryption_key(BASE64_STANDARD.encode([7_u8; 33]))
+            .build()
+            .unwrap_err();
+
+        assert!(err.to_string().contains("must decode to 32 bytes, got 33"));
     }
 
     #[test]
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -1670,6 +1670,7 @@ Authorization: Bearer static-token\r
         assert!(debug.contains("key_configured: true"));
     }
 
+    #[cfg(feature = "reqwest")]
     #[test]
     fn test_azure_sensitive_headers_redact_client_request_debug() {
         let encryption_key = BASE64_STANDARD.encode([7_u8; 32]);
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -41,6 +41,7 @@ use http::{
     header::{CONTENT_LENGTH, CONTENT_TYPE, HeaderMap, HeaderValue, IF_MATCH, IF_NONE_MATCH},
 };
 use rand::RngExt;
+use ring::digest;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -58,6 +59,22 @@ static MS_CONTENT_ENCODING: HeaderName = HeaderName::from_static("x-ms-blob-cont
 static MS_CONTENT_LANGUAGE: HeaderName = HeaderName::from_static("x-ms-blob-content-language");
 
 static TAGS_HEADER: HeaderName = HeaderName::from_static("x-ms-tags");
+static ENCRYPTION_KEY_HEADER: HeaderName = HeaderName::from_static("x-ms-encryption-key");
+static ENCRYPTION_KEY_SHA256_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-encryption-key-sha256");
+static ENCRYPTION_ALGORITHM_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-encryption-algorithm");
+static SOURCE_ENCRYPTION_KEY_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-key");
+static SOURCE_ENCRYPTION_KEY_SHA256_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-key-sha256");
+static SOURCE_ENCRYPTION_ALGORITHM_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-algorithm");
+// Put Blob From URL added source CPK headers in 2026-02-06.
+// https://learn.microsoft.com/en-us/rest/api/storageservices/version-2026-02-06
+// before this version you could only specify CPK headers for the destination.
+// we only upgrade to this version if you are applying CPK headers to a copy request.
+const PUT_BLOB_FROM_URL_SOURCE_CPK_VERSION: &str = "2026-02-06";
 
 /// A specialized `Error` for object store-related errors
 #[derive(Debug, thiserror::Error)]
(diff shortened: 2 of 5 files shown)


src/azure/mod.rs
```diff
@@ -378,6 +381,44 @@ mod tests {
         }
     }
 
+    // Azurite doesn't support CPK (just ignores it)
```
Braedon-Wooding-Displayr, line 447: I've run some tests against some real storage accounts (using this + a manual cmd line app), looks good.

It really is just passing 3 headers.

src/azure/builder.rs
```diff
@@ -85,6 +85,11 @@ enum Error {
     #[error("Missing component in SAS query pair")]
     MissingSasComponent {},
 
+    #[error("Invalid encryption key: {source}")]
```
kylebarron, line 88: Just noting this is a breaking change

src/azure/client.rs
```diff
@@ -190,6 +197,57 @@ impl AzureConfig {
...
+        };
+
+        let decoded_key = BASE64_STANDARD
+            .decode(encryption_key.as_bytes())
+            .map_err(|source| crate::Error::Generic {
+                store: STORE,
+                source: Box::new(source),
+            })?;
+
+        if decoded_key.len() != 32 {
```
kevinjqliu, line 267: nit: why 32 here? 

is it because
> An AES-256 key is 256 bits = 32 bytes raw.

src/azure/client.rs
```diff
@@ -190,6 +197,57 @@ impl AzureConfig {
...
+        })
+    }
+
+    fn apply(&self, builder: HttpRequestBuilder) -> HttpRequestBuilder {
+        match (&self.encryption_key, &self.encryption_key_sha256) {
+            (Some(encryption_key), Some(encryption_key_sha256)) => builder
+                .header(&ENCRYPTION_KEY_HEADER, encryption_key)
+                .header(&ENCRYPTION_KEY_SHA256_HEADER, encryption_key_sha256)
+                // Azure only supports this algorithm header
+                .header(&ENCRYPTION_ALGORITHM_HEADER, "AES256"),
```
kevinjqliu: nit: move this string up to the class

src/azure/builder.rs
kevinjqliu: curious, why this is a breaking change?

src/azure/client.rs
```diff
@@ -264,8 +322,8 @@ impl PutRequest<'_> {
             .as_deref()
             .map(|c| c.sensitive_request())
             .unwrap_or_default();
-        let response = self
-            .builder
+        let builder = self.config.encryption_headers.apply(self.builder);
+        let response = builder
             .header(CONTENT_LENGTH, self.payload.content_length())
             .with_azure_authorization(&credential, &self.config.account)
```
kevinjqliu: nit: this call feels awkward, could we move this into the request-builder extension layer, similar to `with_azure_authorization`?

Right now `AzureEncryptionHeaders::apply(builder)` feels a little detached from the rest of the request construction, and it also makes it easy to apply the CPK headers without remembering that the request now contains secret material. A fluent helper like `with_azure_encryption_headers(&self.config.encryption_headers)` would match the existing style better.

Ideally that helper would also mark the request as sensitive when `x-ms-encryption-key` is present, so callers do not need to separately remember to combine credential sensitivity with CPK sensitivity.

src/azure/client.rs
```diff
@@ -190,6 +197,57 @@ impl AzureConfig {
     }
 }
 
+#[derive(Default, Clone, Debug)]
```
kevinjqliu: nit: Can we avoid deriving Debug here? This struct contains the raw customer-provided encryption key, so debug output could leak key material. I’d prefer a custom Debug impl that only reports whether a key is configured, and maybe the SHA-256 value if we consider that safe enough, but not the raw key.

src/azure/builder.rs
```diff
@@ -1235,6 +1270,17 @@ mod tests {
...
 
+    #[test]
+    fn azure_encryption_key_roundtrip() {
+        let key = BASE64_STANDARD.encode([7_u8; 32]);
+        let builder = MicrosoftAzureBuilder::new().with_encryption_key(&key);
+
+        assert_eq!(
+            builder.get_config_value(&AzureConfigKey::EncryptionKey).as_deref(),
+            Some(key.as_str())
+        );
```
kevinjqliu, line 1420: could we also add tests for invalid values
- malformed base64
- base64 that decodes to fewer than 32 bytes
- base64 that decodes to more than 32 bytes

The happy-path roundtrip test is useful, but the main correctness behavior here is that invalid keys fail at build time rather than producing bad Azure requests.

kevinjqliu (commented):
Generally LGTM.

One small suggestion: could we move the encryption header application into the request-builder extension layer, similar to `with_azure_authorization`? Something like `with_azure_encryption_headers(...)` would make the call sites a bit cleaner and keep Azure request-specific header handling in one place.

BTW, it looks like copy_request doesn’t utilize CPK. Is that intentional?

Here’s the Azure CPK documentation for reference:
https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#request-headers-for-specifying-customer-provided-keys

src/azure/builder.rs
kylebarron: Oh ignore me; I thought this error was exposed publicly, and I noticed it didn't have `#[non_exhaustive]`. But I see now it's an internal error enum.

src/azure/client.rs
Braedon-Wooding-Displayr: Yes it is!  I'll add a comment.

src/azure/client.rs
Braedon-Wooding-Displayr: Done.

src/azure/client.rs
Braedon-Wooding-Displayr: Yeh, this is fair.

I unified the 2 sensitive handling and moved the handling to a EncryptionHeadersExt to match CredentialExt.

src/azure/client.rs
Braedon-Wooding-Displayr: Done, I decided not to leak the SHA_256 hash since that is pretty secure still.

Braedon-Wooding-Displayr:
I've fixed all of above @kevinjqliu  / @kylebarron 

I've extend it to support copy blobs (requires a different source- header).  This means that both destination & source will share the same CPK.  This is something only supported in [2026-02-06](https://learn.microsoft.com/en-us/rest/api/storageservices/version-2026-02-06) so I've had to do a version upgrade when using this.  I don't want to upgrade our base version required because azurite often trails behind and this is very recent, so instead I've just applied the version upgrade when this scenario is hit.  Happy to simplify this down just upgrading our base version if this is too much.

Ready for re-review.

src/azure/builder.rs
Braedon-Wooding-Displayr: Added these as well.

src/azure/client.rs
```diff
@@ -58,6 +59,19 @@ static MS_CONTENT_ENCODING: HeaderName = HeaderName::from_static("x-ms-blob-cont
...
+static ENCRYPTION_ALGORITHM_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-encryption-algorithm");
+static SOURCE_ENCRYPTION_KEY_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-key");
+static SOURCE_ENCRYPTION_KEY_SHA256_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-key-sha256");
+static SOURCE_ENCRYPTION_ALGORITHM_HEADER: HeaderName =
+    HeaderName::from_static("x-ms-source-encryption-algorithm");
+// Put Blob From URL added source CPK headers in 2026-02-06.
+const PUT_BLOB_FROM_URL_SOURCE_CPK_VERSION: &str = "2026-02-06";
```
kevinjqliu, line 79: nit: took me a while to figure out the "source" headers and "2026-02-06"

could we add a comment on why this is necessary? Its used for copy_request and 2026-02-06 is the first supported version. Its newer and than the base version and thus the requires the override
 
https://learn.microsoft.com/en-us/rest/api/storageservices/version-2026-02-06

src/azure/client.rs
```diff
@@ -181,6 +196,21 @@ impl AzureConfig {
...
         url
     }
+
+    /// Whether a request built with this config must be treated as sensitive.
+    ///
+    /// This is true when the credential is carried in the URL (SAS tokens) or
+    /// when customer-provided encryption keys are configured, since CPK requests
+    /// carry secret key material. Combining both here means callers only need to
+    /// remember a single source of truth for request sensitivity.
+    fn is_sensitive(&self, credential: &Option<Arc<AzureCredential>>) -> bool {
```
kevinjqliu, line 219: nit: i noticed a few places still calculate `is_sensitive` using previous method. can we update all for uniformity? 
In
- get_user_delegation_key
- list_request

src/azure/client.rs
kevinjqliu: nit: i still see the string here :P

src/azure/client.rs
```diff
@@ -190,6 +220,97 @@ impl AzureConfig {
...
+    fn is_enabled(&self) -> bool {
+        self.encryption_key.is_some()
+    }
+}
+
+/// Request-builder extension for customer-provided encryption keys (CPK).
+///
+/// Mirrors [`CredentialExt`] so that CPK headers are applied as part of the
+/// fluent request-construction chain, keeping the secret key material close to
+/// the rest of the request building rather than as a detached side-call.
```
kevinjqliu:
```suggestion
```
nit: i feel like we dont need this explanation

src/azure/client.rs
```diff
@@ -190,6 +220,97 @@ impl AzureConfig {
...
+pub(crate) trait EncryptionHeadersExt {
+    /// Apply the customer-provided encryption headers for the request target.
```
kevinjqliu, line 311:
```suggestion
    /// Apply the customer-provided encryption headers for the request target.
    /// <https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys>
```

src/azure/client.rs
```diff
@@ -190,6 +220,97 @@ impl AzureConfig {
...
+    fn with_azure_encryption_headers(self, headers: &AzureEncryptionHeaders) -> Self;
+
+    /// Apply the customer-provided encryption headers for a copy *source*.
```
kevinjqliu, line 315:
```suggestion
    /// Apply the customer-provided encryption headers for a copy *source*.
    /// <https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-from-url?tabs=microsoft-entra-id#request-headers-source-customer-provided-encryption-keys>
```

kevinjqliu: Could we add a short comment explaining the difference between `with_azure_encryption_headers` and `with_azure_source_encryption_headers`?

For copy with CPK, the `x-ms-encryption-*` headers apply to the destination blob, while `x-ms-source-encryption-*` applies to decrypting the source blob. Since both use the same configured key here, it is easy to miss why both sets are required.

src/azure/client.rs
```diff
@@ -637,6 +756,7 @@ impl AzureClient {
                 .client
                 .delete(url.as_str())
                 .header(CONTENT_LENGTH, HeaderValue::from(0))
+                .with_azure_encryption_headers(&self.config.encryption_headers)
```
kevinjqliu:
```suggestion
```
remove this, delete operation does not require CPK, so we dont need to send headers

https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#blob-storage-operations-supporting-customer-provided-keys

src/azure/client.rs
```diff
@@ -829,13 +958,11 @@ impl AzureClient {
...
         let url = self.config.path_url(path);
-        let sensitive = credential
-            .as_deref()
-            .map(|c| c.sensitive_request())
-            .unwrap_or_default();
+        let sensitive = self.config.is_sensitive(&credential);
         let response = self
             .client
             .get(url.as_str())
+            .with_azure_encryption_headers(&self.config.encryption_headers)
```
kevinjqliu:
```suggestion
```

get blob tags also does not require CPK, so lets remove setting header

https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#blob-storage-operations-supporting-customer-provided-keys

src/azure/builder.rs
```diff
@@ -898,6 +916,12 @@ impl MicrosoftAzureBuilder {
         self
     }
 
+    /// Set the base64-encoded 256-bit customer-provided encryption key.
```
kevinjqliu:
```suggestion
    /// Set the customer-provided encryption key (CPK) used to encrypt blob content.
    ///
    /// `key` must be a base64-encoded 256-bit AES key (the decoded value must be
    /// exactly 32 bytes). The same key must be supplied on every subsequent read,
    /// write, or copy of any blob created with it; if the key is lost or omitted
    /// the data is unrecoverable. CPK material is sent to Azure on every request,
    /// so the configured endpoint must use HTTPS.
    ///
    /// Only a subset of Blob storage operations support CPK
    /// (see the [Azure documentation][cpk-ops]). When CPK is enabled, `copy`
    /// switches from the asynchronous `Copy Blob` API to `Put Blob From URL`,
    /// which is synchronous and limits the source blob to 5,000 MiB.
    ///
    /// [cpk-ops]: https://learn.microsoft.com/en-us/azure/storage/blobs/encryption-customer-provided-keys#blob-storage-operations-supporting-customer-provided-keys
```

src/azure/credential.rs
```diff
@@ -278,6 +282,12 @@ pub(crate) trait CredentialExt {
         credential: &Option<impl Deref<Target = AzureCredential>>,
         account: &str,
     ) -> Self;
+
+    /// Override the Azure Blob service version used for a single request.
+    ///
+    /// [`with_azure_authorization`](Self::with_azure_authorization) preserves an
+    /// explicit version header instead of replacing it with the backend default.
+    fn with_azure_version(self, version: &'static str) -> Self;
```
kevinjqliu: nit: agent suggested pulling this out into its own trait since its not tied to authorization

```
/// Override the Azure Blob service version used for a single request.
///
/// [`with_azure_authorization`](CredentialExt::with_azure_authorization) preserves
/// an explicit version header instead of replacing it with the backend default.
pub(crate) trait RequestVersionExt {
    fn with_azure_version(self, version: &'static str) -> Self;
}
```

```
impl RequestVersionExt for HttpRequestBuilder {
    fn with_azure_version(self, version: &'static str) -> Self {
        self.header(&VERSION, version)
    }
}
```

src/azure/client.rs
kevinjqliu, line 839:
```suggestion
        let sensitive = self.config.is_sensitive(&credential);
        let batch_response = self
```

kevinjqliu:
```suggestion
            .retryable(&self.config.retry_config)
            .sensitive(sensitive)
            .send()
```
propagate `sensitive` on retry

src/azure/client.rs
```diff
@@ -705,7 +825,11 @@ impl AzureClient {
         Ok(results)
     }
 
-    /// Make an Azure Copy request <https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob>
+    /// Make an Azure copy request <https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob>.
+    ///
+    /// When customer-provided keys are enabled, this request uses the Put Blob
+    /// From URL header shape and must opt into the service version that added
+    /// source CPK headers.
```
kevinjqliu:
```suggestion
    /// Make an Azure copy request <https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob>.
    ///
    /// The classic `Copy Blob` API does not accept CPK headers, so when
    /// customer-provided keys are enabled this falls back to
    /// [Put Blob From URL][put-blob-from-url] and opts into the service version
    /// that added source CPK headers. That changes the semantics: the operation
    /// is synchronous, the source must be a block blob no larger than 5,000 MiB,
    /// and uncommitted blocks / the source block list are not preserved.
    ///
    /// [put-blob-from-url]: https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob-from-url
```

src/azure/client.rs
```diff
@@ -181,6 +196,21 @@ impl AzureConfig {
         }
         url
     }
+
+    /// Whether a request built with this config must be treated as sensitive.
+    ///
+    /// This is true when the credential is carried in the URL (SAS tokens) or
+    /// when customer-provided encryption keys are configured, since CPK requests
+    /// carry secret key material. Combining both here means callers only need to
+    /// remember a single source of truth for request sensitivity.
```
kevinjqliu:
```suggestion
    /// The retry layer's `sensitive` flag suppresses the request URL from
    /// error messages (see [`RetryableRequestBuilder::sensitive`]). For SAS
    /// credentials this is load-bearing because the token is carried as URL
    /// query parameters.
    ///
    /// CPK material lives in request *headers* (`x-ms-encryption-key` etc.),
    /// not in the URL, so today's URL-only redaction does not actively hide
    /// it. The flag is still set for CPK requests so that any future
    /// expansion of the redaction surface (headers, response bodies) covers
    /// CPK without further changes here, and so that operators have a single
    /// "this request touches secret material" signal for both auth modes.
    ///
    /// [`RetryableRequestBuilder::sensitive`]: crate::client::retry::RetryableRequestBuilder
```

kevinjqliu (commented):
Thanks for working on this! Super excited for this feature for blob and [OneLake](https://community.fabric.microsoft.com/t5/Fabric-Updates-Blog/Customer-Managed-Keys-in-OneLake-Strengthening-Data-Protection/ba-p/5172698).

Another round of comments, most are just on more documentations.

One important thing to point out about `is_sensitive`. It currently only suppress URL. CPK leaves in the headers and is not redacted by setting the flag. I added a comment explain this behavior. We should redact the headers but this is getting into the scope creep territory

alamb (commented):
Thank you @Braedon-Wooding-Displayr 
In general, I am not likely to have time to help research and review all Azure specific features. 

 I will defer to @kevinjqliu on the azure specifics. 

I think we were trying to keep the object store specific feature configuration relatively small in this repo, and fallback to the cloud provider native SDKs for the more advanced (but less used) features. Is that relevant for this PR?

kevinjqliu:
CI issue is unrelated to this PR, can rebase https://github.com/apache/arrow-rs-object-store/pull/746 to resolve it

src/azure/client.rs
Braedon-Wooding-Displayr: Oops, yep, moved it up to a class constant. Happy to move it somewhere else too (since this class structure has changed).

src/azure/client.rs
Braedon-Wooding-Displayr: Done

src/azure/client.rs
Braedon-Wooding-Displayr: Ohhh that's interesting.  It does make sense why it wouldn't.  Removed.  I added a small comment stating that we don't have to pass these keys.

src/azure/client.rs
Braedon-Wooding-Displayr: Same as delete I've also added a small comment and removed this set.

src/azure/builder.rs
Braedon-Wooding-Displayr: Updated.

src/azure/credential.rs
Braedon-Wooding-Displayr: Yeh I was being lazy here keeping it in credentials, moved out.

Braedon-Wooding-Displayr:
@kevinjqliu I've applied all of your comments above.

Secondarily, I've had a deep scan and we don't seem to log headers anywhere but what I've done is added a new function to client/builder.rs called sensitive_header that does what you are saying and we call that when adding the encryption-key/SHA256.  I felt it's better to be safe now rather than expose accidentally later if someone adds header logging.

GCP had a special function to add bearer_auth which I simplified to use this sensitive header function too.

kevinjqliu (commented):
I tested this PR using my own azure account against all functions that supports CPK. Everything look good, found a few minor issues. 

1. PR’s CPK copy path authenticates the destination request, but not the private copy source URL.
2. redact copy source when used
3. `signed_url` and `signed_urls` should return `NotSupported` when CPK is configured, since this crate’s signer returns only a URL and Azure CPK requires request headers.
4. Fixed some unit test cases 

I was using Codex for review, and it generated this commit locally with the fix:
https://github.com/apache/arrow-rs-object-store/commit/817f106d87fd48c4ef6db20d1c256d64ba640353
Feel free to use any parts of it! Verified that this fixed the `copy_request` auth issue, I was using `az` cli for auth

src/azure/client.rs
```diff
@@ -181,6 +199,30 @@ impl AzureConfig {
...
+    /// error messages (see [`RetryableRequestBuilder::sensitive`]). For SAS
+    /// credentials this is load-bearing because the token is carried as URL
+    /// query parameters.
+    ///
+    /// CPK material lives in request *headers* (`x-ms-encryption-key` etc.),
+    /// not in the URL, so today's URL-only redaction does not actively hide
+    /// it. The flag is still set for CPK requests so that any future
+    /// expansion of the redaction surface (headers, response bodies) covers
+    /// CPK without further changes here, and so that operators have a single
+    /// "this request touches secret material" signal for both auth modes.
```
kevinjqliu:
```suggestion
    /// CPK material lives in request *headers* (`x-ms-encryption-key` etc.),
    /// not in the URL. Those header values are marked sensitive when added to
    /// the request, while this flag ensures retry/error formatting also treats
    /// the whole request as sensitive.
```
update comment based on new improvements using `sensitive_header`

Braedon-Wooding-Displayr:
@kevinjqliu I've included your commit with just 1 change, I've made it so that COPY_SOURCE is always counted as a sensitive header I felt doing the partial sensitive if it includes query params was not necessary and that it can always be sensitive is just simpler.

Included the other comment fix.

Looks like this is probably ready for merge?

kevinjqliu (approved):
LGTM!

one last thing, codex flagged a rebase conflict issue -- looks like we need to rebase and add `#[cfg(feature = "reqwest")]` to tests due to #750

Braedon-Wooding-Displayr:
@kevinjqliu done!  Rebased + added that feature gate for tests.

Feel free to merge!

kevinjqliu:
> ---- local::tests::test_close_file_detects_error_unix stdout ----

looks like CI failed on something unrelated, might need to rerun this one

kevinjqliu:
@alamb if you could take a look at this, esp the changes in `src/client/builder.rs`. 
I've validated the azure specific changes using my own azure account

alamb (approved):
looks good to me -- thank you @Braedon-Wooding-Displayr and @kevinjqliu  for the verification

PR #743, opened 2026-06-09
Add `Extensions` in `*Result` objects

### LINKED ISSUE #740: Support `Extensions` in `*Result` objects

We wrap object stores in SlateDB. We plan to use [Extensions](https://docs.rs/object_store/latest/object_store/struct.Extensions.html) to send information from the upper (wrapper) layers down into the lower stores. For example, we have a caching object store and we tell the object store if we're retrying because of a failed decode (i.e. the data is corrupt) so the caching object store can drop its copy and re-fetch.

It would be nice to have object stores be able to set extensions in their results as well. For example, the caching object store might signal whether the read was a cache hit or cache miss (or partial hit/partial miss). There doesn't seem to be a way to do this right now.

I'd like to be able to set `Extensions` in GetResult, PutResult, ListResult, and PaginatedListResult.

It would be nice to be able to do this for deletes and head as well, but I see a `DeleteResult`/`HeadResult` aren't returned right now.

We are currently considering providing a wrapper supplier API in https://github.com/slatedb/slatedb/pull/1761:

1. User provides raw object store
2. We wrap the object store in a retrying object store wrapper and instrumented object store wrapper
3. We call "wrap" on our wrapped object store to allow the user to wrap it in their cached implementation.

This is a pretty poor user experience.

criccomini:

- Closes #740

# Rationale for this change

We make extensive use of `ObjectStore` wrappers that do things like caching, metrics tracking, handle retries, and so on. We've found it would be helpful to use `Extensions` not just for requests (which are already supported today), but also responses.

One example: we have a `CachedObjectStore` in SlateDB that we are going to make pluggable. We have an internal metrics wrapper that (among other things) tracks cache hits/misses. Now that we're exposing the cache publicly, we need a way for our wrapper to track whether the reads were a hit or miss. An `Extensions` in the `Result` seems appropriate here. Then users that implement a cache can use the extension to signal such information.

This is just one example.

Adds a public extensions: http::Extensions field to the existing result types (GetResult, PutResult, ListResult, and PaginatedListResult) mirroring the extensions field already present on the options structs (GetOptions, PutOptions, etc.). This gives ObjectStore implementations (e.g. caching wrappers) a way to return context-specific information, such as cache hit/miss status, to callers.

I did not add `Extensions` support for Delete/Head as there aren't `Result` types for these right now. That seemed like a heavier change.

- GetResult, PutResult, ListResult, and PaginatedListResult gain a new public field extensions. Code constructing these types with struct literals must add the field (e.g. extensions: Default::default()).
- PutResult no longer derives PartialEq/Eq; the manual impls compare e_tag and version only, ignoring extensions.

_WARN: This was coded with Claude Fable with some steering from me. I (a human) wrote this PR summary._


src/aws/client.rs
```diff
@@ -1380,4 +1386,114 @@ mod tests {
...
+            req: crate::client::HttpRequest,
+        ) -> Result<HttpResponse, HttpError> {
+            let mut response = crate::client::HttpService::call(&self.0, req).await?;
+            response.extensions_mut().insert(Marker(42));
+            Ok(response)
+        }
+    }
+
+    #[tokio::test]
+    async fn test_response_extensions_propagated() {
```
criccomini: This is the only test. Not sure if you want to clutter up the other ones with tests as well. Can do so if desired.

src/client/header.rs
```diff
@@ -70,15 +70,22 @@ pub(crate) enum Error {
...
 }
 
-/// Extracts a PutResult from the provided [`HeaderMap`]
+/// Extracts a PutResult from the provided response
+///
+/// Propagates the extensions of the response into the [`PutResult`]
 #[cfg(any(feature = "aws", feature = "gcp", feature = "azure"))]
 pub(crate) fn get_put_result(
-    headers: &HeaderMap,
+    response: crate::client::HttpResponse,
```
criccomini, line 78: Different ways to tackle this. LMK if you prefer another approach. Fable initially did a `mut` here which was kinda gross.

src/http/client.rs
```diff
@@ -415,6 +419,9 @@ impl GetClient for Client {
 #[derive(Deserialize, Default)]
 pub(crate) struct MultiStatus {
     pub response: Vec<MultiStatusResponse>,
+    /// The extensions of the HTTP response this was parsed from
+    #[serde(skip)]
+    pub extensions: ::http::Extensions,
```
criccomini, line 424: I like this style vs. returning a `(MultiStatus, Extensions)` for `list`. LMK if you prefer the tuple.

src/lib.rs
```diff
@@ -1865,16 +1879,42 @@ impl From<Attributes> for PutMultipartOptions {
...
     /// A version indicator for the newly created object
     pub version: Option<String>,
+    /// Implementation-specific extensions. Intended for use by [`ObjectStore`] implementations
+    /// that need to return context-specific information (like cache status) from trait methods.
+    ///
+    /// HTTP-backed stores in this crate populate this with the extensions of the HTTP
+    /// response, allowing custom HTTP middleware to propagate information to callers.
+    ///
+    /// These extensions are excluded from [`PartialEq`] and [`Eq`].
+    pub extensions: Extensions,
```
alamb, line 1897: this is technically a breaking API change (but that is ok as we are about to release a breaking release)

src/aws/client.rs
alamb: I think we should add coverage of all new code, otherwise we coudl potentially break this feature in some future refactoring but not know

> Not sure if you want to clutter up the other ones 

I agree this test seems like a lot of boilerplate setup. It would be nice to avoid that

How about writing something similar to https://github.com/apache/arrow-rs-object-store/blob/e89f62b6508f6f65873ce6a5017cd0d5d7395184/src/integration.rs#L51-L53 

And then calling it with an existing fixture, like 

https://github.com/apache/arrow-rs-object-store/blob/a9a83f18071d9e4c5636d379aaa0513c1e06d0d4/src/aws/mod.rs#L688-L689

alamb (commented):
Thanks @criccomini  -- this looks good to me, other than I think we need test coverage for each object store. I left a suggestion, Let me know what you think

src/list.rs
```diff
@@ -61,6 +61,12 @@ pub struct PaginatedListResult {
     pub result: ListResult,
     /// If result set truncated, the pagination token to fetch next results
     pub page_token: Option<String>,
+    /// Implementation-specific extensions. Intended for use by implementations
+    /// that need to return context-specific information (like cache status) from trait methods.
+    ///
+    /// HTTP-backed stores in this crate populate this with the extensions of the HTTP
+    /// response, allowing custom HTTP middleware to propagate information to callers.
+    pub extensions: http::Extensions,
```
alamb: Given that `ListResult` already has `Extensions` -- does it add additional value to add extensions to the PaginatedListResult as well?

src/http/client.rs
alamb: I agree it looks good

criccomini:
Great! I'll fix things up. 😀

src/aws/client.rs
criccomini: [804ecf9](https://github.com/apache/arrow-rs-object-store/pull/743/commits/804ecf9acdef7d00a61d9dd575c67384b55ba4eb)

src/list.rs
criccomini: [4f5aa13](https://github.com/apache/arrow-rs-object-store/pull/743/commits/4f5aa13658156ef1313f73842902635e3689acc0)

criccomini:
Seems like the failed tests are flakes/unrelated to this PR?

alamb:
> Seems like the failed tests are flakes/unrelated to this PR?

Unrelated to the PR for sure. I'll check them out

src/aws/mod.rs
```diff
@@ -701,6 +704,7 @@ mod tests {
         s3_encryption(&integration).await;
         put_get_attributes(&integration).await;
         list_paginated(&integration, &integration).await;
+        response_extensions(&integration, true).await;
```
alamb, line 707: 👍

alamb (approved):
Looks great to me -- thank you @criccomini 

The MSRV check seems to be failing on installing cargo-msrv so that is unrelated to this PR

PR #752, opened 2026-06-15
Use path-style URL in `resolve_bucket_region` for dotted bucket names

### LINKED ISSUE #747: `resolve_bucket_region` fails TLS handshake for bucket names containing dots
**Describe the bug**

`object_store::aws::resolve_bucket_region` always uses a virtual-hosted-style endpoint ([src/aws/resolve.rs#L52](https://github.com/apache/arrow-rs-object-store/blob/main/src/aws/resolve.rs)):

```rust
let endpoint = format!("https://{bucket}.s3.amazonaws.com");
```

If the bucket name contains dots, the resulting hostname doesn't match S3's `*.s3.amazonaws.com` wildcard cert (a TLS wildcard only covers one DNS label), so the TLS handshake fails before any request is sent:

```
Generic S3 error: Error performing HEAD https://my.dotted.bucket.tbae8c3b8.s3.amazonaws.com in 30ms - HTTP error: error sending request
caused by: client error (Connect): invalid peer certificate: certificate not valid for name "my.dotted.bucket.tbae8c3b8.s3.amazonaws.com"
```

This is a [documented AWS limitation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html): "the SSL wildcard certificate matches only buckets that do not contain dots". The rest of the crate is fine - `AmazonS3Builder` defaults to path-style - `resolve_bucket_region` is the only place the bucket name unconditionally ends up in the TLS hostname.

**To Reproduce**

```rust
object_store::aws::resolve_bucket_region("my.dotted.bucket.tbae8c3b8", &ClientOptions::new()).await
// => invalid peer certificate
```

**Expected behavior**

The region resolves. Path-style works for the same lookup - S3 returns `x-amz-bucket-region` either way (on 200/301/403):

```
$ curl -skI "https://s3.amazonaws.com/my.dotted.bucket.tbae8c3b8"
HTTP/1.1 301 Moved Permanently
x-amz-bucket-region: us-west-2
...
Server: AmazonS3
```

Fix can be simple - fall back to path-style when the bucket name contains dots (same heuristic the official AWS SDKs use):

```rust
let endpoint = if bucket.contains('.') {
    format!("https://s3.amazonaws.com/{bucket}")
} else {
    format!("https://{bucket}.s3.amazonaws.com")
};
```

Path-style deprecation isn't a concern here: AWS [delayed it indefinitely](https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/), largely because dotted buckets have no virtual-hosted HTTPS alternative.

Happy to submit a PR.

brankogrb-db:
## Summary

`resolve_bucket_region` unconditionally uses virtual-hosted-style URLs (`{bucket}.s3.amazonaws.com`). When the bucket name contains dots, this produces a hostname that doesn't match S3's `*.s3.amazonaws.com` wildcard TLS certificate (wildcards only cover a single DNS label), causing the TLS handshake to fail.

This PR falls back to path-style (`s3.amazonaws.com/{bucket}`) when the bucket name contains a dot — the same heuristic the official AWS SDKs use.

- Closes #747

### DIFF
--- a/src/aws/resolve.rs
+++ b/src/aws/resolve.rs
@@ -49,7 +49,12 @@ impl From<Error> for crate::Error {
 pub async fn resolve_bucket_region(bucket: &str, client_options: &ClientOptions) -> Result<String> {
     use http::StatusCode;
 
-    let endpoint = format!("https://{bucket}.s3.amazonaws.com");
+    // S3's wildcard TLS cert (*.s3.amazonaws.com) only covers one DNS label. Use path-style for dotted names.
+    let endpoint = if bucket.contains('.') {
+        format!("https://s3.amazonaws.com/{bucket}")
+    } else {
+        format!("https://{bucket}.s3.amazonaws.com")
+    };
 
     let client = client_options.client()?;
 


kylebarron (approved):

alamb:
Thank you @brankogrb-db and @kylebarron

PR #754, opened 2026-06-16
Support PutMultipartOptions for MultipartStore::create_multipart for AWS/GCP.

### LINKED ISSUE #745: Support PutMultipartOptions for MultipartStore::create_multipart
We are trying to attach SHA-256 user metadata to multipart uploads to S3.

Our current upload flow uses `MultipartStore::create_multipart + put_part + complete_multipart`, because it provides the scheduling, retry, and completion control we need. However, `MultipartStore::create_multipart` does not expose `PutMultipartOptions`; it calls the internal S3 create-multipart path with `PutMultipartOptions::default()`. This means callers using the low-level multipart API cannot attach user metadata during multipart upload initiation.

The high-level `ObjectStore::put_multipart_opts` path does support this via `PutMultipartOptions`, and the AWS implementation already passes those options through to `S3Client::create_multipart`, so the backend implementation appears to have the necessary support.

Expose a way to pass `PutMultipartOptions` when initiating a multipart upload through `MultipartStore`.

One possible change would be adding a `create_multipart_opts` method:

```rust
async fn create_multipart_opts(
    &self,
    path: &Path,
    opts: PutMultipartOptions,
) -> Result<MultipartId>;
```

`create_multipart` could remain as the default-options convenience path.

* Use `ObjectStore::put_multipart_opts`: While this supports user metadata today, it returns a `MultipartUpload` writer rather than exposing the multipart upload ID. It also assigns part indexes internally using a monotonically increasing counter, so callers cannot explicitly choose which part index they are uploading. That makes it unsuitable for lower-level upload flows that need direct control over upload IDs, part indexes, part scheduling, completion inputs, and individual part retry behavior.
* Extend `MultipartUpload` with a method such as `put_part_with_idx(part_idx, payload)` or `put_part_at(part_idx, payload)`: This would allow callers to use `ObjectStore::put_multipart_opts` while still controlling part index assignment. It would solve the metadata plus custom part scheduling case, although it would still not expose the underlying multipart upload ID for callers that need it.
* Use `buffered::BufWriter::with_attributes`: This also supports attributes and eventually calls `put_multipart_opts`, but it is a writer-style API and does not provide direct control over multipart upload creation, part scheduling, completion inputs, or upload IDs.
* Use the backend-specific AWS/S3 SDK or fork/wrap `object_store`: This would work, but loses the backend-agnostic `object_store` abstraction and duplicates functionality that already exists internally in the AWS implementation.

**Additional context**
The gap seems to be that the library supports metadata for multipart uploads, and it supports low-level multipart upload control, but there is currently no published API that supports both at the same time.

Would be happy to submit a PR for this.

BearMinimum98:

Adds `PutMultipartOptions` support to the low-level `MultipartStore` create flow via a new `create_multipart_opts` method.

This is implemented for AWS, GCP, in-memory, and wrapper stores. Azure returns `NotSupported` for non-default tags/attributes because Azure applies these options when finalizing block uploads, not when creating them. Supporting this in the low-level API would require local upload state, which could leak if uploads are never completed or aborted.

Closes #745

# Rationale for this change
 

See issue.

* Adds `MultipartStore::create_multipart_opts`
* For AWS/GCP, passes options through to the backend client
* For `PrefixStore`/`ThrottledStore`, delegates options to the wrapped multipart store
* For in-memory, stores attributes in `PartStorage` until completion
* For Azure, returns `NotSupported` for non-default tags/attributes

Yes. `MultipartStore` has a new default method, `create_multipart_opts`, so existing implementations remain source-compatible.

There are no breaking changes to the public API; the new method has a default implementation.

### DIFF
--- a/src/gcp/mod.rs
+++ b/src/gcp/mod.rs
@@ -233,6 +233,14 @@ impl MultipartStore for GoogleCloudStorage {
             .await
     }
 
+    async fn create_multipart_opts(
+        &self,
+        path: &Path,
+        opts: PutMultipartOptions,
+    ) -> Result<MultipartId> {
+        self.client.multipart_initiate(path, opts).await
+    }
+
     async fn put_part(
         &self,
         path: &Path,
@@ -326,6 +334,7 @@ mod test {
             // https://github.com/fsouza/fake-gcs-server/issues/852
             stream_get(&integration).await;
             multipart(&integration, &integration).await;
+            multipart_with_opts(&integration, &integration).await;
             multipart_put_part_out_of_order(&integration, &integration).await;
             multipart_race_condition(&integration, true).await;
             multipart_out_of_order(&integration).await;
--- a/src/integration.rs
+++ b/src/integration.rs
@@ -1083,7 +1083,10 @@ pub async fn multipart_with_opts(storage: &dyn ObjectStore, multipart: &dyn Mult
 
     let result = storage.get(&path).await.unwrap();
     assert_eq!(result.meta.size, chunk_size as u64 * 2);
-    assert_eq!(result.attributes, attributes);
+    // Some providers add default attributes, e.g. S3 may report a default Content-Type.
+    for (attribute, value) in &attributes {
+        assert_eq!(result.attributes.get(attribute), Some(value));
+    }
 }
 
 /// Tests that [`MultipartStore::put_part`] may be invoked with non-sequential part indices.
--- a/src/integration.rs
+++ b/src/integration.rs
@@ -29,7 +29,8 @@ use crate::multipart::MultipartStore;
 use crate::path::Path;
 use crate::{
     Attribute, Attributes, DynObjectStore, Error, GetOptions, GetRange, MultipartUpload,
-    ObjectStore, ObjectStoreExt, PutMode, PutPayload, UpdateVersion, WriteMultipart,
+    ObjectStore, ObjectStoreExt, PutMode, PutMultipartOptions, PutPayload, UpdateVersion,
+    WriteMultipart,
 };
 use bytes::Bytes;
 use futures_util::stream::FuturesUnordered;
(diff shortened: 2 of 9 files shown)


BearMinimum98:
@alamb Would greatly appreciate a review today to be able to make it into the release tomorrow, if not possible, disregard. If it doesn't cut the release tomorrow, about when will the next minor version release be?

src/integration.rs
```diff
@@ -1053,6 +1054,41 @@ pub async fn multipart(storage: &dyn ObjectStore, multipart: &dyn MultipartStore
...
+        .unwrap();
+
+    multipart
+        .complete_multipart(&path, &id, parts)
+        .await
+        .unwrap();
+
+    let result = storage.get(&path).await.unwrap();
+    assert_eq!(result.meta.size, chunk_size as u64 * 2);
+    // Some providers add default attributes, e.g. S3 may report a default Content-Type.
```
alamb, line 1086: 👍

src/gcp/mod.rs
```diff
@@ -236,6 +236,14 @@ impl MultipartStore for GoogleCloudStorage {
             .await
     }
 
+    async fn create_multipart_opts(
+        &self,
+        path: &Path,
+        opts: PutMultipartOptions,
+    ) -> Result<MultipartId> {
+        self.client.multipart_initiate(path, opts).await
```
alamb, line 244: This makes sense to me as `create_multipart` is simply passing in `PutMultipartOptions::default()` -- adding the options explicitly makes sense

```rust
    async fn create_multipart(&self, path: &Path) -> Result<MultipartId> {
        self.client
            .multipart_initiate(path, PutMultipartOptions::default())
            .await
    }
```

alamb (approved):
Thank you @BearMinimum98  -- this change makes sense to me and I found the PR well documented and well tested

🙏

PR #710, opened 2026-05-02
feat: add AzureConfigKey::CredentialType to select Azure credential method

jackye1995:
## Summary

- Adds `AzureConfigKey::CredentialType` config key (`azure_credential_type` / `credential_type`) that lets users explicitly select which Azure credential method to use
- When multiple credential configurations are present in environment variables (e.g. both workload identity and client secret), the builder's fixed resolution order may pick an undesired credential type
- Supported values: `auto` (default, preserves existing behavior), `bearer_token`, `access_key`, `client_secret`, `workload_identity`, `sas_token`, `azure_cli`, `managed_identity`
- Refactors credential resolution into helper methods for each credential type, improving readability

PR #757, opened 2026-06-17
Fix flaky local close-file tests

kevinjqliu:

Closes #.

# Rationale for this change
 

The close-file error tests tried to force an error by closing a temp file, rebuilding a `File` from the now-invalid fd/handle, and closing it again. In parallel test runs, that fd/handle value can be reused by another test, causing the close test to unexpectedly succeed or close another test's resource.

This makes CI flaky and has affected two PRs recently:
- apache/arrow-rs-object-store#742: https://github.com/apache/arrow-rs-object-store/actions/runs/27587416868/job/81712651684?pr=742
- apache/arrow-rs-object-store#710: https://github.com/apache/arrow-rs-object-store/actions/runs/25238554718/job/74009722459?pr=710

This adds small Unix and Windows helpers for closing raw fd/handle values, and keeps `close_file(File)` delegating to those helpers.

The close-file error tests now assert the expected invalid fd/handle errors directly, instead of reconstructing a `File` from a closed resource.

No. This only changes internal test coverage for local filesystem close-error handling.

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -143,23 +143,34 @@ impl From<Error> for super::Error {
 fn close_file(file: File) -> std::result::Result<(), io::Error> {
     #[cfg(target_family = "unix")]
     {
-        nix::unistd::close(file).map_err(|e| e.into())
+        use std::os::fd::IntoRawFd;
+
+        close_fd(file.into_raw_fd())
     }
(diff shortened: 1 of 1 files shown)


alamb (approved):
looks good to me -- thank you @kevinjqliu

PR #758, opened 2026-06-17
fix(local): fsync create-mode rename source delete

kevinjqliu:
- Follow-up to #643.
- part of https://github.com/apache/arrow-rs-object-store/issues/661

`rename_if_not_exists` uses `RenameTargetMode::Create`, which is implemented as copy-then-delete. The copy side now goes through the fsync-aware path, but the source delete also needs to sync the source parent directory when `with_fsync(true)` is enabled so the source-name removal is durable before `rename_opts` returns.

This keeps normal `delete_stream` behavior unchanged and only passes `self.fsync` for the internal delete that is part of create-mode rename.

Validation:
- cargo test -p object_store fsync_rename_if_not_exists_propagates_source_delete_sync_error
- cargo test -p object_store local::tests::

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -513,9 +513,8 @@ impl ObjectStore for LocalFileSystem {
                 let config = Arc::clone(&config);
                 maybe_spawn_blocking(move || {
                     let location = location?;
-                    // Do not apply `with_fsync` to standalone deletes.
-                    // Use `fsync=true` only when removing the source after a synced copy in
-                    // `rename_if_not_exists`.
+                    // `with_fsync` does not apply to standalone deletes; only create-mode rename
+                    // fsyncs its internal source removal as part of the durable copy-and-delete.
                     Self::delete_location(config, automatic_cleanup, &location, false)?;
                     Ok(location)
                 })
--- a/src/local.rs
+++ b/src/local.rs
@@ -44,7 +44,7 @@ use crate::{
     path::{Path, absolute_path_to_url},
     util::InvalidGetRange,
 };
-use crate::{CopyMode, CopyOptions, ObjectStoreExt, RenameOptions, RenameTargetMode};
+use crate::{CopyMode, CopyOptions, RenameOptions, RenameTargetMode};
 
 /// A specialized `Error` for filesystem object store-related errors
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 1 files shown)


kevinjqliu:
cc @Barre this is a follow up to #643 do you mind taking a look and see if it makes sense directionally?

src/local.rs
```diff
@@ -513,7 +513,7 @@ impl ObjectStore for LocalFileSystem {
                 let config = Arc::clone(&config);
                 maybe_spawn_blocking(move || {
                     let location = location?;
-                    Self::delete_location(config, automatic_cleanup, &location)?;
+                    Self::delete_location(config, automatic_cleanup, &location, false)?;
```
alamb, line 518: why doesn't this need to `fsync`? Maybe we can add a comment explaining why

alamb (approved):
makes sense to me -- thank you @kevinjqliu

src/local.rs
```diff
@@ -524,7 +513,10 @@ impl ObjectStore for LocalFileSystem {
                 let config = Arc::clone(&config);
                 maybe_spawn_blocking(move || {
                     let location = location?;
-                    Self::delete_location(config, automatic_cleanup, &location)?;
+                    // Do not apply `with_fsync` to standalone deletes.
+                    // Use `fsync=true` only when removing the source after a synced copy in
+                    // `rename_if_not_exists`.
```
kevinjqliu:
```suggestion
                    // `with_fsync` does not apply to standalone deletes; only create-mode rename
                    // fsyncs its internal source removal as part of the durable copy-and-delete.
```

src/local.rs
alamb: WHy doesn't fsync apply to standalone deletes (if we don't fsync isn't there some chance that a deleted file will get "undeleted" on crash)? 

This code is no worse than main so we can merge it in, but I don't fully understand. This can be a follow on PR

alamb (approved):
Thanks @kevinjqliu  -- I think this is progress. Let's do it

src/local.rs
kevinjqliu: > WHy doesn't fsync apply to standalone deletes (if we don't fsync isn't there some chance that a deleted file will get "undeleted" on crash)?

it should! but it doesnt right now, the documented mentions only write-like operations and not deletes
https://github.com/apache/arrow-rs-object-store/blob/c9db3e22d2eed690b7a9c14a988ec85d6b9304ad/src/local.rs#L310-L316

Let's do a follow up to make standalone `delete` / `delete_stream` crash-durable

kevinjqliu: created tracking issue https://github.com/apache/arrow-rs-object-store/issues/767

PR #766, opened 2026-06-18
digest and hmac: drop intermediate builder and directly take slices for input

Tpt:
It removes the need for intermediate traits.

All callers can easily build a slice of slices but the S3 object content SHA 256 computation that is expensive anyway.

Follow up to #707


src/client/crypto.rs
```diff
@@ -150,26 +124,24 @@ pub(crate) mod ring {
...
+            }
+            Ok(ctx.finish().as_ref().to_vec())
         }
 
-        fn hmac(&self, algorithm: DigestAlgorithm, secret: &[u8]) -> Result<Box<dyn HmacContext>> {
+        fn hmac(&self, algorithm: DigestAlgorithm, secret: &[u8], data: &[u8]) -> Result<Vec<u8>> {
             let algorithm = match algorithm {
                 DigestAlgorithm::Sha256 => hmac::HMAC_SHA256,
             };
             let ctx = hmac::Context::with_key(&hmac::Key::new(algorithm, secret));
```
kevinjqliu:
```suggestion
            let mut ctx = hmac::Context::with_key(&hmac::Key::new(algorithm, secret));
```
add mut, align with the other change

src/client/crypto.rs
```diff
@@ -35,42 +35,16 @@ pub enum SigningAlgorithm {
...
-pub trait DigestContext: Send {
-    /// Updates the digest with all the data in data.
-    ///
-    /// It is implementation-defined behaviour to call this after calling [`Self::finish`]
-    fn update(&mut self, data: &[u8]);
-
-    /// Finalizes the digest calculation and returns the digest value.
-    ///
-    /// It is implementation-defined behaviour to call this after calling [`Self::finish`]
-    fn finish(&mut self) -> Result<&[u8]>;
```
alamb, line 58: We could possibly have a smaller API change and just allow the traits to return a Vec<u8> directly 🤔  that would also avoid the need for a copy in some cases and might be a smaller API change?

```rust
fn build(self) -> Result<Vec<u8>> { 
...
}
```

alamb (commented):
Thanks @Tpt and @kevinjqliu

kevinjqliu:
btw ci issue is from docker failing

> Unable to find image 'amazon/amazon-ec2-metadata-mock:v1.9.2' locally
docker: Error response from daemon: Head "https://registry-1.docker.io/v2/amazon/amazon-ec2-metadata-mock/manifests/v1.9.2": Get "https://auth.docker.io/token?account=githubactions&scope=repository%3Aamazon%2Famazon-ec2-metadata-mock%3Apull&service=registry.docker.io": context deadline exceeded (Client.Timeout exceeded while awaiting headers)

EDIT: #776 should fix this going forward

src/client/crypto.rs
Tpt: Thank you for the review!

All digests are currently stored in fixed length arrays so it would not save any allocation in our case, just a possible memcopy at the cost of maybe a useless allocations.

On the breaking changes, I hoped that this changed would be reviewed before the release, seems it's not the case sadly. I just found quite sad to have a complicated API with a lot of dynamic traits where a smaller API could work. Feel free to close this MR.

alamb:
Merged up from main to get CI fixes

src/client/crypto.rs
alamb: > . I just found quite sad to have a complicated API with a lot of dynamic traits where a smaller API could work. Feel free to close this MR.

Yeah, this is a new API that we added to support pluggable crypto providers.

tustvold (changes requested):
The API was explicitly designed to not require all data to be resident in memory, which AFAICT the proposed change reverts.

Whilst we don't currently exploit this, it is an easily foreseeable future requirement. Whilst I agree the API surface is larger, most people shouldn't be implementing it, and any performance impact of the additional dynamic dispatch should be marginal in comparison to the cryptography.

Tpt:
@tustvold Make sense! Closing

PR #770, opened 2026-06-19
fix: quote ETags per RFC 9110 in InMemory and LocalFileSystem

### LINKED ISSUE #769: InMemory and LocalFileSystem return unquoted ETags, violating RFC 9110
**Describe the bug**

`InMemory` and `LocalFileSystem` return ETags as bare tokens (e.g. `abc123`) instead of the quoted-string form required by [RFC 9110 §8.8.3](https://datatracker.ietf.org/doc/html/rfc9110#section-8.8.3) (e.g. `"abc123"`).

Cloud-backed stores (AWS, Azure, GCP) return ETags in the correct quoted form, since they reflect what the HTTP server returns. This makes `InMemory` and `LocalFileSystem` inconsistent with the cloud backends and non-compliant with the [preconditions spec](https://datatracker.ietf.org/doc/html/rfc9110#name-preconditions) that `GetOptions::if_match` and `GetOptions::if_none_match` are documented against in this crate's own [docs](https://docs.rs/object_store/latest/object_store/struct.GetOptions.html).

**To Reproduce**

```rust
use object_store::ObjectStoreExt;
use object_store::aws::AmazonS3Builder;
use object_store::gcp::GoogleCloudStorageBuilder;
use object_store::local::LocalFileSystem;
use object_store::memory::InMemory;
use object_store::path::Path;
use tempfile::tempdir;

#[tokio::test]
async fn etag_format() {
    let path = Path::from("etag_format_test");

    let store = InMemory::new();
    let result = store.put(&path, "hello".into()).await.unwrap();
    println!("InMemory e_tag: {:?}", result.e_tag);

    let tmp = tempdir().unwrap();
    let store = LocalFileSystem::new_with_prefix(tmp.path()).unwrap();
    let result = store.put(&path, "hello".into()).await.unwrap();
    println!("Local    e_tag: {:?}", result.e_tag);

    let store = AmazonS3Builder::from_env().build().unwrap();
    let result = store.put(&path, "hello".into()).await.unwrap();
    println!("AWS      e_tag: {:?}", result.e_tag);

    let store = MicrosoftAzureBuilder::from_env().build().unwrap();
    let result = store.put(&path, "hello".into()).await.unwrap();
    println!("Azure    e_tag: {:?}", result.e_tag);

    let store = GoogleCloudStorageBuilder::from_env().build().unwrap();
    let result = store.put(&path, "hello".into()).await.unwrap();
    println!("GCP      e_tag: {:?}", result.e_tag);
}
```
```
InMemory e_tag: Some("0")
Local    e_tag: Some("1cc9cd-6549c8a45a90c-5")
AWS      e_tag: Some("\"5d41402abc4b2a76b9719d911017c592\"")
Azure    e_tag: Some("\"0x1BEDBB9371F4B80\"")
GCP      e_tag: Some("\"XUFAKrxLKna5cZ2REBfFkg==\"")
```

**Expected behavior**

`InMemory` and `LocalFileSystem` should return ETags in the form `"<value>"`, consistent with cloud backends and RFC 9110.

**Additional context**

N/A

bidord:

Closes #769

# Rationale for this change
 
Cloud-backed stores (AWS, Azure, GCP) return ETags as quoted-strings (e.g. `"abc123"`) and expect `If-Match`/`If-None-Match` values in that same form, as required by [RFC 9110 §8.8.3](https://datatracker.ietf.org/doc/html/rfc9110#section-8.8.3) and [§13 (Preconditions)](https://datatracker.ietf.org/doc/html/rfc9110#name-preconditions).

`InMemory` and `LocalFileSystem`, by contrast, were producing bare tokens, making them inconsistent with the cloud backends and harder to use as drop-in replacements.

RFC 9110 is already the reference linked from `ObjectMeta::e_tag`, `GetOptions::if_match`, and `GetOptions::if_none_match` in this crate's [docs](https://docs.rs/object_store/latest/object_store/struct.GetOptions.html), where examples already use the quoted form.

- `src/local.rs`: wrap `get_etag` output in double quotes
- `src/memory.rs`: wrap all ETag values produced by `InMemory` / `InMemoryUpload` in double quotes
- `src/lib.rs`: update precondition tests to use properly quoted ETags

`InMemory` and `LocalFileSystem` now return ETags in the form `"<value>"` instead of `<value>`. Other backends are unaffected.

Code that treats ETags as opaque strings and passes them back via `if_match` or `if_none_match` as-is will not be impacted.

However, code that expected the implementation-specific unquoted ETag format from one of these two backends might be impacted.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2492,7 +2492,7 @@ mod tests {
             location: Path::from("test"),
             last_modified: Utc.timestamp_nanos(100),
             size: 100,
-            e_tag: Some("123".to_string()),
+            e_tag: Some("\"123\"".to_string()),
             version: None,
         };
 
@@ -2521,16 +2521,16 @@ mod tests {
 
         options = GetOptions::default();
 
-        options.if_match = Some("123".to_string());
+        options.if_match = Some("\"123\"".to_string());
         options.check_preconditions(&meta).unwrap();
 
-        options.if_match = Some("123,354".to_string());
+        options.if_match = Some("\"123\",\"354\"".to_string());
         options.check_preconditions(&meta).unwrap();
 
-        options.if_match = Some("354, 123,".to_string());
+        options.if_match = Some("\"354\", \"123\"".to_string());
         options.check_preconditions(&meta).unwrap();
 
-        options.if_match = Some("354".to_string());
+        options.if_match = Some("\"354\"".to_string());
         options.check_preconditions(&meta).unwrap_err();
 
         options.if_match = Some("*".to_string());
@@ -2542,16 +2542,16 @@ mod tests {
 
         options = GetOptions::default();
 
-        options.if_none_match = Some("123".to_string());
+        options.if_none_match = Some("\"123\"".to_string());
         options.check_preconditions(&meta).unwrap_err();
 
         options.if_none_match = Some("*".to_string());
         options.check_preconditions(&meta).unwrap_err();
 
-        options.if_none_match = Some("1232".to_string());
+        options.if_none_match = Some("\"1232\"".to_string());
         options.check_preconditions(&meta).unwrap();
 
-        options.if_none_match = Some("23, 123".to_string());
+        options.if_none_match = Some("\"23\", \"123\"".to_string());
         options.check_preconditions(&meta).unwrap_err();
 
         // If-None-Match takes precedence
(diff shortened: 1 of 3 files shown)


kylebarron (approved):

alamb:
Thank you @bidord and @kylebarron

PR #777, opened 2026-06-22
Coalesce contiguous ranges in `LocalFileSystem::get_ranges`

### LINKED ISSUE #778: Coalesce contiguous ranges in LocalFileSystem::get_ranges
`LocalFileSystem` doesn't coalesce reads. It is faster to do a single read than do n calls when the reads are adjacent.

Like objectstore, coalesce ranges in `get_ranges` - but only do it when the gap is 0. 

Users could merge adjacent reach instead.

**Additional context**

Dandandan:
Closes: #778

`LocalFileSystem::get_ranges` issued one read per range. This coalesces contiguous/overlapping ranges into a single read, collapsing many adjacent ranges into far fewer syscalls.

- Uses a coalesce gap of **0** (vs the 1 MiB network default), so we never read bytes that weren't requested.

On a warm-cache benchmark, many adjacent small ranges drop ~2.6× (e.g. 1024 × 4 KiB contiguous: ~700 µs → ~260 µs); disjoint ranges are unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -42,7 +42,7 @@ use crate::{
     ObjectStore, PutMode, PutMultipartOptions, PutOptions, PutPayload, PutResult, Result,
     UploadPart, maybe_spawn_blocking,
     path::{Path, absolute_path_to_url},
-    util::InvalidGetRange,
+    util::{InvalidGetRange, merge_ranges},
 };
 use crate::{CopyMode, CopyOptions, RenameOptions, RenameTargetMode};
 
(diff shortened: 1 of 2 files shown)


src/local.rs
```diff
@@ -1739,6 +1765,42 @@ mod tests {
...
+
+        let cases: Vec<Vec<Range<u64>>> = vec![
+            vec![0..16, 16..32, 32..48],                    // adjacent -> coalesced
+            vec![0..16, 64..80, 1000..1016],                // gapped -> not coalesced
+            vec![100..120, 0..50, 40..60, 0..50, 110..130], // out-of-order, overlapping, duplicate
+            vec![],                                         // empty
+        ];
+
+        for ranges in cases {
+            let got = integration.get_ranges(&location, &ranges).await.unwrap();
```
alamb, line 1792: I don't think this test verifies that the underlying requests were coalesced -- is it possible to do that? Like I think these tests pass even if we revert all the code changes.

src/local.rs
```diff
@@ -502,13 +502,39 @@ impl ObjectStore for LocalFileSystem {
         let path = self.path_to_filesystem(location)?;
         let ranges = ranges.to_vec();
         maybe_spawn_blocking(move || {
-            // Vectored IO might be faster
```
kevinjqliu, line 505: worth keeping this?

kevinjqliu (approved):
LGTM!

src/local.rs
Dandandan: The comment is incorrect, there doesn't exist any "Vectored IO" API for this (well closest would be io_uring)

kevinjqliu: oh yea, makes sense 👍

src/local.rs
alamb: (sorry I forgot to click submit on the PR)

PR #711, opened 2026-05-05
Whitelisting Onelake API & Workspace PL FQDNs

SmritiAgrawal04:
# Rationale for this change
 
1. With the release of the [WS PL FQDNs](https://learn.microsoft.com/en-us/fabric/security/security-workspace-level-private-links-overview#connecting-to-workspaces) with workspace private links, we need to add support for the new workspace FQDNs (“{wsid}.z{xy}.dfs.fabric.microsoft.com), so users can continue to use Delta-RS with OneLake and workspace private links.

# What changes are included in this PR?

1. Valid WS-PL FQDN egs:
* abfss://7c09f7b8-d05f-4a0d-9534-c3cad6e25525@7c09f7b8d05f4a0d9534c3cad6e25525.z7c.msit-blob.fabri
c.microsoft.com/3f49a318-25a0-4100-948d-d08229390322/Files/test_delta_table
* https://c047b3e34e89407a98d7cf9949ae92a3.zc0.onelake.fabric.microsoft.com
* https://c047b3e34e89407a98d7cf9949ae92a3.zc0.dfs.fabric.microsoft.com
* https://7c09f7b8d05f4a0d9534c3cad6e25525.z7c.msit-blob.fabric.microsoft.com/7c09f7b8-d05f-4a0d-9534-c3cad6e25525/3f49a318-25a0-4100-948d-d08229390322/Files/Readme.txt
2. Invalid WS-PL FQDN egs:
* https://c047b3e34e89407a98d7cf9949ae92a3.z12.blob.fabric.microsoft.com // xy mismatch 
* https://c047b3e34e89407a98d7cf9949ae923.z12.blob.fabric.microsoft.com  // workspaceid 31 chars

# Testing Details

1. Unit Tests
 - _parse_url()_ correctly extracts account_name, endpoint, container_name, and sets use_fabric_endpoint = true for 
WS-PL hosts
 - Validates both abfss:// and https:// schemes with WS-PL FQDN patterns like 
{workspaceid}.z{xy}.msit-dfs.fabric.microsoft.com

2. Integration Testing
- In a VM where the workspace supporting private link is hosted, successfully ran newly added integration test on URLs like:
~ https://7c09f7b8d05f4a0d9534c3cad6e25525.z7c.msit-dfs.fabric.microsoft.com/7c09f7b8-d05f-4a0d-9534-c3cad6e25525/3f49a318-25a0-4100-948d-d08229390322/Files/integration_test_wspl.txt
[...]

PR #780, opened 2026-06-23
fix(aws): recognize S3 dual-stack endpoints in URL parsing

### LINKED ISSUE #779: Support for S3 dual-stack endpoints
**Describe the bug**

S3 supports both standard endpoints, e.g. `s3.REGION.amazonaws.com`, as well as a dual-stack endpoint variant, which allows clients to resolve either IPv4/v6, e.g. `s3.dualstack.REGION.amazonaws.com`

**To Reproduce**

The following unit test using the `DefaultObjectStoreRegistry` fails to properly resolve a store for a dualstack endpoint:

```
    #[test]
    fn test_resolve_s3_dualstack_endpoint() {
        let registry = DefaultObjectStoreRegistry::new();

        let url = Url::parse(
            "https://s3.dualstack.us-east-1.amazonaws.com/static.customizedgirl.com/images/design/3747d20127af39238a380d4fface92b1_1639568_0_bigger.jpg",
        )
        .unwrap();

        let (_store, path) = registry
            .resolve(&url)
            .expect("registry should resolve S3 dual-stack endpoints");

        // The bucket (`static.customizedgirl.com`) is the first path segment and
        // should be stripped, leaving the object key as the trailing path.
        assert_eq!(
            path.as_ref(),
            "images/design/3747d20127af39238a380d4fface92b1_1639568_0_bigger.jpg"
        );
    }
```

**Expected behavior**

The registry should return an `AmazonS3` store with the dualstack endpoint configured and the `us-east-1` region set.

**Additional context**

I noticed this while trying to write some code that downloads data from the [PIXMO dataset](https://huggingface.co/datasets/allenai/pixmo-cap) on HuggingFace, which is a collection of image URLs and descriptions.

The repro above uses the URL which failed for me.

a10y:

- Closes #779

Adds support for parsing and resolving `AmazonS3` stores for S3 dual-stack endpoints. Adds cases to the existing URL parsing tests.

I encountered this trying to use `object_store` to read some data from a HuggingFace dataset, where one column of image URLs pointed to many values like `https://s3.dualstack.us-east-1.amazonaws.com/....`

In the first table of the [S3 endpoint docs](https://docs.aws.amazon.com/general/latest/gr/s3.html) you can see an example of these:

image

URLs that previously would error will now resolve.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -625,6 +625,8 @@ impl AmazonS3Builder {
     /// - `s3a://<bucket>/<path>`
     /// - `https://s3.<region>.amazonaws.com/<bucket>`
     /// - `https://<bucket>.s3.<region>.amazonaws.com`
+    /// - `https://s3.dualstack.<region>.amazonaws.com/<bucket>`
+    /// - `https://<bucket>.s3.dualstack.<region>.amazonaws.com`
     /// - `https://ACCOUNT_ID.r2.cloudflarestorage.com/bucket`
     ///
     /// Note: Settings derived from the URL will override any others set on this builder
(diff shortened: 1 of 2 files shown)


kevinjqliu (approved):
LGTM. TIL about dualstack.

My only nit is around strict URL parsing. I wonder if we should be as loose as possible when parsing URLs, rather than trying to determine what is “valid” ourselves. We can let the server make that decision.

alamb:
> My only nit is around strict URL parsing. I wonder if we should be as loose as possible when parsing URLs, rather than trying to determine what is “valid” ourselves. We can let the server make that decision.

I am not quite sure what you are suggesting  @kevinjqliu. 

Are you thinking maybe preserve the original URL in the builder rather than try and recreate it? If so that seems reasonable to me

However, I think this PR is consistent with the exsting pattern in the code -- so we can proceed with this one while we can investigate more flexible/futureproof configuration as a follow on PR / issue

alamb (approved):
Looks good to me -- thank you @a10y and @kevinjqliu

kevinjqliu:
Please proceed with this one! The change is reasonable since the "dual-stack endpoint" conforms to a pre-defined shape. 

I was pointing out that `parse_url` currently rejects urls that does not conform to the pre-defined shapes. To support "dual-stack endpoint", we have to add it as a approved shape. Alternatively, we can allow `parse_url` to passthrough any given `url` it does not recognize and set `self.endpoint` accordingly. This way, perhaps we did not need to add the special logic for "dual-stack endpoint" and can also future proof any other url format AWS decides to create. 

This was an idea based on #711 where we once again had to allowlist a pre-defined shape into the url parsing logic for Azure/OneLake. 

EDIT: I created https://github.com/apache/arrow-rs-object-store/issues/793. Let's move the conversation there 😄

alamb:
>  EDIT: I created https://github.com/apache/arrow-rs-object-store/issues/793. Let's move the conversation there 😄

Thank you 🙏

PR #781, opened 2026-06-25
Cache Azure user delegation key for SAS signing

### LINKED ISSUE #800: Cache Azure user delegation key for SAS signing
Reported by @emilk on https://github.com/apache/arrow-rs-object-store/pull/781:

`AzureClient::signer()` calls `get_user_delegation_key()` — a `GetUserDelegationKey` network round-trip (`POST /?restype=service&comp=userdelegationkey`) — on **every** `signed_url` / `signed_urls` call. A user delegation key is reusable until its expiry, so re-fetching it per request is wasteful and, under load, gets throttled by Azure (HTTP 503 `ServerBusy`).

Notably the other backends do **not** make an uncached network call while signing: the AWS signer signs locally from its cached credential, and GCP caches its signing credentials. Azure is the outlier — it makes a `GetUserDelegationKey` call on top of the already-cached AAD token.

We hit this in production: a workload issuing many presigned-URL requests against Azure Blob drove ~100k `GetUserDelegationKey` POSTs in 2h (~840/min), ~35% of which returned 503. The retry client backed off 10× (~10s) and then surfaced the error to the caller.

**Additional context**

emilk:
# Rationale for this change
- closes https://github.com/apache/arrow-rs-object-store/issues/800

`AzureClient::signer()` calls `get_user_delegation_key()` — a `GetUserDelegationKey` network round-trip (`POST /?restype=service&comp=userdelegationkey`) — on **every** `signed_url` / `signed_urls` call. A user delegation key is reusable until its expiry, so re-fetching it per request is wasteful and, under load, gets throttled by Azure (HTTP 503 `ServerBusy`).

Notably the other backends do **not** make an uncached network call while signing: the AWS signer signs locally from its cached credential, and GCP caches its signing credentials. Azure is the outlier — it makes a `GetUserDelegationKey` call on top of the already-cached AAD token.

We hit this in production: a workload issuing many presigned-URL requests against Azure Blob drove ~100k `GetUserDelegationKey` POSTs in 2h (~840/min), ~35% of which returned 503. The retry client backed off 10× (~10s) and then surfaced the error to the caller.

Reuse the existing `TokenCache<T>` to cache the user delegation key on `AzureClient`:

- A long-lived key (12h) is fetched once and reused to sign many short-lived SAS tokens.
- `TokenCache` only returns a key with more than its `min_ttl` (2h here) remaining, so any SAS no longer than that is guaranteed to expire before its key (a SAS must not outlive the key it is signed with). The rare longer-lived SAS fetch a dedicated key.
- The cache entry expires when the key Azure actually granted does (`SignedExpiry`), not when we requested.
- Inherits `TokenCache`’s refresh-race double-check and fetch backoff.

`TokenCache::with_min_ttl` is extended to the `azure-base` feature so the Azure client can configure the cache (previously gated to `aws-base`/`gcp-base`).

The signed SAS tokens themselves are unchanged (still caller-specified `expires_in`).

No public API changes. Presigned-URL generation against Azure now performs far fewer `GetUserDelegationKey` requests.

# Tests

Added `azure::client::tests::test_delegation_key_expiry` for the granted-vs-requested expiry parsing. The caching/refresh machinery itself is covered by the existing `client::token` tests.

# Related issues / PRs

No existing issue tracks this specifically. Related:

- #101 — *Azure Signed URL Support* (closed): the feature that introduced Azure user-delegation-key SAS signing, i.e. the `GetUserDelegationKey` call this PR caches.
- #771 — *presigned URLs with extra query params and signed headers* (open): also touches the Azure presigning path (`azure/mod.rs` `signed_url`/`signed_urls`); orthogonal change, flagging for awareness of a possible minor merge interaction.

### DIFF
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -1461,7 +1461,7 @@ impl BlockList {
     }
 }
 
-#[derive(Debug, Clone, Default, PartialEq, Deserialize)]
+#[derive(Clone, Default, PartialEq, Deserialize)]
 #[serde(rename_all = "PascalCase")]
 pub(crate) struct UserDelegationKey {
     pub signed_oid: String,
@@ -1471,6 +1471,29 @@ pub(crate) struct UserDelegationKey {
     pub signed_service: String,
     pub signed_version: String,
     pub value: String,
+}
+
+impl std::fmt::Debug for UserDelegationKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let Self {
+            signed_oid,
+            signed_tid,
+            signed_start,
+            signed_expiry,
+            signed_service,
+            signed_version,
+            value: _, // secret => redacted
+        } = self;
+        f.debug_struct("UserDelegationKey")
+            .field("signed_oid", signed_oid)
+            .field("signed_tid", signed_tid)
+            .field("signed_start", signed_start)
+            .field("signed_expiry", signed_expiry)
+            .field("signed_service", signed_service)
+            .field("signed_version", signed_version)
+            .field("value", &"******")
+            .finish()
+    }
 }
 
 #[cfg(test)]
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -23,6 +23,7 @@ use crate::client::get::GetClient;
 use crate::client::header::{HeaderConfig, get_put_result};
 use crate::client::list::ListClient;
 use crate::client::retry::{RetryContext, RetryExt};
+use crate::client::token::{TemporaryToken, TokenCache};
 use crate::client::{
     CryptoProvider, DigestAlgorithm, GetOptionsExt, HttpClient, HttpError, HttpRequest,
     HttpResponse, crypto_provider,
@@ -47,7 +48,7 @@ use rand::RngExt;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use url::Url;
 
 const VERSION_HEADER: &str = "x-ms-version-id";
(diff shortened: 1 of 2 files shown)


src/azure/client.rs
```diff
@@ -662,16 +663,47 @@ async fn parse_blob_batch_delete_body(
...
 #[derive(Debug)]
 pub(crate) struct AzureClient {
     config: AzureConfig,
     client: HttpClient,
+    /// Caches the user delegation key used to sign SAS URLs.
+    ///
+    /// Fetching a key is a network round-trip (`GetUserDelegationKey`) that Azure
+    /// throttles under load, so we fetch a long-lived key once and reuse it to
+    /// mint many short-lived SAS tokens.
+    delegation_key_cache: TokenCache<UserDelegationKey>,
```
kevinjqliu, line 699: we should redact these, similar to aws/credentials.rs

https://github.com/apache/arrow-rs-object-store/blob/93f6f88f4d2d58e89c93f917db05ebf3b24825b7/src/aws/credential.rs#L83-L91

maybe some of this logic should live in azure/credentials.rs

emilk: Good catch! I've changed the `Debug for UserDelegationKey` implementation to redact the secret

alamb:
> Notably the other backends do **not** make an uncached network call while signing: the AWS signer signs locally from its cached credential, and GCP caches its signing credentials. Azure is the outlier — it makes a `GetUserDelegationKey` call on top of the already-cached AAD token.

Agreed -- they use TokenCache 

S3:
https://github.com/apache/arrow-rs-object-store/blob/09adc41b05faada6b233746d866e0ab63a56f208/src/aws/credential.rs#L759

GCP: https://github.com/apache/arrow-rs-object-store/blob/09adc41b05faada6b233746d866e0ab63a56f208/src/client/mod.rs#L1053

src/azure/client.rs
```diff
@@ -662,16 +663,47 @@ async fn parse_blob_batch_delete_body(
     Ok(results)
 }
 
+/// How long a freshly fetched user delegation key is requested to remain valid.
+///
+/// The SAS tokens we sign with it stay short-lived; this only bounds how often
+/// we call `GetUserDelegationKey`. Azure caps the key lifetime at 7 days.
```
alamb: Can you add a URL that with some evidence for this claim? It seems this could be a good one

  https://learn.microsoft.com/en-us/rest/api/storageservices/get-user-delegation-key#request-body


> Start Required. The start time for the user delegation SAS, in ISO date format. It must be a valid date and time within seven days of the current date.

src/azure/client.rs
```diff
@@ -662,16 +663,47 @@ async fn parse_blob_batch_delete_body(
     Ok(results)
 }
 
+/// How long a freshly fetched user delegation key is requested to remain valid.
+///
+/// The SAS tokens we sign with it stay short-lived; this only bounds how often
```
alamb: I think it would help to define SAS at least once (I realize it is used elsewhere in this module without that)

```suggestion
/// The shared access signature (SAS) tokens we sign with it stay short-lived; this only bounds how often
```

src/azure/client.rs
```diff
@@ -1043,6 +1075,50 @@ impl AzureClient {
...
+                .get_or_insert_with(|| self.fetch_delegation_key(DELEGATION_KEY_VALIDITY))
+                .await
+        } else {
+            self.get_user_delegation_key(&sas_start, &sas_expiry).await
+        }
+    }
+
+    /// Fetch a user delegation key valid for `validity` and wrap it as a
+    /// [`TemporaryToken`] so [`TokenCache`] can expire it.
+    async fn fetch_delegation_key(
```
alamb: I found the naming somewhat confusing here (as it wasn't clear to me what the relationship between `get_user_delegation_key` and `fetch_user_delegation_key` was (`get` and `fetch` are very similar). I wa sall the more confused because `get` actually does the network call, when I would normally exepct `fetch` to do that

Something like this might be easier to understand
* fetch_delegation_key --> get_delegation_key
* (current) get_delegation_key -->  get_delegation_key_inner

alamb (approved):
Thanks @emilk  -- I went through this carefully and I think it makes sense to me

I left a few small comments but I don't think any of them are required for merge (we can do them as follow on PRs for example)

The one thing I would like to see before we merge this is some test of the actual flow (e.g. that cached credentials are actually fetched and reused).  I took the liberty of pushing such a test -- please give it a look

Also, @kevinjqliu as our resident Azure expert, does this behavior make sense to you?

alamb:
I took the liberty of pushing the comment changes and function name changes. I'll plan to merge this once CI passes

Thanks @emilk

kevinjqliu:
I verified this with my own azure storage account, it works great. Thanks for adding this! 

I think there are a few follow-ups could make the delegation-key cache more flexible and robust:
- Make `DELEGATION_KEY_VALIDITY` optionally configurable so users can limit how long reusable keys remain valid. This setting should never shorten a key below the requested SAS lifetime.
- Remove the fixed `DELEGATION_KEY_MIN_TTL`. Azure returns the key’s actual `SignedExpiry`, so cache it and reuse the key only when `SignedExpiry` is later than the requested SAS expiry. Otherwise, fetch a new key.
- Treat Azure’s returned `SignedStart` and `SignedExpiry` as authoritative. Return a clear error if they are malformed or do not cover the full SAS lifetime.


EDIT: tracking as #807

alamb:
Thank you @emilk and @kevinjqliu

Issue #774, opened 2026-06-21
GCS: `get()` fails with `MissingContentLength` on chunked responses (`Content-Encoding: gzip`)


nkemnitz:
**Describe the bug**

`ObjectStore::get` on the GCS backend fails with `Error::Generic { store: "GCS", source: Header { source: MissingContentLength } }` for any object stored with `Content-Encoding: gzip`. GCS serves these via `Transfer-Encoding: chunked` with **no `Content-Length`**, and object_store treats a missing `Content-Length` as fatal.

It manifests in two ways, and crucially **a client cannot fully avoid it**:

- **Default reads (no `Accept-Encoding`):** GCS applies *decompressive transcoding* — it decompresses the object server-side and streams the result chunked, with no `Content-Length`. Every gzip object fails, at any size.
- **Even with `Accept-Encoding: gzip`** (which returns the raw stored bytes): objects whose **stored size exceeds ~8 MiB** are *still* served chunked with no `Content-Length` (empirically the cutover is between 9 and 10 MB), so they fail too. Only *small* objects read with `Accept-Encoding: gzip` succeed.

So whether you receive the transcoded (uncompressed) bytes or the raw (compressed) bytes, the response is chunked with no `Content-Length` — and object_store rejects it before returning any data. `head()` and range reads on gzip objects fail as well.

These are valid HTTP/1.1 responses: a chunked body is self-delimiting, and [RFC 9112 §6.2](https://www.rfc-editor.org/rfc/rfc9112#section-6.2) states a sender **MUST NOT** send `Content-Length` together with `Transfer-Encoding`;
[RFC 9112 §6.3](https://www.rfc-editor.org/rfc/rfc9112#section-6.3) says the chunked framing determines the body length and overrides any `Content-Length`. object_store is therefore requiring a header the spec forbids on exactly these responses.

**To Reproduce**

Observe the offending response with no credentials against Google's public demo data (the `?cb=` cache-buster forces origin past the public edge cache):

```console
$ curl -sD - -o /dev/null \
  "https://storage.googleapis.com/neuroglancer-public-data/kasthuri2011/ground_truth/6_6_30/5376-5440_6656-6720_896-960?cb=$RANDOM" \
  | grep -iE "HTTP/|content-length|transfer-encoding|x-goog-stored-content-encoding"

HTTP/2 200
transfer-encoding: chunked
x-goog-stored-content-encoding: gzip
# (no content-length)
```

[...]

PR #782, opened 2026-06-25
fix: accept GCS gzip responses without Content-Length

nkemnitz:

Part of #774 (does not fully close it — see below).

# Rationale for this change

GCS serves large objects stored with `Content-Encoding: gzip` using chunked transfer with no `Content-Length` (and decompressive transcoding when the client does not accept gzip encoding). `ObjectStore::get`/`head` on GCS required
`Content-Length` unconditionally and failed with `Generic { store: "GCS", source: Header { source: MissingContentLength } }`, even though a chunked, self-delimiting body is a valid response (RFC 9112 §6.2 forbids `Content-Length` alongside `Transfer-Encoding: chunked`).

`HeaderConfig` gains `stored_size_header: Option<&'static str>`. When `Content-Length` is absent, `header_meta` reads the object size from this header. GCS sets it to `x-goog-stored-content-length` (always present); S3, Azure and the
HTTP store leave it `None`, so a missing `Content-Length` stays a hard error for them.

`get()`/`head()` now succeed on chunked gzip GCS objects. On a server-decompressed (transcoded) read, `ObjectMeta.size` is the *stored* (compressed) size, since the decompressed length is not known without reading the body; on a passthrough read (`Accept-Encoding: gzip`) it is exact.

**Not fully resolved:** some transcoded GCS responses (default reads without `Accept-Encoding: gzip`) also omit the ETag entirely and still fail with `MissingEtag`. Left for a follow-up.

-------------------

**:robot: AI disclaimer**:
All the code written by Claude. I made the changes as targeted and minimal as possible, for now only focusing on the chunked encoding, because that's my major blocker. Decompressive transcoding feels kind of niche. And the ETag handling involves some more thought and knowledge about this repo. E.g. I think the different Cloud vendors rely on custom metadata `version` headers to allow resuming downloads, rather than the ETag(?)...

### DIFF
--- a/src/client/header.rs
+++ b/src/client/header.rs
@@ -36,6 +36,14 @@ pub(crate) struct HeaderConfig {
     /// Defaults to `true`
     pub last_modified_required: bool,
 
+    /// Header to read the object size from when `Content-Length` is absent.
+    ///
+    /// GCS omits `Content-Length` on chunked `Content-Encoding: gzip` responses — large
+    /// bodies, or decompressive transcoding — but always sends `x-goog-stored-content-length`.
+    /// Stores that always send a `Content-Length` (S3, Azure) leave this `None`, so a missing
+    /// `Content-Length` stays an error for them.
+    pub stored_size_header: Option<&'static str>,
+
     /// The version header name if any
     pub version_header: Option<&'static str>,
 
@@ -139,8 +147,12 @@ pub(crate) fn header_meta(
         Err(e) => return Err(e),
     };
 
+    // Prefer `Content-Length`, falling back to a store-provided size header: GCS omits
+    // `Content-Length` on chunked gzip responses (large bodies, or transcoding) but always
+    // sends `x-goog-stored-content-length`. Stores without such a header still require it.
     let content_length = headers
         .get(CONTENT_LENGTH)
+        .or_else(|| cfg.stored_size_header.and_then(|h| headers.get(h)))
         .ok_or(Error::MissingContentLength)?;
 
     let content_length = content_length
--- a/src/aws/client.rs
+++ b/src/aws/client.rs
@@ -922,6 +922,7 @@ impl GetClient for S3Client {
     const HEADER_CONFIG: HeaderConfig = HeaderConfig {
         etag_required: false,
         last_modified_required: false,
+        stored_size_header: None,
         version_header: Some(VERSION_HEADER),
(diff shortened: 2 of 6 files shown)


src/client/header.rs
```diff
@@ -139,8 +147,12 @@ pub(crate) fn header_meta(
         Err(e) => return Err(e),
     };
 
+    // Prefer `Content-Length`, falling back to a store-provided size header: GCS omits
```
alamb, line 150: is there documentation on this behavior you can link to so we can verify this?

alamb (commented):
Thanks @nkemnitz  -- this seems ok to me, but I am not a GCS expert

Can you please provide some documentation / evidence that GCS behaves the way that this PR's description claims? Ideally we can include those links in the docs for future readers who find the behavior?

src/client/header.rs
nkemnitz: The description for what `x-goog-stored-content-length` contains: https://docs.cloud.google.com/storage/docs/xml-api/reference-headers#xgoogstoredcontentlength

In the Google Cloud Rust repo, a code maintainer reports that this header is indeed always present: https://github.com/googleapis/google-cloud-rust/discussions/3713#discussioncomment-14883469

The part that confirms that `Content-Length` and `Content-Encoding` get dropped for decompressive transcoding: https://docs.cloud.google.com/storage/docs/transcoding#decompressive_transcoding
(Although I don't really care about decompressive transcoding - it creates large transfers and as described under the same link in the paragraph below, it prevents integrity checks)

I could not find any official documentation regarding the chunked Transfer-Encoding switch... only can see that it does... Could weaken that claim to "Google _may_ omit the `Content-Length`".


Other finds:
* [google-cloud-cpp](https://github.com/googleapis/google-cloud-cpp/blob/95511adfa1a6e733ca20dd17aaefd898611fe2f5/google/cloud/storage/internal/rest/object_read_source.cc#L47-L68) prefers `x-goog-stored-content-length` > `content-range` > `content-length`
* [google-cloud-rust](https://github.com/googleapis/google-cloud-rust/blob/18cd4258a67b424a814d29c4b63fad501952c43d/src/storage/src/storage/read_object/parse_http_response.rs#L42) only uses x-goog-stored-content-length and nothing else?
* Their [python code](https://github.com/googleapis/python-storage/blob/ab4997ce0f7b85947e84b226bd0edf6d714a946a/google/cloud/storage/_media/requests/download.py#L163) seems to use it to distinguish truncated downloads from transcoded ones

PR #784, opened 2026-06-30
feat: add deregister to ObjectStoreRegistry

kszucs:

No tracking issue; happy to file one if preferred.

# Rationale for this change

`ObjectStoreRegistry` can `register` and `resolve` stores but has no way to remove one. Consumers that register stores dynamically (e.g. per-session) need to deregister them as well.

- Add `ObjectStoreRegistry::deregister(&self, url: &Url) -> Option<Arc<dyn ObjectStore>>`, the inverse of `register`, with a default implementation returning `None`.
- Implement it for `DefaultObjectStoreRegistry`: removes the store at the exact scheme/authority/path of `url` (siblings and longer prefixes are left in place) and prunes emptied tree branches.
- Tests covering exact removal, sibling/longer-prefix preservation, and unknown URLs.

Yes: a new `ObjectStoreRegistry::deregister` trait method. It has a default implementation, so existing implementors are unaffected.

### DIFF
--- a/src/registry.rs
+++ b/src/registry.rs
@@ -82,6 +82,36 @@ pub trait ObjectStoreRegistry: Send + Sync + std::fmt::Debug + 'static {
     /// assert!(Arc::ptr_eq(&ret, &bucket3));
     /// ```
     fn resolve(&self, url: &Url) -> crate::Result<(Arc<dyn ObjectStore>, Path)>;
+
+    /// Deregister the store registered for exactly `url`, returning it if present.
+    ///
+    /// Unlike [`resolve`](Self::resolve), which matches the longest registered
+    /// path that is a prefix of the object URL, this removes the store
+    /// registered at the *exact* scheme, authority and path of `url` (the
+    /// inverse of [`register`](Self::register)). A store registered at a
+    /// different path under the same authority is left in place.
+    ///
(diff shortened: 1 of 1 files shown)


kylebarron (approved):

src/registry.rs
```diff
@@ -215,6 +261,19 @@ impl ObjectStoreRegistry for DefaultObjectStoreRegistry {
 
         Err(Error::NotFound.into())
     }
+
+    fn deregister(&self, url: &Url) -> Option<Arc<dyn ObjectStore>> {
+        let mut map = self.map.write();
```
alamb, line 266: minor: you could move the lock acquisition after computing the key (so right before `map.get_mut()`) to minimize the time the lock is held

alamb (approved):
Looks good to me -- thank you @kszucs and @kylebarron

src/registry.rs
kszucs: Thank you! Created a PR for it https://github.com/apache/arrow-rs-object-store/pull/792

alamb:
Filed a ticket for this
- https://github.com/apache/arrow-rs-object-store/issues/795

(so that it shows up nicely in the release notes)

PR #789, opened 2026-07-04
impl HttpError::new_boxed to provide user with ability to ctor from already boxed error

DoumanAsh:

None, it is minor improvement to API

# Rationale for this change
 
I would like to be able to created HttpError from already boxed error for my alternative HTTP client implementation
Otherwise I need to wrap boxed error into new type

Introduces new method to HttpError to create error from boxed `dyn std::error::Error`

New public method on `HttpError`
No breaking changes

### DIFF
--- a/src/client/http/connection.rs
+++ b/src/client/http/connection.rs
@@ -81,10 +81,13 @@ impl HttpError {
     where
         E: Error + Send + Sync + 'static,
     {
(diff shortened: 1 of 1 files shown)


alamb (approved):
Looks like a nice improvement to me -- thank you @DoumanAsh

PR #790, opened 2026-07-06
Do not explicitly add `Host` header when making requests with  S3Builder

### LINKED ISSUE #706: Support for HF Storage Buckets
Hugging Face now offers Storage Buckets. These are described as "S3-like" but I think are actually implemented using Xet. Would you consider adding an implementation to support this backend in object-store? Thanks!

https://huggingface.co/storage

https://huggingface.co/docs/hub/storage-buckets

ntjohnson1:

- Closes #706 (if sufficient for the request)

I included a commit with the exemplar I was using to verify arrow-rs-object-store works with HF buckets (to demonstrate things work if you want to try it yourself)

# Rationale for this change
With hugging face buckets S3 compatible API you can create a builder via:
```rust
let mut builder = AmazonS3Builder::new()
        .with_endpoint(endpoint)
        .with_region(region)
        .with_bucket_name(bucket)
        .with_access_key_id(env_req("AWS_ACCESS_KEY_ID"))
        .with_secret_access_key(env_req("AWS_SECRET_ACCESS_KEY"))
        .with_disable_bulk_delete(disable_bulk_delete);
```
However, when I did this and tried connecting to a bucket I was getting issues with the redirect.

object_store's SigV4 signer inserted an explicit Host header on every request. reqwest carries user-set headers across redirects, so when an S3-compatible gateway answers GetObject with a cross-host 302 ( e.g. the Hugging Face gateway: s3.hf.co redirecting to its Xet CDN: cas-bridge.xethub.hf.co) the stale Host: s3.hf.co header poisoned the redirected request and the download failed with a connection error. list/head worked (no redirect); every get failed.

Host is still part of the SigV4 signature, but it's now added to a throwaway copy of the headers used only to compute the signature, instead of being pinned on the outgoing request. The actual Host header is left to hyper, which derives it from the URL and regenerates it on redirect. The produced signature is identical, so AWS S3 and all other providers are unaffected.

No


src/aws/credential.rs
```diff
@@ -262,7 +258,14 @@ impl<'a> AwsAuthorizer<'a> {
                 .insert(&REQUEST_PAYER_HEADER, REQUEST_PAYER_HEADER_VALUE.clone());
         }
 
-        let (signed_headers, canonical_headers) = canonicalize_headers(request.headers());
+        // SigV4 must sign the `host` header, but the actual Host header is managed by the HTTP
+        // transport (hyper derives it from the request URL). We therefore add `host` to a
+        // throwaway copy of the headers used only to build the signature, rather than pinning it
+        // on the request.
+        let host = &url[url::Position::BeforeHost..url::Position::AfterPort];
+        let mut headers_to_sign = request.headers().clone();
```
alamb: I personally think this clone/sign pattern is confusing -- it took me a while to grok what this is doing and realize the net effect of this is to remove the HOST from the headers in reqwest

I think it would be both clearer and more performant (no `clone`) to do something more like

```rust
  request.headers_mut().insert(HOST, host);
  let sig = compute_signature(request.headers())?;
  request.headers_mut().remove(HOST);
```

src/aws/credential.rs
```diff
@@ -262,7 +258,14 @@ impl<'a> AwsAuthorizer<'a> {
                 .insert(&REQUEST_PAYER_HEADER, REQUEST_PAYER_HEADER_VALUE.clone());
         }
 
-        let (signed_headers, canonical_headers) = canonicalize_headers(request.headers());
+        // SigV4 must sign the `host` header, but the actual Host header is managed by the HTTP
+        // transport (hyper derives it from the request URL). We therefore add `host` to a
+        // throwaway copy of the headers used only to build the signature, rather than pinning it
+        // on the request.
```
alamb: I personally found this comment hard to understand. Mostly because it asserted that the actual Host header is managed by the transport when I think it is more accurate to say reqwest will add the `Host` header if not already present. It also doesn't mention the redirect problem this PR is trying to solve 

So maybe something more like

```suggestion
        // the SigV4 must include a value for `host`, but the actual Host header may need to change 
        // due to a redirect. Therefore do not override the Host header for the http 
        // request, let the client set it. 
```

alamb (approved):
Thank you @ntjohnson1  -- this seems ok to me, though I found the code and comments a bit confusing (I left some suggestions to improve them)

I also don't understand what is special about Hugging face -- does this mean S3 requests that return redirect's aren't working at all (for any provider)?

What I am most worried about is that removing the Host header from the request headers will mess up some other existing users. 

However, I spent a bunch of time researching this and I think this is ok -- under the covers if the HeadersMap from reqwest don't have Headers, it is set by the lower level hyper library, so I think all is good





BTW it took me a while to verify the assertion in your  comments:

> the actual Host header is managed by the HTTP transport (hyper derives it from the request URL). 

But it took me a while to get codex to prove to me that this is the case

  - object_store sends requests through reqwest::Client here: src/client/http/connection.rs:217
  - reqwest 0.13.4 defines its internal client type as hyper_util::client::legacy::Client:
    https://github.com/seanmonstar/reqwest/blob/v0.13.4/src/async_impl/client.rs#L2473

  - reqwest builds that client here:
    https://github.com/seanmonstar/reqwest/blob/v0.13.4/src/async_impl/client.rs#L939-L940

  - reqwest’s Cargo.toml explicitly enables hyper-util’s client-legacy feature:
    https://github.com/seanmonstar/reqwest/blob/v0.13.4/Cargo.toml#L346-L354

  - That hyper-util client defaults set_host: true:
    https://github.com/hyperium/hyper-util/blob/v0.1.20/src/client/legacy/client.rs#L1031-L1036

  - And when sending HTTP/1 requests, it inserts HOST if absent, derived from req.uri():
    https://github.com/hyperium/hyper-util/blob/v0.1.20/src/client/legacy/client.rs#L298-L309

ntjohnson1:
Thanks for an in-depth review as always.

> I also don't understand what is special about Hugging face -- does this mean S3 requests that return redirect's aren't working at all (for any provider)?

I suspect the more general title makes more sense. I'm not sure how popular the redirects are for s3 like apis. We mostly just use the major supported ones here today but are explore R2/B2 etc starting with HF buckets.

That's weird Codex had such a hard time digging through to the root cause. Claude seemed to stumble on it pretty easily, next time I can provide the better trace through for easier ability to verify.

> What I am most worried about is that removing the Host header from the request headers will mess up some other existing users.

I'm kind of relying on the existing tests CI to cover this here. Alternatively I could try to include this branch as a patch on a system we use with aws to see if things still go through correctly over there (but would probably take some time to get that setup). I don't have super in depth familiarity here to gauge the risk.

alamb:
> That's weird Codex had such a hard time digging through to the root cause. Claude seemed to stumble on it pretty easily, next time I can provide the better trace through for easier ability to verify.

Well, it is more like its description of "hyper just does this" was not convincing to me and I made it come up with the exact call chai

alamb:
clippy failure seems unrelated

Update
- https://github.com/apache/arrow-rs-object-store/issues/798

alamb:
Merged up to get a clean CI run

alamb:
Thanks again @ntjohnson1

PR #792, opened 2026-07-08
chore: reduce lock hold time in DefaultObjectStoreRegistry::deregister

kszucs:

N/A

# Rationale for this change

`deregister` acquired the write lock before computing the URL key and
path segments, unnecessarily extending the time the lock is held.

Move the `url_key` and path segment computation ahead of the write
lock acquisition in `DefaultObjectStoreRegistry::deregister`, so the
lock is only held for the actual map mutation.

No.


kylebarron (approved):

alamb:
🚀 

Thanks @kszucs and @kylebarron

PR #801, opened 2026-07-10
Add doc example for multipart upload to AmazonS3::create_multipart

alamb:
- Follow on from 

# Rationale for this change
 
While reviewing a PR from @zfarrell
- https://github.com/apache/arrow-rs-object-store/pull/771

I found that there was not good example of actually using the mulit-part API in the docs (so there is no good place to add an example of using pre-signed URLs)

Update docs / examples

docs only


kevinjqliu (approved):
LGTM!

kylebarron (approved):

alamb:
Thank you for the review @kevinjqliu and @kylebarron

PR #802, opened 2026-07-10
Add doc example for multipart upload to `MicrosoftAzure::create_multipart`

alamb:
# Rationale for this change

The low-level `MultipartStore` API had no worked example showing the full `create_multipart` → `put_part` → `complete_multipart` flow. This adds one for Azure, mirroring the S3 example in apache/arrow-rs-object-store#801.

- A runnable (`no_run`) doc example on `MicrosoftAzure::create_multipart` showing the full low-level flow.
- Documents Azure-specific behavior

Documentation only. No API changes.

### DIFF
--- a/src/azure/mod.rs
+++ b/src/azure/mod.rs
@@ -304,6 +304,59 @@ impl MultipartUpload for AzureMultiPartUpload {
 
 #[async_trait]
 impl MultipartStore for MicrosoftAzure {
+    /// Create a new multipart upload, returning its [`MultipartId`].
+    ///
+    /// This is the low-level [`MultipartStore`] API, which gives direct control
+    /// over individual parts. See [`ObjectStoreExt::put_multipart`] for a
+    /// higher-level API that handles parts automatically.
+    ///
+    /// # Example
+    ///
+    /// Create a multipart upload, upload two parts, and finalize it by calling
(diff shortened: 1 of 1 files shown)


src/azure/mod.rs
```diff
@@ -304,6 +304,59 @@ impl MultipartUpload for AzureMultiPartUpload {
...
+    /// // Finalize the upload. The parts must be provided in `part_idx` order.
+    /// azure.complete_multipart(&path, &id, vec![part0, part1]).await?;
+    /// #     Ok(())
+    /// # }
+    /// ```
+    ///
+    /// Azure stores each part as a block, so a single blob may contain at most
+    /// 50,000 parts, each up to 4,000 MiB. Parts may be uploaded concurrently and
+    /// in any order, provided each is given the correct `part_idx`. See
+    /// [Azure block blob limits] for the full set of size and count constraints.
```
alamb, line 350: Image

kevinjqliu: from https://learn.microsoft.com/en-us/rest/api/storageservices/put-block 

which is already linked in the comment, 👍

kevinjqliu (approved):
LGTM i tested this against my azure storage account, works like a charm

alamb:
Thanks @kevinjqliu

PR #815, opened 2026-07-25
Support R2 URLs with jurisdictions

### LINKED ISSUE #814: Support URLs for jurisdiction-restricted buckets on R2

Creating a jurisdiction-restricted bucket on R2 yields a URL that is not compatible with the current S3 URL parser. The current workaround is to pass the bucket and the endpoint manually, but it could be quite confusing for people to copy the URL from the Cloudflare dashboard and not having it work despite that the crate is R2-compatible in general.

Accept and parse such URL.

Yielding an error message that explicitly prompts to pass URL parts separately, though it appears to require approximately the same amount of code as accepting the URLs in the first place.

**Additional context**

Upstream docs: https://developers.cloudflare.com/r2/reference/data-location/#jurisdictional-restrictions

Kharacternyk:

Closes #814.

# Rationale for this change
 
Jurisdiction-restricted buckets on R2 are only accessible through special URLs, which are not currently accepted by the parser. Currently there are only two jurisdictions, `eu` and `fedramp`, but there may be more in the future, so I've kept the pattern general.

Extended the URL parser.

Yes, valid R2 URLs for jurisdiction-restricted buckets no longer fail to be parsed. I've updated the relevant doc comment.

### DIFF
--- a/src/aws/builder.rs
+++ b/src/aws/builder.rs
@@ -628,6 +628,7 @@ impl AmazonS3Builder {
     /// - `https://s3.dualstack.<region>.amazonaws.com/<bucket>`
     /// - `https://<bucket>.s3.dualstack.<region>.amazonaws.com`
     /// - `https://ACCOUNT_ID.r2.cloudflarestorage.com/bucket`
+    /// - `https://ACCOUNT_ID.JURISDICTION.r2.cloudflarestorage.com/bucket`
     ///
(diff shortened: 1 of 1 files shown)


Kharacternyk:
The CI fail appears unrelated to the changes here

alamb (approved):
Looks good to me -- thank you @Kharacternyk

PR #817, opened 2026-07-30
Support streaming put payloads

### LINKED ISSUE #281: object_store: Support for streaming body in `put/put_opts`

Currently, `put/put_opts` takes a `PutPayload` which is just a wrapper around an array of `Bytes`. It would be useful (for us) to be able to stream the body without having to do a multipart upload. 

The underlying http library `reqwest` already supports this. So I think we can change the definition of `PutPayload` to an enum like so:

```
type StreamingBody = Box<dyn futures::stream::Stream<Item = Result<Bytes, Box<dyn std::error::Error + Send + Sync + 'static>>> + Send + 'static>;

pub enum PutPayload {
  Bytes(Arc<[Bytes]>),
  Streaming(StreamingBody,usize),
}
```

where `PutPayload::Streaming` holds the stream and the total content length (which must be known ahead of time). 

The (unfortunate) requirement of the S3 API to know the `content-length` up front means this isn't fully general but can be useful in certain situations such as buffering a largeish file on disk before uploading. In that case, a streaming body would allow the upload to not have to load the entire object in memory to upload it. 

This can already be done with `put_multipart` but mutli-part upload is complicated, may have some performance implications for reads.

**Additional context**

senzzzi:

Closes #281.

# Rationale for this change

`PutPayload` currently only supports collections of `Bytes`. As a result, uploading a file through `put` or `put_opts` requires loading the entire file into memory or using a multipart upload.

This change allows callers to stream a payload with a known content length through a single PUT request. File-backed payloads are a primary use case.

- Extends `PutPayload` to support replayable streaming bodies with a known content length.
- Adds `PutPayload::from_stream` for constructing streaming payloads.
- Adds `PutPayload::from_file`, which reads files in 16 KiB chunks by default.
- Adds `PutPayload::from_file_with_chunk_size` for configuring the file chunk size.
- Updates HTTP request bodies to poll payload streams asynchronously.
- Recreates the stream for every request attempt so retryable uploads can resend the complete payload.
- Updates S3 signing and checksum calculation to consume streaming payloads.
- Updates the local filesystem and in-memory backends to accept streaming payloads.
- Adds regression tests covering:
  - the default 16 KiB file chunk size;
  - replaying cloned streaming request bodies; and
  - retrying an upload using a file-backed streaming payload.

Validation performed:

- `cargo test`
- `cargo test --all-features`
- `cargo clippy --all-targets --all-features -- -D warnings`
- compilation with only the `fs` feature
- compilation without default features

Yes. Callers can now upload a file without first loading the complete file into memory:

```rust
let payload = PutPayload::from_file(file_path).await?;
store.put(&location, payload).await?;
```

Custom streaming payloads can be created with PutPayload::from_stream. The stream factory must produce a new stream from the beginning for every invocation, and the
supplied content length must exactly match the number of bytes yielded.

Existing byte-backed PutPayload construction and iteration behavior is unchanged. Synchronous APIs such as iter, AsRef<[Bytes]>, and conversion into Bytes are not supported for streaming payloads; callers should use PutPayload::stream or PutPayload::bytes instead.

## API compatibility consideration

`PutPayload` remains a public struct backed by a private enum so that existing byte-backed construction and usage remain source compatible:

```rust
enum PutPayloadInner {
    Bytes(Arc<[Bytes]>),
    Streaming {
        factory: StreamFactory,
        content_length: usize,
    },
}
```

There is an API-design limitation around the existing synchronous accessors. Methods and trait implementations such as PutPayload::iter, AsRef<[Bytes]>, IntoIterator, and conversion into Bytes cannot synchronously represent an asynchronous, fallible stream.

In this implementation, their existing behavior is unchanged for byte-backed payloads, but calling them with a streaming payload panics. Streaming-aware implementations should instead use PutPayload::stream or PutPayload::bytes.

Changing the existing synchronous APIs to return Option or Result would avoid the panic but would be a breaking public API change. Returning an empty iterator for streaming payloads was rejected because it could silently result in incomplete uploads.

Feedback from maintainers on the preferred long-term API shape would be appreciated. Possible alternatives include:
- accepting the documented limitation for synchronous APIs;
- changing these APIs in a future breaking release; or
- introducing streaming uploads through a separate payload type or method.

### DIFF
--- a/src/client/http/body.rs
+++ b/src/client/http/body.rs
@@ -18,8 +18,8 @@
 use crate::client::{HttpError, HttpErrorKind};
 use crate::{PutPayload, collect_bytes};
 use bytes::Bytes;
-use futures_util::StreamExt;
 use futures_util::stream::BoxStream;
+use futures_util::{StreamExt, TryStreamExt};
 use http_body_util::combinators::BoxBody;
 use http_body_util::{BodyExt, Full};
 use hyper::body::{Body, Frame, SizeHint};
@@ -30,8 +30,16 @@ use std::task::{Context, Poll};
 pub type HttpRequest = http::Request<HttpRequestBody>;
 
 /// The [`Body`] of an [`HttpRequest`]
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct HttpRequestBody(Inner);
+
+impl std::fmt::Debug for HttpRequestBody {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("HttpRequestBody")
+            .field("content_length", &self.content_length())
+            .finish()
+    }
+}
 
 impl HttpRequestBody {
     /// An empty [`HttpRequestBody`]
@@ -46,6 +54,11 @@ impl HttpRequestBody {
             Inner::PutPayload(_, payload) => reqwest::Body::wrap_stream(
                 futures_util::stream::iter(payload.into_iter().map(Ok::<_, HttpError>)),
             ),
+            Inner::Streaming(body) => reqwest::Body::wrap_stream(
+                body.payload
+                    .stream()
+                    .map_err(|e| HttpError::new_boxed(HttpErrorKind::Unknown, e)),
+            ),
         }
     }
 
@@ -54,6 +67,7 @@ impl HttpRequestBody {
         match self.0 {
             Inner::Bytes(b) => b.into(),
             Inner::PutPayload(_, payload) => Bytes::from(payload).into(),
+            Inner::Streaming(body) => Bytes::from(body.payload).into(),
         }
     }
 
@@ -62,6 +76,7 @@ impl HttpRequestBody {
         match &self.0 {
             Inner::Bytes(x) => x.is_empty(),
             Inner::PutPayload(_, x) => x.iter().any(|x| !x.is_empty()),
+            Inner::Streaming(x) => x.payload.content_length() != 0,
         }
     }
 
@@ -70,6 +85,7 @@ impl HttpRequestBody {
         match &self.0 {
             Inner::Bytes(x) => x.len(),
             Inner::PutPayload(_, x) => x.content_length(),
+            Inner::Streaming(x) => x.payload.content_length(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


alamb:
You can also use Mulit-part upload to achieve something similar (though most object stores would still require you to buffer the parts in 5mb chunks)

We haven't released these docs yet, btu here is an example of how to do it

Screenshot 2026-08-17 at 5 43 00 PM

alamb:
There is more rationale for non streaming APIs here:
- https://github.com/apache/arrow-rs/issues/5458
- https://github.com/apache/arrow-rs/pull/5500

TLDR is that streaming writes make a bunch of other operations (like retry , error handling, etc)

alamb:
I see you have talked about multi-part uploads here: https://github.com/apache/arrow-rs-object-store/issues/281#issuecomment-5123463473

senzzzi:
> You can also use Mulit-part upload to achieve something similar (though most object stores would still require you to buffer the parts in 5mb chunks)
> 
> We haven't released these docs yet, btu here is an example of how to do it
> 
> Screenshot 2026-08-17 at 5 43 00 PM

It's not the same because of costs. Uploading a 64Mb file in a single put request is still quite fast (depending on your needs) and you are charged for 1 request. If you upload 64 different 1Mb parts, you are charged by the initial request, 64 parts individually and 1 final requests. So it's 66 times more expensive. You can check the link in this comment: https://github.com/apache/arrow-rs-object-store/issues/281#issuecomment-5123463473

src/client/http/body.rs
```diff
@@ -102,14 +118,41 @@ impl From<String> for HttpRequestBody {
...
+    stream: Option<BoxStream<'static, Result<Bytes, HttpError>>>,
+    bytes_read: usize,
+    finished: bool,
+}
+
+impl Clone for PutPayloadBody {
+    fn clone(&self) -> Self {
+        Self {
+            payload: self.payload.clone(),
+            stream: None,
```
tustvold, line 151: This would likely be rather limiting in practice as it would break retries in client implementations

tustvold (commented):
I think the idea is a good one, but as Andrew says we need to think about how retries, etc... would work

alamb:
There is a bunch more discussion on https://github.com/apache/arrow-rs-object-store/issues/281 now too

PR #819, opened 2026-07-31
Fix #818 - WriteMultipart::finish should abort after part upload failure

### LINKED ISSUE #818: WriteMultipart::finish should abort after part upload failure
**Describe the bug**

`WriteMultipart::finish` does not abort the multipart upload when an in-flight `put_part` fails, because that error is surfaced in `self.wait_for_capacity(0).await?;` and returned early, skipping the cleanup that happens on an error in `self.upload.complete().await`

**Expected behavior**

Once a multipart upload has started, finish() should attempt to abort it when any part upload fails, as it already does when complete() fails.

**Additional context**

anson-vandoren:

- Closes #818

# Rationale for this change
 
Prevent an avoidable retention of partial multipart upload without resorting to lifecycle rules. The same is already done for the other error case in the fn, just not here.

Do a clean multipart upload abort if one of the final parts failed the upload.

No

# Are there any breaking changes?

No

### DIFF
--- a/src/upload.rs
+++ b/src/upload.rs
@@ -229,7 +229,10 @@ impl WriteMultipart {
             self.put_part(part.into())
         }
 
-        self.wait_for_capacity(0).await?;
+        if let Err(e) = self.wait_for_capacity(0).await {
+            self.abort().await?;
+            return Err(e);
+        }
 
         match self.upload.complete().await {
             Err(e) => {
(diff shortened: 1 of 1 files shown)


kylebarron (approved):

alamb:
Thanks @anson-vandoren and @kylebarron

PR #826, opened 2026-08-07
Support LocalFileSystem on Android

### LINKED ISSUE #459: Fails to create table on Android

mrkorchun:
Closes #459. Supersedes the closed #460 with the requested Linux-testable path.

Android app storage rejects hard links with EACCES. This keeps PutMode::Create and CopyMode::Create atomic by copying to a unique staging file and publishing it with renameat2(RENAME_NOREPLACE). CopyMode::Overwrite also stages by byte copy. The current fsync/close-error guarantees are preserved.

A private test-only toggle makes the existing LocalFileSystem integration helpers exercise the Android path on Linux; it adds no public feature.

Verified:

- cargo check --lib
- cargo test file_test_without_hard_links
- the v0.13.2 backport of the same mechanism: native aarch64-linux-android LocalFileSystem suite, 17 passed / 1 ignored

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -234,8 +234,9 @@ fn close_handle(handle: std::os::windows::io::RawHandle) -> std::result::Result<
 ///
 /// # Cross-Filesystem Copy
 ///
-/// [`LocalFileSystem::copy_opts`] is implemented using [`std::fs::hard_link`], and therefore
-/// does not support copying across filesystem boundaries.
+/// On platforms that support hard links, [`LocalFileSystem::copy_opts`] uses
+/// [`std::fs::hard_link`] and cannot copy across filesystem boundaries. Android uses a staged
+/// byte copy because app storage rejects hard links.
 ///
 #[derive(Clone, Debug)]
 pub struct LocalFileSystem {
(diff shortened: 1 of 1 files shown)


mrkorchun:
Confirmed on the affected device from ordinary Termux, outside PRoot:

- aarch64 Android, kernel 6.6.98
- Rust host: aarch64-linux-android
- std::fs::hard_link: EACCES (reproduced)
- OmniGraph/Lance dataset initialization using the backport: succeeded
- exit status: 0

This is an end-to-end confirmation that the renameat2 fallback works where the original hard-link path fails.

PR #827, opened 2026-08-07
Signer: reexport `url::Url` and `http::Method` types in the `signer` module

Tpt:
Minor "ease of use" nit: allow downstream create to not have to explicitly import `url` and `http` crates to use the `Signer` trait

# Rationale for this change

Makes downstream life easier. I had multiple times to add explicit dependencies to these crates just to be able to use the `Signer` trait 

`pub use url::Url` and `pub use http::method::Method` to the `signer` module

### DIFF
--- a/src/signer.rs
+++ b/src/signer.rs
@@ -19,9 +19,9 @@
 
 use crate::{Result, path::Path};
 use async_trait::async_trait;
-use http::Method;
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thank you @Tpt

PR #830, opened 2026-08-12
Return AlreadyExists in azure backend when using PutMode::Create and precondition fails

### LINKED ISSUE #829: Using `PutMode::Create` on Azure puts doesn't return `AlreadyExists` on failure
[Docs](https://docs.rs/object_store/latest/object_store/enum.PutMode.html#variant.Create) say that using `PutMode::Create` will return `AlreadyExists` if a put errors due to a failed precondition. That [doesn't seem to be the case for the implemented azure backend, though.](https://github.com/apache/arrow-rs-object-store/blob/main/src/azure/client.rs/#L767)

I think azure should return `AlreadyExists` for this case, just like the aws and gcp backends.

itsjunetime:

- Closes #829.

This simply translates the `Precondition` error case to `AlreadyExists` in azure put requests when using `PutMode::Create`.

# Rationale for this change

This is simply what is necessary to fix #829.

Yes, but this change brings behavior in-line with documentation, so there's no documentation changes needed. It may need calling-out in release notes, though, if people were relying on `Precondition` being returned in this case instead of `AlreadyExists`.

### DIFF
--- a/src/azure/client.rs
+++ b/src/azure/client.rs
@@ -764,7 +764,16 @@ impl AzureClient {
             }
         };
 
-        let response = builder.header(&BLOB_TYPE, "BlockBlob").send().await?;
+        // based on https://learn.microsoft.com/en-us/azure/storage/blobs/concurrency-manage, azure
+        // responds with `Precondition` when any put with a precondition fails, but we promise to
+        // return `AlreadyExists` when that put mode is `Create`.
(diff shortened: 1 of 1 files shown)


alamb (approved):
Thank you @itsjunetime  -- this makes sense to me. I took the liberty of merging this branch up from main and pushing a unit test to prevent regressions

PR #812, opened 2026-07-21
fix: restore 1.85 MSRV and fix MSRV CI

LDeakin:
- Closes https://github.com/apache/arrow-rs-object-store/issues/811

This just avoids using let chains, which were introduced in 1.88 or something.

PR #837, opened 2026-08-17
Fix MSRV violation in `parse.rs` test, check `--all-targets` in MSRV CI

alamb:

N/A — follow-up to a report by @progval in https://github.com/apache/arrow-rs-object-store/pull/812#issuecomment-5061268463

# Rationale for this change

Some things still fail to compile on main with 1.85

```
error[E0599]: no function or associated item named `from_utf8` found for type `str` in the current scope
   --> src/parse.rs:477:25
    |
477 |         let body = str::from_utf8(&body).unwrap();
    |                         ^^^^^^^^^ function or associated item not found in `str`
```

CI did not catch this because the MSRV job runs doesn't check all targeets

1. Update MSRV check to include all targets
1. Fix MSRV 1.85 issues

# Are these changes tested?

Yes by CI

No, test and CI changes only.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -55,4 +55,6 @@ jobs:
       - name: Check
         run: |
           # run `cargo msrv verify` to see problems
-          cargo msrv verify --all-features --output-format=json || exit 1
+          # use a custom check command so tests, examples, and benches are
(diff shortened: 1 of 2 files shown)


kylebarron (approved):
Thanks!

PR #838, opened 2026-08-17
Speed up MSRV CI check by installing cargo-msrv via taiki-e/install-action

alamb:
# Rationale
- The MSRV check is already the longest CI check
- https://github.com/apache/arrow-rs-object-store/pull/812 makes it even longer (reasonably so)

It would be nice to make this faster

# Changes
Installs `cargo-msrv` as a prebuilt binary via `taiki-e/install-action` (already used for `typos`) instead of compiling it from source with `cargo install`, which dominated the MSRV job's runtime.

## Before this pR
Here is an example from a recent run
https://github.com/apache/arrow-rs-object-store/actions/runs/31621559671/job/95504364705?pr=830

Screenshot 2026-08-17 at 5 03 02 PM

## After this PR
Screenshot 2026-08-17 at 5 04 57 PM
(note the actual check takes longer due to https://github.com/apache/arrow-rs-object-store/pull/812)

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -51,7 +51,9 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - name: Install cargo-msrv
-        run: cargo install cargo-msrv --locked
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-msrv
       - name: Check
         run: |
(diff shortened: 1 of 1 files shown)


kylebarron (approved):

kylebarron:
Looks great, thanks!

PR #839, opened 2026-08-17
Docs: point users uploading large objects at the multipart API

alamb:

N/A -- documentation only

# Rationale for this change

While discussing https://github.com/apache/arrow-rs-object-store/pull/817 it became clear the docs don't make it obvious that `put` / `put_opts` require the entire payload to be buffered in memory, and that the multipart API  is the intended way to upload large objects with bounded memory use.

Documentation hints (no code changes) at the places users are likely to land when looking for how to upload data:

Documentation only.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -263,6 +263,9 @@
 //!
 //! Use the [`ObjectStoreExt::put`] method to atomically write data.
 //!
+//! To upload large objects without buffering them entirely in [`PutPayload`],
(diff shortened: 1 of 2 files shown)


alamb:
@kylebarron  or @tustvold  any chance I can get a review of this (doc only) PR?

kylebarron (approved):
Thanks!

alamb:
Thank you for the review @kylebarron

Issue #831, opened 2026-08-17
Retrying failed part PUTs in multipart writes


criccomini:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**

SlateDB currently has a `RetryingObjectStore` that handles `ObjectStore` retries. The `object_store` retry mechanisms are unsuitable for us. This is something we've discussed in various GH issues in the past. We do something like this:

```rs
    #[inline]
    fn should_retry(err: &object_store::Error) -> bool {
        let retry = !matches!(
            err,
            object_store::Error::AlreadyExists { .. }
                | object_store::Error::Precondition { .. }
                | object_store::Error::NotModified { .. }
                | object_store::Error::NotFound { .. }
                | object_store::Error::NotImplemented { .. }
                | object_store::Error::NotSupported { .. }
        );
        if !retry {
            debug!("not retrying object store operation [error={:?}]", err);
        }
        retry
    }
```

This approach is somewhat similar to what Lance does (https://github.com/apache/arrow-rs-object-store/issues/737).

The problem we have right now is there's no good way for us to handle retries for multipart PUTs. We currently inherit `object_store`'s behavior for those writes. This behavior is insufficient (https://github.com/slatedb/slatedb/issues/1956).

**Describe the solution you'd like**

We've been wrestling a bit with the best way to handle this. The current `object_store` API makes it impossible for us to retry a multipart upload without buffering the entire byte object and retrying from scratch (as far as I can tell--please correct me if I'm wrong). I am seeking advice/suggestions at this point. 

There are a few challenges here:

- The multipart puts are async. We don't see a part failure until later.
- We don't know which parts failed AFAICT.
- There's no high-level API for us to retry a failed part. I see `MultipartStore::put_part(.., part_idx, data)`, but that isn't exposed at the `ObjectStore` level.

What do you think? Should `object_store` to support custom retry policies for multipart uploads (or for uploads in general)? Should users be allowed to retry a specific part write?

It's quite possible I'm misunderstanding something here, so LMK if that's the case.

**Describe alternatives you've considered**

Living with what we have, which can halt SlateDB databases in certain scenarios.

**Additional context**

See GH links above.

tustvold:
For context https://github.com/apache/arrow-rs/pull/5500 contains the original design intent for the trait and some of the reasons for the current design.

[...]

Issue #762, opened 2026-06-17
Release object store `0.15.0` (breaking) - Target TBD


alamb:
**Is your feature request related to a problem or challenge? Please describe what you are trying to do.**
- part of https://github.com/apache/arrow-rs-object-store/issues/392

We will make this release on demand, when we accumulate any breaking changes

**Describe the solution you'd like**
- [ ] Update version + make release notes: 
- [ ] Prepare release candidate + begin vote thread
- [ ] Release to crates.io
- [ ] File tickets for next planned releases

# Breaking Changes

# Request

alamb:
Given we don't have any breaking changes currently planned, I updated the schedule of this ticket to say TBD and filed some other minor release issues
- https://github.com/apache/arrow-rs-object-store/issues/832
- https://github.com/apache/arrow-rs-object-store/issues/833

PR #841, opened 2026-08-18
feat(retry): allow customized retry policy

### LINKED ISSUE #840: Feature request: allow customized retry policy

Hi team, I'm using OCI object storage in production, and I found it returns 403 or 404 when temporary authentication failures.

For now, we only retry [limited error status](https://github.com/apache/arrow-rs-object-store/blob/d5737ddec752bb7de38ea94482f1f8fbaa3b5aa9/src/client/retry.rs#L408-L411), and there's no way for us to customize our retry policy; for my case specially, I do want to retry on 403/404.

Make retry policy a functor, so users could customize their own retry logic; meanwhile default policy shouldn't be changed.

There're some well-known retry libraries on the market, like [backon](https://github.com/Xuanwo/backon), we could also adopt it as well.

**Additional context**

I'm discussing with OCI team on the HTTP status code, but would like to have customized retry policy to unblock.

dentiny:

- Closes https://github.com/apache/arrow-rs-object-store/issues/840

# Rationale for this change
 
Different storage backends have different behaviors -- not all of them conform the "common sense" of HTTP status code, so I think deciding whether to retry based on the status code and message is beneficial.

This PR abstracts a retry policy, so users could customize with own policy.

Yes.

### DIFF
--- a/src/client/retry.rs
+++ b/src/client/retry.rs
@@ -25,6 +25,7 @@ use futures_util::future::BoxFuture;
 use http::StatusCode;
 use http::header::LOCATION;
 use http::{Method, Uri};
+use std::sync::Arc;
 #[cfg(not(all(target_arch = "wasm32", target_os = "unknown")))]
 use std::time::{Duration, Instant};
 use tracing::info;
@@ -81,6 +82,7 @@ pub(crate) struct RetryContext {
     retries: usize,
     max_retries: usize,
     retry_timeout: Duration,
+    retry_status_policy: Option<RetryStatusPolicy>,
     start: Instant,
 }
 
@@ -89,6 +91,7 @@ impl RetryContext {
         Self {
             max_retries: config.max_retries,
             retry_timeout: config.retry_timeout,
+            retry_status_policy: config.retry_status_policy.clone(),
             backoff: Backoff::new(&config.backoff),
             retries: 0,
             start: Instant::now(),
@@ -102,6 +105,18 @@ impl RetryContext {
     pub(crate) fn backoff(&mut self) -> Duration {
         self.retries += 1;
         self.backoff.next()
+    }
+
+    fn should_retry_status(&self, status: StatusCode, retry_on_conflict: bool) -> bool {
+        match &self.retry_status_policy {
+            Some(policy) => policy(status),
+            None => {
+                status.is_server_error()
+                    || status == StatusCode::TOO_MANY_REQUESTS
+                    || status == StatusCode::REQUEST_TIMEOUT
+                    || (retry_on_conflict && status == StatusCode::CONFLICT)
+            }
+        }
     }
 }
 
@@ -212,6 +227,9 @@ impl From<RetryError> for std::io::Error {
 
 pub(crate) type Result<T, E = RetryError> = std::result::Result<T, E>;
 
+/// A function that determines whether a response status should be retried
+pub type RetryStatusPolicy = Arc<dyn Fn(StatusCode) -> bool + Send + Sync>;
+
 /// The configuration for how to respond to request errors
 ///
 /// The following categories of error will be retried:
@@ -225,7 +243,7 @@ pub(crate) type Result<T, E = RetryError> = std::result::Result<T, E>;
 /// backoff with jitter. See [`BackoffConfig`] for more information
 ///
 /// [safe]: https://datatracker.ietf.org/doc/html/rfc7231#section-4.2.1
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct RetryConfig {
     /// The backoff configuration
     pub backoff: BackoffConfig,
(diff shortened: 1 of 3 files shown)


src/client/retry.rs
```diff
@@ -247,6 +265,52 @@ pub struct RetryConfig {
...
+            .field("retry_timeout", &self.retry_timeout)
+            .field(
+                "retry_status_policy",
+                &self.retry_status_policy.as_ref().map(|_| "<custom>"),
+            )
+            .finish()
+    }
+}
+
+impl RetryConfig {
```
peterxcli, line 291: either
```suggestion
impl RetryConfig {
    fn default_should_retry_status(status: StatusCode) -> bool {
        status.is_server_error()
            || status == StatusCode::TOO_MANY_REQUESTS
            || status == StatusCode::REQUEST_TIMEOUT
    }

}
```

src/client/retry.rs
```diff
@@ -103,6 +106,18 @@ impl RetryContext {
...
+
+    fn should_retry_status(&self, status: StatusCode, retry_on_conflict: bool) -> bool {
+        match &self.retry_status_policy {
+            Some(policy) => policy(status),
+            None => {
+                status.is_server_error()
+                    || status == StatusCode::TOO_MANY_REQUESTS
+                    || status == StatusCode::REQUEST_TIMEOUT
+                    || (retry_on_conflict && status == StatusCode::CONFLICT)
+            }
```
peterxcli, line 116:
```suggestion
        match &self.retry_status_policy {
    Some(policy) => policy(status),
    None => {
        RetryConfig::default_should_retry_status(status)
            || (self.retry_on_conflict && status == StatusCode::CONFLICT)
    }
};
```

src/client/retry.rs
```diff
@@ -405,10 +470,7 @@ impl RetryableRequest {
                     } else {
                         let status = r.status();
                         if ctx.exhausted()
-                            || !(status.is_server_error()
-                                || status == StatusCode::TOO_MANY_REQUESTS
-                                || status == StatusCode::REQUEST_TIMEOUT
-                                || (self.retry_on_conflict && status == StatusCode::CONFLICT))
+                            || !ctx.should_retry_status(status, self.retry_on_conflict)
```
peterxcli, line 477: or inline 
```rust
let should_retry = match &ctx.retry_status_policy {
    Some(policy) => policy(status),
    None => {
        RetryConfig::default_should_retry_status(status)
            || (self.retry_on_conflict && status == StatusCode::CONFLICT)
    }
};

if ctx.exhausted() || !should_retry {
    // ...
}
```

peterxcli (commented):
I think we could give `RetryConfig` two functions — an internal one that calls another function that can be overridden/customized — so we can move the default behavior into `RetryConfig`. We could also simplify `should_retry_status` to just `if let Some(policy) => policy`, or even inline that and get rid of the new `should_retry_status` function entirely.

Xuanwo:
How about we adopt `backon` 👻

src/client/retry.rs
```diff
@@ -247,6 +265,52 @@ pub struct RetryConfig {
...
     pub retry_timeout: Duration,
+
+    /// An optional function that determines whether a response status should be retried
+    ///
+    /// When set, this replaces the default status policy, including any provider-specific
+    /// status handling. Successful responses, redirects, and `304 Not Modified` responses
+    /// are handled before this policy is invoked.
+    ///
+    /// Transport errors are classified separately and are not affected by this policy.
+    pub retry_status_policy: Option<RetryStatusPolicy>,
```
alamb, line 274: since this is new field in a struct with all public fields, this is breaking API change

alamb:
> How about we adopt `backon` 👻

I would rather allow people to adopt whatever API they want :)

alamb:
I think this sounds like a reasonable idea -- thank you @dentiny -- this would be nice to include in the next major release though I am not sure when that would be
- https://github.com/apache/arrow-rs-object-store/issues/762

I haven't looked at the code closely, but it sounds like @peterxcli has some good suggestions.

dentiny:
> I think this sounds like a reasonable idea -- thank you @dentiny -- this would be nice to include in the next major release though I am not sure when that would be
> 
> * [Release object store `0.15.0` (breaking) - Target TBD #762](https://github.com/apache/arrow-rs-object-store/issues/762)
> 
> I haven't looked at the code closely, but it sounds like @peterxcli has some good suggestions.

Thanks for the review! I will address the comments later today.

dentiny:
> How about we adopt `backon` 👻

Thanks for the notice and comment! I mentioned it as alternative solution in the feature request issue, but thought it'd be a better change -- would like to focus some scoped improvement first, `backon` could be a separate issue and PR if maintainers are happy with it.

src/client/retry.rs
dentiny: Yes, it's intentional :)

dentiny:
@peterxcli Hey I moved the default behavior into `default_should_retry_status` based on your suggestion and kept the small `should_retry_status` util to centralize custom-policy. Would appreciate if you could take another look.

peterxcli (approved):
LGTM, thanks for the update!

tustvold:
I'm a little lukewarm on this as it extends the existing retry policy introducing a breaking change, whilst also being relatively limited. I feel if we want to allow customisation of retry functionality we'd be better off making the whole retry function pluggable so people can choose different back off logic, etc...

For the specific case in the ticket I _think_ you could just provide an HttpClient wrapper that maps the status codes to the appropriate HttpError?

dentiny:
> I feel if we want to allow customisation of retry functionality we'd be better off making the whole retry function pluggable so people can choose different back off logic, etc...

I agree that, if we are going to introduce a breaking change, making the entire retry mechanism pluggable would provide a more complete extension point.

One possible approach would be to adopt backon, which would provide configurable retry and backoff behavior instead of introducing another custom abstraction. There has already been some discussion about this above ([1](https://github.com/apache/arrow-rs-object-store/pull/841#issuecomment-5348160598), [2](https://github.com/apache/arrow-rs-object-store/pull/841#issuecomment-5349620311)).

Would you prefer adopting backon, or exposing a more general pluggable retry interface that does not require users to adopt a specific retry library?

tustvold:
I left some thoughts on https://github.com/apache/arrow-rs-object-store/issues/831#issuecomment-5389162716

IMO we already have an HttpService abstraction that encapsulates how HTTP requests are made, it would seem ideal to me if retries would just be a layer in an HttpService stack.

As for adding an explicit dependency on a retry library, I am generally pretty lukewarm on adding additional dependencies unless strictly necessary - I am not sure this passes that bar.

alamb:
> IMO we already have an HttpService abstraction that encapsulates how HTTP requests are made, it would seem ideal to me if retries would just be a layer in an HttpService stack.

I agree -- this would help users as well so they could extend out the retry mechanism with whatever policy they wanted, rather than be limited to what was provided in the library

criccomini:
I took a crack at a _very_ limited `RetryPolicy` that specifically addresses #831:

https://github.com/apache/arrow-rs-object-store/pull/849

A subsequent PR could theoretically use the `RetryPolicy` trait for the `HttpService` layer as well (by adding a `RetryService` and `AmazonS3Builder::with_retry_policy`, etc).

PR #848, opened 2026-08-27
feat: add `Any` as a supertrait of `ObjectStore`

timsaucer:

None — this is a one-line trait change, so I skipped filing an issue. Happy to open one for changelog purposes if maintainers prefer.

# Rationale for this change

Downstream crates that hold an `Arc<dyn ObjectStore>` sometimes need to recover the concrete store behind it. In DataFusion, stores are registered and later retrieved from an `ObjectStoreRegistry` as `Arc<dyn ObjectStore>`, and there is currently no supported way to get back to the concrete type once it has been erased. The alternatives — an `as_any` method every implementor has to write, or a downstream wrapper trait — both add boilerplate for something the language already gives us.

`ObjectStore` already requires `'static`, and `std::any::Any` is blanket implemented for every `?Sized + 'static` type, so adding `Any` as a supertrait requires no changes from any implementation, in-tree or downstream.

- Add `std::any::Any` to the `ObjectStore` supertrait list.
- Document the new capability under a `# Downcasting` heading on the trait, including the Rust 1.86 caveat below.
- Add `test_dyn_impl_is_any`, alongside the existing `test_dyn_impl` / `test_generic_impl` bound tests, as a regression guard so the supertrait is not silently dropped later.

No implementation in this crate needed changing.

## A note on MSRV

**This PR does not change the MSRV.** Declaring the supertrait compiles on 1.85, and `cargo check` (what the `msrv` CI job runs) passes on 1.85 with this change — verified locally with `cargo +1.85.0 check`.

However, the *trait object upcast* that makes downcasting possible —
`Arc<dyn ObjectStore>` → `Arc<dyn Any + Send + Sync>` — was stabilized in Rust 1.86. So:

- On Rust >= 1.86, users get full `downcast` / `downcast_ref`.
- On Rust 1.85, users only get `Any::type_id` on a `dyn ObjectStore`.

The trait docs state this explicitly, and the upcast example is marked ```` ```ignore ```` so it is not compiled at the current MSRV. The in-crate test asserts `type_id` rather than downcasting, for the same reason.

If maintainers would rather bump the MSRV to 1.86 and document/test the downcast directly, I'm happy to do that instead — say the word.

Yes, additive:

- `dyn ObjectStore` can now be downcast to the concrete store type (Rust >= 1.86).

On breaking changes: adding a supertrait is technically a public API change, but it should not break anyone in practice. Every existing implementation already satisfies `Any` via the `'static` bound the trait has always required, so no downstream `impl ObjectStore for ...` needs to change. The only theoretical break is method resolution for `type_id`, if a downstream crate has some other trait with a `type_id` method in scope for its store type.

Verified locally: `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features` (no new warnings), `cargo test --all-features --lib`, `cargo test --all-features --doc`, and `cargo +1.85.0 check`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -801,6 +801,31 @@ pub type MultipartId = String;
 /// If a store implements [`Clone`], that will only clone the handle to the underlying data. It will NOT clone/fork the
 /// actual key-value data. Hence, the cloned instance and the original instance share the same state.
 ///
+/// # Downcasting
+/// [`Any`] is a supertrait of [`ObjectStore`] so that users can recover the concrete
+/// store behind a `dyn ObjectStore`. This requires nothing from implementors: the
+/// trait already requires `'static`, and [`Any`] is blanket implemented for every
+/// `'static` type.
+///
+/// ```
+/// # use std::any::TypeId;
+/// # use std::sync::Arc;
+/// # use object_store::{ObjectStore, memory::InMemory};
+/// let store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+/// assert_eq!(store.as_ref().type_id(), TypeId::of::<InMemory>());
+/// ```
+///
+/// Recovering the concrete type additionally requires a trait object upcast, which
+/// is only available on Rust 1.86 and later (this crate's MSRV is 1.85):
+///
+/// ```ignore
+/// let store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+/// let any: Arc<dyn Any + Send + Sync> = store; // requires Rust >= 1.86
+/// let store: Arc<InMemory> = any.downcast().unwrap();
+/// ```
+///
+/// [`Any`]: std::any::Any
+///
 /// # Minimal Default Implementations
 /// There are only a few default implementations for methods in this trait by
 /// design. This was different from versions prior to `0.13.0`, which had many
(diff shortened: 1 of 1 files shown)


alamb:
> adding a supertrait is technically a public API change, but it should not break anyone in practice.

I am not sure about this ...

src/lib.rs
```diff
@@ -801,6 +801,31 @@ pub type MultipartId = String;
 /// If a store implements [`Clone`], that will only clone the handle to the underlying data. It will NOT clone/fork the
 /// actual key-value data. Hence, the cloned instance and the original instance share the same state.
 ///
+/// # Downcasting
+/// [`Any`] is a supertrait of [`ObjectStore`] so that users can recover the concrete
+/// store behind a `dyn ObjectStore`. This requires nothing from implementors: the
```
alamb, line 806: I think we can probably remove the " This requires nothing from implementors: the" sentenece -- or at least make it easier to udnerstand ("most users should not have to do anything special...")

src/lib.rs
```diff
@@ -801,6 +801,31 @@ pub type MultipartId = String;
...
+/// trait already requires `'static`, and [`Any`] is blanket implemented for every
+/// `'static` type.
+///
+/// ```
+/// # use std::any::TypeId;
+/// # use std::sync::Arc;
+/// # use object_store::{ObjectStore, memory::InMemory};
+/// let store: Arc<dyn ObjectStore> = Arc::new(InMemory::new());
+/// assert_eq!(store.as_ref().type_id(), TypeId::of::<InMemory>());
```
alamb, line 815: can we use `downcast::<InMemory>().unwrap()` perhaps rather than checking the type id? That would show how to get backa concrete pointer.

alamb (commented):
Overall the idea looks good to me -- I am not sure how to evaluate the potential API changes (and if we could release this in 0.14.2 or have to wait for 0.15.0)

PR #849, opened 2026-08-29
feat: retry failed multipart part uploads

criccomini:
This PR addresses #831 by adding caller-controlled retries for multipart part uploads.

- Adds `RetryPolicy` and `PutMultipartOptions::with_retry_policy`.
- Updates the Amazon S3, Azure, and GCS multipart implementations to apply the policy to individual part uploads.

The existing HTTP retry logic is bounded and retries an already-built request. That works for short-lived failures, but retrying for longer periods can outlive the request's credentials or signature.

This implementation retries outside the HTTP layer. After the existing bounded retries are exhausted, the backend repeats the logical part-upload operation. This rebuilds and re-signs the request while preserving the part identifier and payload.

I initially explored integrating `RetryPolicy` at the `HttpService` layer, following the discussion in #831 and #841. That is possible, but it requires a much broader change. This PR deliberately limits the scope to multipart uploads, where the retry gap exists today.

`RetryPolicy` itself is not multipart-specific. A future change could expose the same policy through `AmazonS3Builder`, `MicrosoftAzureBuilder`, and `GoogleCloudStorageBuilder` for other operations.

(WARN: This is Sol 5.6 xhigh coded. I just spent a bunch of time iterating with it.)


criccomini:
Nudging this: any interest?

alamb:
I'll try and review this shortly -- sorry for the delay

criccomini:
@alamb no need to apologize! Just wanted to see if there's interest at least. :) 

One open question is whether to include the actual Error in the payload in addition to the kind and status.

src/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+#[cfg(all(target_arch = "wasm32", target_os = "unknown"))]
+use web_time::Instant;
+
+/// A normalized HTTP failure passed to a [`RetryPolicy`]
+///
+/// Backends wrap request failures in provider-specific [`Error`] variants.
+/// Before invoking a policy, `object_store` reduces a supported failure to an
+/// HTTP response status or transport error kind. Failures that cannot be
+/// represented by this enum are returned without invoking the policy.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
alamb, line 49: deriving Copy here I think makes it impossible to add the actual error (which may not derive copy). So I am going to remove that derivation to give us more options in the future

src/retry.rs
```diff
@@ -0,0 +1,257 @@
...
+    /// operation boundary, allowing credentials and request signatures to be
+    /// refreshed. Initiating, completing, and aborting the multipart upload
+    /// retain their existing retry behavior.
+    #[must_use]
+    pub fn with_retry_policy(mut self, policy: Arc<dyn RetryPolicy>) -> Self {
+        self.extensions.insert(MultipartRetryPolicy(policy));
+        self
+    }
+
+    pub(crate) fn retry_policy(&self) -> Option<Arc<dyn RetryPolicy>> {
```
alamb, line 122: Another thing that Claude flagged is that since `MultipartRetryPolicy` extension wrapper is private and `retry_policy()` is pub(crate), so only in-crate backends can ever read the policy;

As a follow on it may make sense to make this getter public (and perhaps document via doc examples)  so others implementations can use it

src/azure/mod.rs
```diff
@@ -287,10 +291,23 @@ impl MultipartUpload for AzureMultiPartUpload {
         let idx = self.part_idx;
         self.part_idx += 1;
         let state = Arc::clone(&self.state);
+        let content_id = AzureClient::new_block_id();
         Box::pin(async move {
-            let part = state.client.put_block(&state.location, idx, data).await?;
-            state.parts.put(idx, part);
-            Ok(())
+            let mut retry = MultipartRetry::new(state.retry_policy.clone());
```
alamb, line 296: these are very elegant loops (in that they express the "retry on failure" very clearly)

alamb (approved):
Thank you @criccomini  -- I went though this carefully and I think it looks great. As you mention, it sets the stage for a more generic retry mechanism, but keeps the initial implementation reasonably scoped. 

> One open question is whether to include the actual Error in the payload in addition to the kind and status.

I think it would probably be good to include the actual Error , but we can perhaps do that as a follow on PR. In order to do this in a non breaking way, I removed the `derive(Copy)` from the `RetryFailure` struct.

Since I am about to make a release candidate, I am going to merge this PR in and then cut a release

src/azure/mod.rs
criccomini: Sol has totally replaced me 😭

